PR #6, opened 2021-08-04
make gradf return None for non-trainable parameters

NTT123:
Currently, `gradf` returns `0` for non-trainable parameters. A better approach is to return `None` instead.

### DIFF
--- a/examples/train_mlp.py
+++ b/examples/train_mlp.py
@@ -77,10 +77,7 @@ def main(
     for step, (x, y) in zip(range(steps), data):
         value, grads = loss(model, x, y)
         updates, opt_state = optim.update(grads, opt_state)
-        # Essentially equivalent to optax.apply_updates, it just doesn't try to update anything with a zero gradient.
-        # Anything we filtered out above will have a zero gradient. But in general some of the things we filtered out
-        # might not even be jnp.arrays, and might not even have a notion of addition. We don't want to try adding zero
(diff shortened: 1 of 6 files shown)


patrick-kidger:
Sounds reasonable to me; thanks for the PR.

Looks like the tests are failing because of an unused import, though.

Could you update the README please? I think there's two pieces that need changing: API/gradf and API/apply_updates, both of which make explicit reference to 0 rather than `None`.

examples/train_mlp.py
```diff
@@ -77,10 +77,7 @@ def loss(model, x, y):
     for step, (x, y) in zip(range(steps), data):
         value, grads = loss(model, x, y)
         updates, opt_state = optim.update(grads, opt_state)
-        # Essentially equivalent to optax.apply_updates, it just doesn't try to update anything with a zero gradient.
-        # Anything we filtered out above will have a zero gradient. But in general some of the things we filtered out
-        # might not even be jnp.arrays, and might not even have a notion of addition. We don't want to try adding zero
-        # to arbitrary Python objects.
+        # Essentially equivalent to optax.apply_updates, it just doesn't try to update anything with a `None` gradient.
```
patrick-kidger, line 80: I'd keep the latter half of this comment, just switch "zero" for "`None`".

patrick-kidger:
Alright, LGTM! Thanks for contributing.

PR #8, opened 2021-08-04
Create a Parameter container to specify trainable parameter

NTT123:
The problem of having a filter function to decide if a parameter is trainable is that, for `batch_norm` or any non-trainable real value tensor, we currently cannot decide if its parameters are trainable or not.

My proposal is: to have a `Parameter` class which is a sub-class of `Module` to indicate if a tensor is a trainable parameter or not.  The model is still a pytree but now can store `requires_grad` information like Pytorch.

I just want to see if you are interested in this proposal. If not, just say so!

Adding a new method `parameters` to `Module` class. This method returns a Boolean tree of the model to indicate if a leave is trainable. The output of this method is used as the `filter_tree` parameter of `gradf` function.

In order to JIT compile an `update` function, which inputs a model and outputs an updated model, I added two helper methods.
* `remove_unjitable_fields`  replace all non-tensor fields of the tree by `None`. 
* `update` update all tensor values (non-None) of the tree.

### DIFF
--- a/equinox/module.py
+++ b/equinox/module.py
@@ -2,6 +2,7 @@ import abc
 from dataclasses import dataclass, fields
 
 import jax
+import jax.numpy as jnp
 
 from equinox.custom_types import Array
 
@@ -98,6 +99,15 @@ class Module(metaclass=_ModuleMeta):
         ans = cls.__new__(cls, *mask)
         cls.__dataclass_init__(ans, *mask)
         return ans
+
+    def remove_unjitable_fields(self):
+        """remove all fields of type int, float, str, etc."""
+        def filter_fn(x): return x if isinstance(x, (jnp.DeviceArray, jax.core.Tracer)) else None
+        return jax.tree_map(filter_fn, self)
+
+    def update(self, updates):
+        """update non-None components"""
+        return jax.tree_multimap(lambda s, u: s if u is None else u, self, updates)
 
 
 class Parameter(Module):
--- a/equinox/module.py
+++ b/equinox/module.py
@@ -3,6 +3,8 @@ from dataclasses import dataclass, fields
 
 import jax
 
+from equinox.custom_types import Array
+
 from .tree import tree_equal
 
 
@@ -91,10 +93,15 @@ class Module(metaclass=_ModuleMeta):
             f = getattr(self, field.name)
             if isinstance(f, Module):
                 mask.append(f.parameters())
-            elif field.name.startswith('_'):
-                mask.append(False)
             else:
-                mask.append(True)
+                mask.append(False)
         ans = cls.__new__(cls, *mask)
         cls.__dataclass_init__(ans, *mask)
         return ans
+
+
+class Parameter(Module):
+    value: Array
+
+    def parameters(self):
+        return Parameter(True)
--- a/equinox/module.py
+++ b/equinox/module.py
@@ -83,3 +83,18 @@ class _ModuleMeta(abc.ABCMeta):
 class Module(metaclass=_ModuleMeta):
     def __eq__(self, other):
         return tree_equal(self, other)
+
+    def parameters(self):
+        cls = self.__class__
+        mask = []
+        for field in fields(cls):
+            f = getattr(self, field.name)
+            if isinstance(f, Module):
+                mask.append(f.parameters())
+            elif field.name.startswith('_'):
+                mask.append(False)
+            else:
+                mask.append(True)
+        ans = cls.__new__(cls, *mask)
+        cls.__dataclass_init__(ans, *mask)
+        return ans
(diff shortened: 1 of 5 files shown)


patrick-kidger:
Thanks for the PR!

I am afraid this is actually an idea I'd already thought of and rejected. Let me explain the rationale.

First of all, there isn't just `filter_fn` but also `filter_tree` as an argument to `eqx.jitf`, `eqx.gradf`. If `filter_fn` is too general then `filter_tree` allows you to specify inclusion/noninclusion on a per-leaf level -- so we're definitely not losing expressivity under the current approach.

Second, philosophically speaking, JAX has already chosen to distinguish the differentiablity/nondifferentiability/etc. at the `jax.jit`/`jax.grad` level, when these operations are called. I wanted to maintain that way of doing things, rather than encoding it directly into the structure of a Module.

And if you really do want to encode it into the Module anyway, then -- third -- you can just use a `lax.stop_gradient` prior to accessing such a buffer.

Fourth, I just didn't like all the extra `.value`s that needed to be added everywhere to unpack the parameter.

Thanks for your work on this nonetheless! I'm open to discussion on the design decisions here.

NTT123:
Thank you for your thoughtful reply.

Let me answer your points:

> First of all, there isn't just filter_fn but also filter_tree as an argument to eqx.jitf, eqx.gradf. If filter_fn is too general then filter_tree allows you to specify inclusion/noninclusion on a per-leaf level -- so we're definitely not losing expressivity under the current approach.

For a ResNet50, there are many `batch_norm` layers (~50 layers, iirc) each has two non-trainable parameters. These parameters look the same from `filter_fn` function point of view. Also, It is very laborious  to manually generate a `filter_tree` in this case. We have to store `is_trainable` information somewhere. 

> Second, philosophically speaking, JAX has already chosen to distinguish the differentiablity/nondifferentiability/etc. at the jax.jit/jax.grad level, when these operations are called. I wanted to maintain that way of doing things, rather than encoding it directly into the structure of a Module.

My proposal is to store this information using the structure of the `tree` itself. I think this will not affect Jax's way of doing thing as from Jax's point of view, it is just a pytree.

> And if you really do want to encode it into the Module anyway, then -- third -- you can just use a lax.stop_gradient prior to accessing such a buffer.

You make me to think about a potential bug here, the end-user can forget to call `stop_gradient` causing a silent bug that very difficult to detect.

> Fourth, I just didn't like all the extra .values that needed to be added everywhere to unpack the parameter.

From an end-user's point of view, one rarely needs to create a new custom module with `Parameter` class, so this is not too inconvenient. A different approach is to have `NonTrainable` class to specify `non-differentiable` parameters but one can forget to declare that causing silent bug.

As you said, one can use `stop_gradient` in case of `batch_norm`, however, there is a cost that we have to pay in memory and GPU/TPU time.

Philosophically speaking, I think having a structure separation between `trainable` and `non-trainable` parameters is a good thing. It is in fact a bit inconvenient, but it has a clear mental model: every parameters is non-trainable unless specified.

patrick-kidger:
I don't believe generating filter trees should be laborious. To a great extent it's an academic question as batch norm isn't implemented in Equinox at the moment, but hypothetically producing a filter tree would look something like

```python
is_batchnorm = lambda model: tuple(layer.stats for layer in model.layers if isinstance(layer, BatchNorm))
filter_tree = jax.tree_map(lambda _: True, model)
filter_tree = eqx.tree_at(is_batchnorm, model, replace=False)
```

I disagree that `stop_gradient` is too much of a bug-risk. The end-user has nothing to do with it as that's something that would be encoded in the forward pass of a hypothethical batch norm layer. 

In particular, I think I would argue that `Parameter` introduces the risk of forgetting to use `Parameter`, so that subsequently the model doesn't train certain weights. This is equally insidious, but as a use-case far more common.

I'm afraid we may have to agree to disagree on this one.

NTT123:
Thank you for your comments!

luweizheng:
Hi @patrick-kidger 

Just want to know the easiest way to specify non-trainable parameter in eqx?

In the following example, `sf` is just a constant array and we do not need to update grads.

```python
DummyClass(eqx.Module):
    sf: jnp.ndarray
```

In the optax part, I do it like this:

```
optimizer = optax.adam(learning_rate=learning_rate_fn)

opt_state = optimizer.init(eqx.filter(model, eqx.is_inexact_array))
```

But I guess `eqx.filter(model, eqx.is_inexact_array)` will only filter out functions, python int or other variables.

patrick-kidger:
The appropriate thing to do is to partition the model into trainable/non-trainable pieces.

A convenient way to do this is provided by `eqx.filter_grad`; see the example in the documentation on freezing some parameters.

PR #12, opened 2021-09-02
fix minor variable mix up in examples

jrfinkbeiner:
Replaced `depth` with `width_size` in `train_mlp.py` and `frozen_layer.py` examples


patrick-kidger:
LGTM! Thanks for the fix.

Issue #1, opened 2021-07-29
Add convolutional layers + transposed convolutional layers


patrick-kidger:
See for example https://jax.readthedocs.io/en/latest/_modules/jax/experimental/stax.html#GeneralConv or https://flax.readthedocs.io/en/latest/_modules/flax/linen/linear.html#Conv

- [x] Convolutional layers (#17, #18)
- [ ] Transposed convolutional layers

PR #17, opened 2021-12-15
Add convolutional layers

andyehrenberg:
I've really enjoyed learning about Equinox, and would love to be able to contribute!

Here's a first effort at adding convolutions (#1) - likely needs some fixes, and transposed convs need to be added.

It currently uses the workaround used by flax and haiku to handle non-batched inputs. Would be great if [https://github.com/google/jax/discussions/4092](url) is addressed so no hacks are required to make vmap work with convolutions.

### DIFF
--- a/equinox/nn/conv.py
+++ b/equinox/nn/conv.py
@@ -1,14 +1,15 @@
-import math
-from typing import Optional, Sequence, Union
 import collections
+import math
 from itertools import repeat
+from typing import Optional, Sequence, Union
 
+import jax.numpy as jnp
 import jax.random as jrandom
 from jax.lax import conv_general_dilated
-import jax.numpy as jnp
 
 from ..custom_types import Array
 from ..module import Module, static_field
+
 
 def _ntuple(n):
     def parse(x):
@@ -17,6 +18,7 @@ def _ntuple(n):
         return tuple(repeat(x, n))
 
     return parse
+
 
 class Conv(Module):
     num_spatial_dims: int = static_field()
@@ -42,7 +44,7 @@ class Conv(Module):
         use_bias=True,
         *,
         key,
-        **kwargs
+        **kwargs,
     ):
         super().__init__()
         self.num_spatial_dims = num_spatial_dims
@@ -56,22 +58,37 @@ class Conv(Module):
 
         self.weight = jrandom.uniform(
             wkey,
-            (self.out_channels, self.in_channels,) + self.kernel_size,
+            (
+                self.out_channels,
+                self.in_channels,
+            )
+            + self.kernel_size,
             minval=-lim,
-            maxval=lim
+            maxval=lim,
         )
         if self.use_bias:
-            self.bias = jrandom.uniform(bkey, (self.out_channels,) + (1,) * self.num_spatial_dims, minval=-lim, maxval=lim)
+            self.bias = jrandom.uniform(
+                bkey,
+                (self.out_channels,) + (1,) * self.num_spatial_dims,
+                minval=-lim,
+                maxval=lim,
+            )
         else:
             self.bias = None
 
         self.stride = parse(stride)
         if type(padding) == tuple and len(padding) == self.num_spatial_dims:
-            self.padding = tuple([parse(padding[i]) for i in range(self.num_spatial_dims)])
+            self.padding = tuple(
+                [parse(padding[i]) for i in range(self.num_spatial_dims)]
+            )
         elif type(padding) == int:
-            self.padding = tuple([(padding, padding) for _ in range(self.num_spatial_dims)])
+            self.padding = tuple(
+                [(padding, padding) for _ in range(self.num_spatial_dims)]
+            )
         else:
-            raise ValueError(f"Padding must either be int or tuple of length {self.num_spatial_dims}")
+            raise ValueError(
+                f"Padding must either be int or tuple of length {self.num_spatial_dims}"
+            )
         self.dilation = parse(dilation)
 
     def __call__(self, x, *, key=None):
@@ -80,7 +97,7 @@ class Conv(Module):
         if x.ndim not in allowed_ranks:
             raise ValueError(
                 f"Input to Conv needs to have rank in {allowed_ranks},",
-                f" but input has shape {x.shape}."
+                f" but input has shape {x.shape}.",
             )
         unbatched = x.ndim == unbatched_rank
         if unbatched:
@@ -90,13 +107,14 @@ class Conv(Module):
             rhs=self.weight,
             window_strides=self.stride,
             padding=self.padding,
-            rhs_dilation=self.dilation
+            rhs_dilation=self.dilation,
         )
         if self.bias is not None:
             x += jnp.broadcast_to(self.bias, x.shape)
         if unbatched:
             x = jnp.squeeze(x, axis=0)
         return x
+
 
 class Conv1d(Conv):
     weight: Array
@@ -120,7 +138,7 @@ class Conv1d(Conv):
         use_bias=True,
         *,
         key,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(
             num_spatial_dims=1,
@@ -131,8 +149,9 @@ class Conv1d(Conv):
             padding=padding,
             dilation=dilation,
             use_bias=use_bias,
-            key=key
+            key=key,
         )
+
 
 class Conv2d(Conv):
     weight: Array
@@ -150,13 +169,13 @@ class Conv2d(Conv):
         in_channels,
         out_channels,
         kernel_size,
-        stride=1,
-        padding=0,
-        dilation=1,
+        stride=(1, 1),
+        padding=(0, 0),
+        dilation=(1, 1),
         use_bias=True,
         *,
         key,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(
             num_spatial_dims=2,
@@ -167,8 +186,9 @@ class Conv2d(Conv):
             padding=padding,
             dilation=dilation,
             use_bias=use_bias,
-            key=key
+            key=key,
         )
+
 
 class Conv3d(Conv):
     weight: Array
@@ -186,13 +206,13 @@ class Conv3d(Conv):
         in_channels,
         out_channels,
         kernel_size,
-        stride=1,
-        padding=0,
-        dilation=1,
+        stride=(1, 1, 1),
+        padding=(0, 0, 0),
+        dilation=(1, 1, 1),
         use_bias=True,
         *,
         key,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(
             num_spatial_dims=3,
@@ -203,5 +223,5 @@ class Conv3d(Conv):
             padding=padding,
             dilation=dilation,
             use_bias=use_bias,
-            key=key
+            key=key,
         )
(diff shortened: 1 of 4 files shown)


equinox/nn/conv.py
```diff
@@ -0,0 +1,227 @@
...
+        kernel_size,
+        stride=1,
+        padding=0,
+        dilation=1,
+        use_bias=True,
+        *,
+        key,
+        **kwargs,
+    ):
+        super().__init__()
```
patrick-kidger, line 49: It would be best to pass `**kwargs` here for the sake of co-operative multiple inheritance.

patrick-kidger: (Admittedly it looks like I've been inconsistent with this one myself in the rest of Equinox!)

equinox/nn/conv.py
```diff
@@ -0,0 +1,227 @@
...
+                minval=-lim,
+                maxval=lim,
+            )
+        else:
+            self.bias = None
+
+        self.stride = parse(stride)
+        if type(padding) == tuple and len(padding) == self.num_spatial_dims:
+            self.padding = tuple(
+                [parse(padding[i]) for i in range(self.num_spatial_dims)]
```
patrick-kidger, line 82: Nit: the `[ ]` brackets are superfluous.

equinox/nn/conv.py
```diff
@@ -0,0 +1,227 @@
...
+from jax.lax import conv_general_dilated
+
+from ..custom_types import Array
+from ..module import Module, static_field
+
+
+def _ntuple(n):
+    def parse(x):
+        if isinstance(x, collections.abc.Iterable):
+            return tuple(x)
```
patrick-kidger, line 17: Would it make sense to check that `x` is of length `n` here? (And raise a `ValueError` if not.)

equinox/nn/conv.py
```diff
@@ -0,0 +1,227 @@
...
+        unbatched_rank = self.num_spatial_dims + 1
+        allowed_ranks = [unbatched_rank, unbatched_rank + 1]
+        if x.ndim not in allowed_ranks:
+            raise ValueError(
+                f"Input to Conv needs to have rank in {allowed_ranks},",
+                f" but input has shape {x.shape}.",
+            )
+        unbatched = x.ndim == unbatched_rank
+        if unbatched:
+            x = jnp.expand_dims(x, axis=0)
```
patrick-kidger, line 104: So I'm tempted to suggest that only the unbatched case should be supported here. (Wth the appropriate un/squeezing.) WDYT?

equinox/nn/conv.py
```diff
@@ -0,0 +1,227 @@
...
+class Conv1d(Conv):
+    weight: Array
+    bias: Optional[Array]
+    in_channels: int = static_field()
+    out_channels: int = static_field()
+    kernel_size: Union[int, Sequence[int]] = static_field()
+    stride: Union[int, Sequence[int]] = static_field()
+    padding: Union[int, Sequence[int]] = static_field()
+    dilation: Union[int, Sequence[int]] = static_field()
+    use_bias: bool = static_field()
```
patrick-kidger, line 128: It actually isn't necessary to restate fields - they'll be inherited automatically from the base class.

patrick-kidger (commented):
So this is amazingly cool to see implemented. Thanks a lot for putting this together, and I can see the care with which you've followed the style of the existing library.

I've left some nitty comments below but otherwise this LGTM.

Can you add the new layers to the documentation in the README + write a quick test in the style of [test_nn.py](https://github.com/patrick-kidger/equinox/blob/main/tests/test_nn.py)?

If you want to bump the version number then this deserves a new release I think. :))

equinox/nn/conv.py
andyehrenberg: Nice catch. I borrowed this from PyTorch - looks like no one spotted the problem there yet!

equinox/nn/conv.py
andyehrenberg: Sounds good to me - probably nice to enforce using vmap for more consistent style

tests/test_nn.py
```diff
@@ -108,3 +108,120 @@ def test_mlp(getkey):
...
+        key=getkey(),
+    )
+    x = jrandom.normal(getkey(), (3, 3, 32, 32))
+    assert conv(x).shape == (1, 2, 16, 16)
+
+
+def test_conv(getkey):
+    test_conv1d(getkey)
+    test_conv2d(getkey)
+    test_conv2d(getkey)
```
patrick-kidger, line 227: This will actually mean that each `test_conv*d` will be run twice. pytest automatically detects all functions whose names begin with `test`, so it'll pick up on `test_conv`, `test_conv1d`, `test_conv2d` and `test_conv3d` separately.

In this case I think all that needs to happen is to delete `test_conv`.

---

Besides that: I think it'd be good to explicitly test the output value (and not just the shape) in just a couple of very simple cases, just as a sanity check. Something like:

```python
conv = Conv1d(...)
new_weight = jnp.array(...)
data = jnp.array(...)
assert new_weight.shape == conv.weight.shape  # make sure the test itself is valid
conv = eqx.tree_at(lambda x: x.weight, conv, new_weight)  # manually set the weight
assert jnp.allclose(conv(data), ...)
```

patrick-kidger (commented):
Thanks for the updates!

It looks like the tests are failing on the pre-commit hooks. These run some basic formatting / sanity checks on the code, to help keep things tidy. If you install pre-commit as per [CONTRIBUTING.md](https://github.com/patrick-kidger/equinox/blob/main/CONTRIBUTING.md) then all of these should run automatically whenever you make a commit.

patrick-kidger:
@andyehrenberg Just checking if you're planning on continuing to work on this? No worries if not -- I'd be happy to pick up where you've left off before merging this, as I think this would be a great addition to the library.

andyehrenberg:
> @andyehrenberg Just checking if you're planning on continuing to work on this? No worries if not -- I'd be happy to pick up where you've left off before merging this, as I think this would be a great addition to the library.

Yeah I'll definitely see this through! Been sidetracked the past couple weeks.

jrfinkbeiner:
Hi all,
nice work! I am using equinox quite a lot and therefore my interest in the conv-layer. Anyways, here are a few (minor) comments and questions. All of them could also be included at a later stage.

### 1. Support batched case
Why is there the check for dimensions with a `ValueError` in `__call__` (https://github.com/andyehrenberg/equinox/blob/7592292742547118cef1d530571976aa410d6a82/equinox/nn/conv.py#L99-L105) that rules out batched input tensors?
```python
        unbatched_rank = self.num_spatial_dims + 1
        if x.ndim != unbatched_rank:
            raise ValueError(
                f"Input to Conv needs to have rank {unbatched_rank},",
                f" but input has shape {x.shape}.",
            )
        x = jnp.expand_dims(x, axis=0)
```
Couldn't instead this be used:         
```python
	unbatched = x.ndim == (self.num_spatial_dims + 1)
        if unbatched:
            x = jnp.expand_dims(x, axis=0)  
```
and before return:
```python
        if unbatched:
            x = jnp.squeeze(x, axis=0)
```            
or is there a reason not to do it?

### 2. Initializers
This might be more general as it is the same for the linear layers: Should the initializer be rewritten in every class? There exist initializers in `jax.nn.initializers`. One could e.g. use `jax.nn.initializers.kaiming`. Additionally, using just zeros as initializer for the bias is more common, I think. 

### 3. Choice of Channel Format
This comment is about your choice of not setting the argument `dimension_numbers`, and thereby using the jax default. I think, the expected format should somehow be documented in the class. (Default behaivour for 2D is stated in `jax.lax.conv_general_dilated`, additionally it is supposed to be defined somewhere in jaxlibs `xla_client.py`, but I couldn't find that yet). Alternatively, one could expose a `data_format` or `channels_last` argument or something similar to the user as done by tensorflow and then set `dimension_numbers` accordingly. 

The choice of channel format can actually make a difference in performace (due to the additional overhead to perform the transpose operation). The NCHW (channel-first) format used to be more performant, however, with newer NVIDIA-GPUs (with tensor-cores) NHWC (channel-last) is faster (if I understand this correctly from https://forums.developer.nvidia.com/t/nhwc-vs-nchw-convolution/111065 and https://docs.nvidia.com/deeplearning/performance/dl-performance-convolutional/index.html). Therefore it might be useful to expose this choice to the user. I am not sure how it is implemented under the hood by xla tho, and whether the performance gains/losses are noticeable.

### 4. Support more arguments
I guess there is no rush for this, but why not expose more arguments of `jax.lax.conv_general_dilated` to the user, especially simple ones like `precision`, where one doesn't even have to perform checks or alter other code? Later, one could think  of multiple groups which are somewhat commonly used as well.

If you want to include one of my comments, but don't want to do it yourselves, I am happy to help.

patrick-kidger:
@jrfinkbeiner -- thanks for getting involved! It's always nice to hear about a project getting used.

1. The batched case is supported by wrapping things in a `jax.vmap` (for consistency with the rest of JAX).

2. The existing initialisers -- for both the existing `eqx.nn.Linear` and the proposed convolutional layers -- are chosen to match the initialisation scheme used by PyTorch. Which for the weights is in turn is actually `kaiming` as you suggest.

For the biases, this is again following PyTorch/`kaiming`. It's true biases are instead sometimes initialised to zero -- or something slightly offset from zero when using ReLUs, etc. -- but I'm not really aware of a strong consensus here. I'm inclined to just follow PyTorch on this one, but happy to hear arguments the other way if this is something you feel strongly about.

3./4. Haha, I swear the optimal format changes every time I look it up. (I've even seen N-last formats being optimal in some cases, probably due to something SIMD-related.)

If you're happy to put together a PR I'd be happy to expose these as extra options for users.

jrfinkbeiner:
@patrick-kidger thanks for the quick reply! Then I'll wait for this PR to be merged and then add a channel format argument and maybe some other arguments.

tests/test_nn.py
andyehrenberg: Latest commit now checks output values and removed the unnecessary test_conv function - let me know if the tests look good!

equinox/nn/conv.py
```diff
@@ -46,15 +50,15 @@ def __init__(
...
+        super().__init__(**kwargs)
         self.num_spatial_dims = num_spatial_dims
         parse = _ntuple(self.num_spatial_dims)
         wkey, bkey = jrandom.split(key, 2)
         self.in_channels = in_channels
         self.out_channels = out_channels
         self.kernel_size = parse(kernel_size)
         self.use_bias = use_bias
-        lim = 1 / math.sqrt(self.in_channels * math.prod(self.kernel_size))
+        lim = 1 / jnp.sqrt(self.in_channels * jnp.prod(jnp.array(self.kernel_size)))
```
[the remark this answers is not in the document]
andyehrenberg: Definitely didn't think this through - switched to using plain numpy

andyehrenberg:
> 

I think i've been using the pre-commit hook this whole time - definitely used it for the latest commit - hopefully formatting is correct.

equinox/nn/conv.py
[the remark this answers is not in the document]
andyehrenberg: And I think i've been using the pre-commit hook this whole time - definitely used it for the latest commit - hopefully formatting is correct now

patrick-kidger:
Alright -- LGTM! Thank you for contributing.

Issue #15, opened 2021-09-23
Add FAQ


patrick-kidger:
- Optax compatibility: `optax.optimiser(eqx.filter(model, eqx.is_array))`
- Make sure your model is a tree, not a DAG.

PR #34, opened 2022-02-25
ConvTranspose layers, MultiheadAttention, lookup embeddings, LayerNorm

andyehrenberg:
Let me know if these should be put into separate PRs!

- For ConvTranspose layers, I'm using the PyTorch style argument for output_padding instead of haiku's output_size - haiku has the limitation that padding can only be either 'same' or 'valid', though other options seem to be much less frequently useful. My current implementation also has this limitation - I haven't thought deeply enough about the best way to deal with any combination of padding and output_padding. Probably needs some fixing.

- The other 3 modules feel like pretty typical, barebones implementations

- When playing around with trying to reimplement [grokking](https://arxiv.org/abs/2201.02177) using equinox/these new layers, I found that trying to initialize an optax optimizer with an `eqx.nn.MLP` via `optim.init(model)` throws a TypeError about `optax.transform.init_fn` requiring all arguments to `jnp.zeros_like` to be arrays or scalars. My workaround is just using `params, static = eqx.partition(model, eqx.is_inexact_array)` and then `opt_state = optim.init(params)` - is this sort of thing unavoidable? Most other modules aren't suffering from this, so following along with your [demo](https://docs.kidger.site/equinox/examples/train_rnn/) usually works.


patrick-kidger:
Okay, this is amazing! I'll go through and add comments in-line, but glancing over this looks very well done.
(I'm not too fussed about making these separate PRs.)

CC @lucidrains -- I recall you saying on Reddit that you were thinking of implementing some of these, so please feel free to express an opinion / leave review comments if you want.

On initialising optimisers: yep, this is something I'm aware of, c.f. #15. I've been meaning to document it. I think what you're doing is the correct way to handle it.

equinox/nn/attention.py
```diff
@@ -0,0 +1,136 @@
...
+from ..custom_types import Array
+from ..module import Module, static_field
+from .dropout import Dropout
+from .linear import Linear
+
+
+class MultiheadAttention(Module):
+    """
+    Multihead Attention layer from 'Attention Is All You Need' (https://arxiv.org/abs/1706.03762)
+    """
```
patrick-kidger, line 17: Would it be possible to add the mathematical formulation here?

PyTorch kind-of do this for attention [here](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention), but I'm imagining something more precise, a la [here](https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html#torch.nn.Conv1d).

The documentation generation supports LaTeX: `stuff $\alpha$ morestuff`.

equinox/nn/attention.py
```diff
@@ -0,0 +1,136 @@
...
+        self,
+        embed_dim: int,
+        num_heads: int,
+        dropout: float = 0.0,
+        use_bias: bool = True,
+        kdim: Optional[int] = None,
+        vdim: Optional[int] = None,
+        add_bias_kv: bool = False,
+        *,
+        key: "jax.random.PRNGKey",
```
patrick-kidger, line 41: The convention so far has been to add `**kwargs` and pass them on through `super().__init__(**kwargs)`, in case of co-operative multiple inheritance.

equinox/nn/attention.py
```diff
@@ -0,0 +1,136 @@
...
+        self.out_proj = Linear(embed_dim, embed_dim, use_bias=use_bias, key=key4)
+
+    def __call__(
+        self,
+        query: Array,
+        key: Array,
+        value: Array,
+        attn_mask: Optional[Array] = None,
+        *,
+        key_: Optional["jax.random.PRNGKey"] = None,
```
patrick-kidger, line 95: Hmm! The naming here is an interesting conflict. I think it'd be better to switch `key` and `key_` here, given the existing convention e.g. used here in [here](https://github.com/patrick-kidger/equinox/blob/328ca86834ae815593d9a26adffed7ff7165e698/equinox/nn/composed.py#L124) in `eqx.nn.Sequential`.

equinox/nn/attention.py
```diff
@@ -0,0 +1,136 @@
...
+
+        **Returns:**
+
+        A JAX array of shape `(sequence_length, embed_dim)`.
+        """
+        d1, _ = query.shape
+        query_heads = self._project(self.q_proj, query)
+        key_heads = self._project(self.k_proj, key)
+        value_heads = self._project(self.v_proj, value)
+        attn_logits = jnp.einsum("...shd,...Shd->...hsS", query_heads, key_heads)
```
patrick-kidger, line 113: Nitpick, but are the `...` needed? Assuming non-batched data. (Since batched data should come through via `vmap`.)

equinox/nn/conv.py
```diff
@@ -26,6 +26,52 @@ def parse(x: Any) -> tuple:
...
+) -> Tuple[int, int]:
+    """Computes adjusted padding for desired ConvTranspose `output_padding`."""
+    output_size = (
+        (input_size - 1) * stride
+        - 2 * padding
+        + dilation * (kernel_size - 1)
+        + output_padding
+        + 1
+    )
+    kernel_size = (kernel_size - 1) * dilation + 1
```
patrick-kidger, line 45: Nit: slightly neater to switch the definitions of `output_size` and `kernel_size`, as the latter is used in the former.

equinox/nn/conv.py
```diff
@@ -246,3 +292,251 @@ def __init__(
...
+        """**Arguments:**
+
+        - `num_spatial_dims`: The number of spatial dimensions. For example traditional
+            convolutions for image processing have this set to `2`.
+        - `in_channels`: The number of input channels.
+        - `out_channels`: The number of output channels.
+        - `kernel_size`: The size of the transposed convolutional kernel.
+        - `stride`: The stride of the transposed convolution.
+        - `padding`: The amount of implicit padding on both sides for `dilation *
+            (kernel_size - 1) - padding points.
```
patrick-kidger, line 337: Typo: I think there needs to be a closing backtick here.

equinox/nn/conv.py
```diff
@@ -246,3 +292,251 @@ def __init__(
...
+        - `use_bias`: Whether to add on a bias after the transposed convolution.
+        - `key`: A `jax.random.PRNGKey` used to provide randomness for parameter
+            initialisation. (Keyword only argument.)
+
+        !!! info
+
+            All of `kernel_size`, `stride`, `padding`, `output_padding`, `dilation` can
+            be either an integer or a sequence of integers. If they are a sequence then
+            the sequence should be of length equal to `num_spatial_dims`, and specify
+            the value of each property down each spatial dimension in turn.. If they
```
patrick-kidger, line 349: Typo: double full stop.

equinox/nn/conv.py
```diff
@@ -246,3 +292,251 @@ def __init__(
...
+        self.padding = parse(padding)
+        lim = 1 / np.sqrt(self.in_channels * np.prod(self.kernel_size))
+        if self.num_spatial_dims == 1:
+            self.dimension_numbers = ("NCH", "IOH", "NCH")
+        elif self.num_spatial_dims == 2:
+            self.dimension_numbers = ("NCHW", "IOHW", "NCHW")
+        elif self.num_spatial_dims == 3:
+            self.dimension_numbers = ("NCDHW", "IODHW", "NCDHW")
+        else:
+            raise ValueError(
```
patrick-kidger, line 372: I think this should probably be a `NotImplementedError`.

equinox/nn/conv.py
```diff
@@ -246,3 +292,251 @@ def __init__(
...
+                f" but input has shape {x.shape}.",
+            )
+        x = jnp.expand_dims(x, axis=0)
+        padding = self.padding
+        if self.output_padding is not None:
+            padding = tuple(
+                map(
+                    lambda s, k, st, p, o, d: compute_adjusted_padding(
+                        s, k, st, p, o, d
+                    ),
```
patrick-kidger, line 426: I think this `lambda` can be replaced with just `compute_adjusted_padding`?

equinox/nn/embedding.py
```diff
@@ -0,0 +1,65 @@
...
+    embedding_dim: int = static_field()
+    weight: Array
+
+    def __init__(
+        self,
+        num_embeddings: int,
+        embedding_dim: int,
+        weight: Optional[Array] = None,
+        *,
+        key: "jax.random.PRNGKey",
```
patrick-kidger, line 24: Same comment as before about `**kwargs`.

equinox/nn/embedding.py
```diff
@@ -0,0 +1,65 @@
...
+            (Keyword only argument.)
+
+        **Returns:**
+
+        A JAX array of shape `embedding_dim` that gives the xth index of the embedding table.
+        """
+        if not jnp.issubdtype(x, jnp.integer):
+            raise ValueError(
+                f"Input must be an array of integer dtype but input was {x.dtype}"
+            )
```
patrick-kidger, line 64: I think check is probably unnecessary -- attempting to index `self.weight` with something invalid already has its own set of checks. Not a strong feeling though.

equinox/nn/embedding.py
```diff
@@ -0,0 +1,65 @@
...
+        return self.weight[(x,)]
```
patrick-kidger, line 65: I think this is equivalent to just `self.weight[x]`.

equinox/nn/normalization.py
```diff
@@ -0,0 +1,69 @@
+from typing import List, Optional, Tuple, Union
+
+import jax
+import jax.numpy as jnp
+
+from ..custom_types import Array
+from ..module import Module, static_field
+
+
+_shape_t = Union[int, Tuple[int], List[int]]
```
patrick-kidger, line 10: `Union[int, Sequence[int]]` is probably what's meant here. (c.f. the implementation of something similar for the convolutional layers.) And as that's a bit shorter I think `_shape_t` can be inlined.

equinox/nn/normalization.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class LayerNorm(Module):
+    """Layer Normalization as described in https://arxiv.org/abs/1607.06450"""
```
patrick-kidger, line 14: I think it'd be great to have the precise mathematical description here, c.f. [here](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html#torch.nn.LayerNorm), including the precise meaning of `elementwise_affine`. (By all means just copy what they've written verbatim if you do/don't wish.)

patrick-kidger: (And the meaning of `normalized_shape`, actually.)

equinox/nn/normalization.py
```diff
@@ -0,0 +1,69 @@
...
+            offset = self.bias
+        else:
+            scale = jnp.array(1.0, dtype=x.dtype)
+            offset = jnp.array(0.0, dtype=x.dtype)
+        scale = jnp.broadcast_to(scale, x.shape)
+        offset = jnp.broadcast_to(offset, x.shape)
+        mean = jnp.broadcast_to(mean, x.shape)
+        eps = jax.lax.convert_element_type(self.eps, variance.dtype)
+        inv = scale * jax.lax.rsqrt(variance + eps)
+        return inv * (x - mean) + offset
```
patrick-kidger, line 69: I'm not sure all these casts and broadcasts are necessary.

Stuff like `scale = jnp.array(1.0)` can usually just be replaced with `scale = 1`. Likewise JAX usually automatically handles dtypes and such.

Untested, but I'm thinking of a slightly neater implementation like:
```python
mean = jnp.mean(x, keepdims=True)
variance = jnp.var(x, keepdims=True)
inv = lax.rsqrt(variance + self.eps)
out = (x - mean) * inv
if self.elemntwise_affine:
    out = self.weight * out + self.bias
return out
```

equinox/nn/normalization.py
```diff
@@ -0,0 +1,69 @@
...
+        - `eps`: Value added to denominator for numerical stability. Default: `1e-5`.
+        - `elementwise_affine`: Whether the module has learnable affine parameters. Default: `True`.
+        - `key`: Ignored; provided for compatibility with the rest of the Equinox API.
+            (Keyword only argument.)
+        """
+        self.normalized_shape = normalized_shape
+        self.eps = eps
+        self.elementwise_affine = elementwise_affine
+        self.weight = jnp.ones(self.normalized_shape)
+        self.bias = jnp.zeros(self.normalized_shape)
```
patrick-kidger, line 41: Maybe it's best to assign `self.weight = None` (and `bias`) if `elementwise_affine` is False? Save some memory.

tests/test_nn.py
```diff
@@ -254,3 +254,269 @@ def test_conv3d(getkey):
...
+            7,
+            3,
+            4,
+            1,
+        ]
+    ).reshape(1, 3, 3, 3)
+    assert jnp.allclose(conv(data), answer)
+
+
+def test_multiheadattn(getkey):
```
patrick-kidger, line 468: The smallest of nits: can this be renamed to `test_multihead_attention`?

patrick-kidger (commented):
Right, done making comments! There's quite a lot but that's only because this is such a big PR / and they're all pretty small/nitty stuff anyway. This is a great PR.

I also haven't thought too hard about padding for transposed convolutional layers. I think having just same/valid padding is probably fine for now, as long as you think it'll be easy to adjust without introducing any backward compatibility concerns, if need be?

By the way, do you want to (a) bump the version number, and (b) add these new layers to the documentation [here](https://github.com/patrick-kidger/equinox/tree/main/docs/api/nn)?

patrick-kidger:
@andyehrenberg Just following up on your plans for this? (I'd be happy to help out on this PR if you want.)

andyehrenberg:
@patrick-kidger Starting to work through your suggestions/corrections - got busy earlier this week. Thanks for the feedback!

equinox/nn/attention.py
andyehrenberg: Added documentation in the latest commit - though I had to skip using flake8 because it's complaining about the '\i' and '\s' in '\intercal' and '\sqrt'. How can this be ignored by the checks in this repo?

patrick-kidger:
Excellent! Let me know when this is ready for re-review.

equinox/nn/attention.py
patrick-kidger: This is because Python/flake8 is trying to interpret those as escape codes, like `\n` for a newline. Parsing of escape codes can be disabled by prefixing an `r` before the string, i.e.: `r""" ... stuff ... """`

PR #38, opened 2022-03-10
Attention, Transposed Convolutions, Embeddings, LayerNorm

patrick-kidger:
Merging #34 into master via this branch, so as to make some tweaks.

Updates relative to #34:
- Added new layers to the documentation.
- Added the mathematical details to the docstrings for `MultiheadAttention` and `LayerNorm`
- Substantially overhauled the `MultiheadAttention` implementation. The implementation in #34 was basically mimicking PyTorch's implementation, which has a very bad API. The new API is much more consistent, and much more general. (For reference, I found the Haiku implementation of `MultiheadAttention` to be the cleanest of all previous implementations.)

Plus some other misc updates in support of these changes:
- Dropout now takes a fast path when `p=0`.
- Enabled MathJax in the documentation.
- Bumped version number.
- Updated `custom_types.Array` and `custom_types.PyTree` to follow the convention in Diffrax; namely that they're now subscriptable and the documentation handles this.
- Standardised on the spelling "normalisation" (rather than "normalization") just because that's the convention I happen to follow.
- Other misc doc tweaks.

CC @andyehrenberg (and @lucidrains ?) for any commentary on these changes prior to merging them in.


andyehrenberg:
I agree that your overhauled `MultiheadAttention` has a better API. When I was implementing it, I was trying to make the Haiku implementation fit in with the PyTorch API, but I agree that it's pretty bad (with sort of misleading arguments). I've been putting more thought into `ConvTranspose` lately and should have some extensions ready soon that take inspiration from how https://github.com/google/jax/pull/5772 computes output sizes (mainly to deal with more `output_padding` and `padding` cases). It looks like I'm getting consistent outputs between this new implementation and Haiku's ConvTranspose but want to test things a bit more.

patrick-kidger:
Excellent. I'm assuming you implemented `MultiheadAttention` on the basis that you need it for your own work? If you could give it a try and check that it looks like it's working for your use cases -- just to be sure -- then that'd be great.

As for transposed convolutions: staring at the implementation I'm not feeling convinced by it. These things jump out at me:

1. this implementation only supports `padding = 0` or `padding = 1` -- perhaps this is what you're describing above.
2.  Would it be simpler to ignore `lax.conv_transpose` and work directly with `lax.conv_general_dilated`? Looking at the implementation of `lax.conv_transpose`, it's just a thin wrapper around `lax.conv_general_dilated` and I don't think we really hit any of the meaningful functionality that that wrapper provides.
3. We can probably compute `dimension_numbers` for all dimensions, just by working directly with `lax.ConvDimensionNumbers` instead of the simplified string replications. (It doesn't look that tricky.)

WDYT?

patrick-kidger:
With the other branch merged in: let me know once you're satisfied that both `MultiheadAttention` and `ConvTranspose` work for your use cases and I'll merge this branch + do a new release.

equinox/nn/attention.py
```diff
@@ -0,0 +1,237 @@
...
+                "query, key, and value must all be sequences of equal length."
+            )
+
+        query_heads = self._project(self.query_proj, query)
+        key_heads = self._project(self.key_proj, key_)
+        value_heads = self._project(self.value_proj, value)
+
+        logits = jnp.einsum("shd,Shd->hsS", query_heads, key_heads)
+        logits = logits / math.sqrt(self.qk_size)
+        logits = self.dropout(logits, key=key, deterministic=deterministic)
```
lucidrains, line 218: so this dropout should actually happen after the softmax below afaik (although there are recent papers that show you can drop / mask out (but using -inf) raw attn logits https://arxiv.org/abs/2202.07765)

```python
weights = jax.nn.softmax(logits, axis=-1)
weights = self.dropout(weights, key=key, deterministic=deterministic)
```

lucidrains: otherwise, lgtm! 🤗

equinox/nn/attention.py
```diff
@@ -0,0 +1,237 @@
...
+        query_size: int,
+        key_size: Optional[int] = None,
+        value_size: Optional[int] = None,
+        output_size: Optional[int] = None,
+        qk_size: Optional[int] = None,
+        vo_size: Optional[int] = None,
+        use_query_bias: bool = False,
+        use_key_bias: bool = False,
+        use_value_bias: bool = False,
+        use_output_bias: bool = False,
```
lucidrains, line 106: oh, this should default to True, very rarely is this turned off

equinox/nn/attention.py
patrick-kidger: Thanks, I'll make this change!

equinox/nn/attention.py
patrick-kidger: Will do. Do you mean all four, or just `use_output_bias`?

If you mean just `use_output_bias`, would it be reasonable to change the default on the other three just for consistency?

lucidrains: just use_output_bias, the rest is fine!

equinox/nn/attention.py
```diff
@@ -0,0 +1,237 @@
...
+
+        **Returns:**
+
+        A JAX array of shape `(seq_length, output_size)`.
+        """
+
+        seq_length, _ = query.shape
+        seq_length2, _ = key_.shape
+        seq_length3, _ = value.shape
+        if seq_length != seq_length2 or seq_length != seq_length3:
```
lucidrains, line 207: sorry, one last thing - so with this condition (and the message below for masking), you will only be allowing for self-attention, and not cross attention

to allow for cross attention (allow for full encoder - decoder), you will need to change this to `seq_length2 != seq_length3`

equinox/nn/attention.py
patrick-kidger: Yeah, the original paper has no biases at all. But if you reckon on having a bias being a more useful/expressive default then I'm happy to keep it this way. (It's already merged now anyway haha!)

lucidrains: @patrick-kidger I was wrong, it is better to have it turned off https://storage.googleapis.com/pathways-language-model/PaLM-paper.pdf :pray: sorry about that!

patrick-kidger: Oh no! Thanks for reporting back on that. I'll try and change this in the next breaking release.

PR #40, opened 2022-03-10
Tidied; simplified; generalised ConvTranspose implementation.

patrick-kidger:
@andyehrenberg I think I've got transposed convolutions sorted out.

WDYT? If you're happy with this then I'll merge this in to the other ongoing feature branch.

The main references I used to implement this were Relation 14 of [this reference](https://arxiv.org/abs/1603.07285) and [these](https://github.com/vdumoulin/conv_arithmetic/blob/af6f818b0bb396c26da79899554682a8a499101d/README.md#transposed-convolution-animations) animations.

### DIFF
--- a/equinox/nn/conv.py
+++ b/equinox/nn/conv.py
@@ -26,46 +26,6 @@ def _ntuple(n: int) -> callable:
     return parse
 
 
-def compute_adjusted_padding(
-    input_size: int,
-    kernel_size: int,
-    stride: int,
-    padding: int,
-    output_padding: int,
-    dilation: int,
-) -> Tuple[int, int]:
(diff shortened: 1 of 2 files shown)


andyehrenberg:
Definitely happy with this! I really like the clear explanation of `output_padding`. Do you think it’s worthwhile to include `output_shape` as an optional argument (that would have to agree with `output_padding`, if given), as haiku has done? While this could make model building more intuitive, specifying `output_padding` enables input images of varying sizes in the case of encoder-decoder style architectures, while `output_shape` could place too much of a restriction. I don’t think it’d be super important/useful to include it. Flax doesn’t seem to include either of these, as noted in https://github.com/google/flax/issues/1872.

PR #50, opened 2022-03-27
Version 0.3.0 -- BatchNorm and stateful

### LINKED ISSUE #15: Add FAQ

### LINKED ISSUE #33: Improve/remove caching for `filter_jit`
At the moment we have an LRU cache here: https://github.com/patrick-kidger/equinox/blob/328ca86834ae815593d9a26adffed7ff7165e698/equinox/jit.py#L16

This shouldn't be necessary, as we can implement the wrapped object as an `eqx.Module`.

```python
class WrappedFunc(eqx.Module):
    func: Callable
    jitkwargs: Dict[str, Any]

    @ft.partial(jax.jit, static_argnums=(0, 1, 2, 5))
    def __call__(self, static_leaves, static_treedef, dynamic_args, dynamic_kwargs, filter_spec_return):
        ...
```

That is, use the existing `jax.jit` cache to handle `self.func` and `self.jitkwargs`, rather than our own cache.

### LINKED ISSUE #45: Implementing Batch Normalization
In Flax, Batch Normalization is a bit finicky since each call to `apply` requires marking `batch_stats` as mutable and updating the `batch_stats` afterward.

```python
bn = flax.linen.BatchNorm(use_running_average=True)

x = jnp.arange(24).reshape(3, 6)

vars = bn.init(random.PRNGKey(0), x)

# Mark the batch stats as mutable so we can update them in the variable dictionary
x_normed, mutated_vars = bn.apply(vars, x, mutable=['batch_stats'])

vars = {**vars, **mutated_vars}  # Update the variables with our diff

x_normed2, mutated_vars2 = bn.apply(vars, x, mutable=['batch_stats'])
```

How could this be implemented as a Module in Equinox? I'm happy to submit an implementation given some guidance.

patrick-kidger:
- Removes all currently-deprecated functionality. This is the reason for the version bump.
- Added stateful operations: `get_state`, `set_state`, `StateIndex`.
  - These are serious magic, and should be used with great care.
- Added batch normalisation via stateful operations.
- Added FAQ.
- Removed an unnecessary level of caching in `filter_jit`.

Closes #45.
Closes #33.
Closes #15.


patrick-kidger:
@FedericoV Following on from where I tagged you in #45: try using `equinox.experimental.{get_state,set_state,StateIndex}` to implement the stateful repeated-optimisations we were discussing.

PR #52, opened 2022-03-28
Fix function names when logging compilation progress

### LINKED ISSUE #51: Compilation logging shows generic function name for every JIT'ed function for `filter_jit`, `filter_grad`, `filter_value_and_grad`
I've recently been debugging long compile times in my neural networks and stumbled upon an XLA compiler flag that can be set using

```python
jax.config.update('jax_log_compiles', True)
```

This setting prints out some messages when each function starts and ends its compilation including the amount of time spent. However, currently, the error message shows a generic function name from within the implementation of `filter_jit`, `filter_grad` and `filter_value_and_grad`, e.g.

```
WARNING:absl:Finished tracing + transforming prim_fun for jit in 0.00043702125549316406 sec
WARNING:absl:Finished tracing + transforming prim_fun for jit in 0.0002541542053222656 sec
WARNING:absl:Compiling prim_fun (4766936320 for args (ShapedArray(float32[]),).
WARNING:absl:Finished XLA compilation of broadcast_in_dim in 0.06613397598266602 sec
WARNING:absl:Finished tracing + transforming f_wrapped for jit in 0.0012121200561523438 sec
WARNING:absl:Compiling f_wrapped (5095206976 for args (ShapedArray(float32[1]),).
WARNING:absl:Finished XLA compilation of f_wrapped in 0.00688481330871582 sec
```

where `f_wrapped` comes from here
https://github.com/patrick-kidger/equinox/blob/3af0cdefd70e7a36850260a34cbc58d52c6a4b59/equinox/jit.py#L17

When looking for the culprit of excessive debugging times, having the names of the functions available is important.

marcelroed:
Fixes #51

### DIFF
--- a/equinox/jit.py
+++ b/equinox/jit.py
@@ -95,7 +95,7 @@ def filter_jit(
         static = (static_fun,) + static_args_kwargs
         static_leaves, static_treedef = jax.tree_flatten(static)
         static_leaves = tuple(static_leaves)
-        if isinstance(fun, ft.partial):
+        if isinstance(static_fun, ft.partial):
             inner_fun = static_fun.func
         else:
             inner_fun = static_fun
--- a/equinox/jit.py
+++ b/equinox/jit.py
@@ -96,9 +96,9 @@ def filter_jit(
         static_leaves, static_treedef = jax.tree_flatten(static)
         static_leaves = tuple(static_leaves)
         if isinstance(fun, ft.partial):
-            inner_fun = fun.func
+            inner_fun = static_fun.func
         else:
-            inner_fun = fun
+            inner_fun = static_fun
         dynamic_out, static_out = _f_wrapped_cache(inner_fun, **jitkwargs)(
             dynamic, static_treedef, static_leaves, filter_spec_return
         )
--- a/equinox/jit.py
+++ b/equinox/jit.py
@@ -25,6 +25,15 @@ def _f_wrapped_cache(fun, **jitkwargs):
     return f_wrapped
 
 
+def _strip_wrapped_partial(fun):
+    """Preserve the outermost wraps call's docstring or traverse to the inner function"""
+    if hasattr(fun, '__wrapped__'):
+        return _strip_wrapped_partial(fun.__wrapped__)
+    if isinstance(fun, ft.partial):
+        return _strip_wrapped_partial(fun.func)
+    return fun
+
+
 def filter_jit(
     fun,
     *,
@@ -95,10 +104,7 @@ def filter_jit(
         static = (static_fun,) + static_args_kwargs
         static_leaves, static_treedef = jax.tree_flatten(static)
         static_leaves = tuple(static_leaves)
-        if isinstance(static_fun, ft.partial):
-            inner_fun = static_fun.func
-        else:
-            inner_fun = static_fun
+        inner_fun = _strip_wrapped_partial(static_fun)
         dynamic_out, static_out = _f_wrapped_cache(inner_fun, **jitkwargs)(
             dynamic, static_treedef, static_leaves, filter_spec_return
         )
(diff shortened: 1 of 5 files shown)


equinox/jit.py
```diff
@@ -94,7 +95,11 @@ def fun_wrapper(*args, **kwargs):
         static = (static_fun,) + static_args_kwargs
         static_leaves, static_treedef = jax.tree_flatten(static)
         static_leaves = tuple(static_leaves)
-        dynamic_out, static_out = _f_wrapped_cache(**jitkwargs)(
+        if isinstance(fun, ft.partial):
+            inner_fun = fun.func
+        else:
+            inner_fun = fun
```
patrick-kidger, line 101: So I think this should be `static_fun` (and likewise `static_fun.func` in the partial branch). This is passed into the LRU cache so it needs to be hashable.

By the way, I think this code will fail in the case of passing in `ft.wraps(...)(ft.partial(...))`. This will see the partial and unwrap it, ignoring the fact that it's been `ft.wrap`'d. What do you think the best fix for that is?

tests/test_filter_jit.py
```diff
@@ -204,3 +204,51 @@ def __call__(self, x):
...
     eqx.filter_jit(m)(y)
     assert num_traces == 1
+
+
+@pytest.fixture
+def log_compiles_config():
+    """Setup and teardown of jax_log_compiles flag"""
+    jax.config.update('jax_log_compiles', True)
+    yield
+    jax.config.update('jax_log_compiles', False)
```
patrick-kidger, line 214: Can this be set to reset `jax_log_compiles` to whatever it was originally, and not necessarily just `False`?

patrick-kidger (commented):
Thanks for the contribution! Basically this LGTM.

tests/test_filter_jit.py
marcelroed: Sure, I'll change that. There's a context manager for all these flags, apparently.

equinox/jit.py
marcelroed: Why should I use `static_fun` here? Don't we care about the function we're decorating regardless?

patrick-kidger: So the reason is that `inner_fun` is being passed into `_f_wrapped_cache`. This has an LRU cache on it, which demands that all of its arguments be hashable. Mostly that's not an issue -- all functions are hashable -- but we might want to JIT something that isn't just a normal function.

For example a bound method, or a class implementing `__call__`. So we treat `fun` as a PyTree, and split it apart into both dynamic and static parts. Usually both pieces will have the same `__name__` etc. so `ft.wraps` should still just work.

patrick-kidger: In order to handle the `ft.wraps(...)(ft.partial(...))` case I think it'd be best to change the `if` condition to
`if isinstance(fun, ft.partial) and not hasattr(fun, "__name__"):`

marcelroed: I had a go at it in 955a7a4, let me know what you think of that solution

patrick-kidger: Ah yeah, that's even better. Nice.

Do you want to bump the version number? We can do a new release for this if this is something that would help you out.

marcelroed: Sure, that would be great. Just committed the bump.

patrick-kidger:
Looks like you need to run the pre-commit hooks. (See [CONTRIBUTING.md](https://github.com/patrick-kidger/equinox/blob/main/CONTRIBUTING.md).)

marcelroed:
> Looks like you need to run the pre-commit hooks. (See [CONTRIBUTING.md](https://github.com/patrick-kidger/equinox/blob/main/CONTRIBUTING.md).)

Aha, I forgot to run `black`

patrick-kidger:
Thanks for contributing! The autorelease bot is hard at work and will push a new release shortly :)

Issue #56, opened 2022-04-01
Better support for broadcasting in linear layer


jaschau:
In equinox's implementation, for a weight matrix `w` of shape `(d_out, d_in)`, the computational rule for the linear layer update is (https://github.com/patrick-kidger/equinox/blob/main/equinox/nn/linear.py#L65)
```
y = w x + b
```
This is problematic if `x` is a higher-dimensional tensor with shape `(d_in, d_1, ..., d_n)`. 
The outcome of w x is a tensor of shape `(d_out, d_1, ..., d_n)`, while `b` is a tensor of shape `(d_out,)`.
Since numpy's broadcasting rule works from right-to-left (https://numpy.org/doc/stable/user/basics.broadcasting.html#general-broadcasting-rules), broadcasting the two won't work.

Example:

```
import equinox as eqx
import equinox.experimental as eqxe
import jax
import jax.nn as jnn
import jax.numpy as jnp
import jax.random as jrandom
seed = 0

key = jrandom.PRNGKey(seed)
key, key_linear, key_input = jrandom.split(key, 3)
#print("key=", key)
d_in = 3
d_out = 4
x = jrandom.normal(key_input, shape=(d_in, 5))
linear = eqx.nn.Linear(d_in, d_out, use_bias=True, key=key_linear)
# fails with ValueError due to impossibility of broadcasting
res = linear(x)
```
To remedy this, I would propose to change the implementation to PyTorch style (https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html). This would allow arbitrary left-most dimensions.
If we could add a reshape to make it work with tensors with arbitrary number of trailing dimensions, that would be even better :)

patrick-kidger:
So this is actually a deliberate choice to keep Equinox's API simple and consistent.

If you need to handle additional dimensions, then this can be done: just use `jax.vmap`, as per `jax.vmap(linear)(x)` or `jax.vmap(linear, in_axes=(1,), out_axes=(1,))(x)`, etc.

jaschau:
Hi @patrick-kidger, thanks for the clarification. I have created a small PR adding the above explanation to the documentation to avoid confusion for future users.

patrick-kidger:
Great, thanks for the contribution! I really appreciate it.

PR #57, opened 2022-04-01
Explain how to use higher-order tensors in Linear

jaschau:
Adresses #56.

### DIFF
--- a/equinox/nn/linear.py
+++ b/equinox/nn/linear.py
@@ -57,6 +57,15 @@ class Linear(Module):
         - `key`: Ignored; provided for compatibility with the rest of the Equinox API.
             (Keyword only argument.)
 
+        !!! info
+
+        Note that in contrast to [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html) or other
+        deep learning frameworks where the input `x` is allowed to be a higher order tensor of shape
+        `(in_features, ...)`, here we require `x` to be a vector. If you want to use higher order tensors as
+        inputs, use `jax.vmap`.
+        For example, for an input `x` of shape `(in_features, d)`, using `jax.vmap(linear, in_axes=1, out_axes=1)(x)`
+        produces the appropriate output of shape `(out_features, d)`.
+
         **Returns:**
 
         A JAX array of shape `(out_features,)`
--- a/equinox/nn/linear.py
+++ b/equinox/nn/linear.py
@@ -59,12 +59,12 @@ class Linear(Module):
 
         !!! info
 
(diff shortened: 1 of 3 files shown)


equinox/nn/linear.py
```diff
@@ -57,6 +57,15 @@ def __call__(
         - `key`: Ignored; provided for compatibility with the rest of the Equinox API.
             (Keyword only argument.)
 
+        !!! info
+
+        Note that in contrast to [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.Linear.html) or other
+        deep learning frameworks where the input `x` is allowed to be a higher order tensor of shape
+        `(in_features, ...)`, here we require `x` to be a vector. If you want to use higher order tensors as
```
patrick-kidger, line 64: **(A)** PyTorch actually has `(..., in_features)`.
This means the Equinox example would equivalently change to
```python
linear = eqx.nn.Linear(...)
jax.vmap(linear)(x)
```

**(B)** The region below `!!! info` should be indented four spaces in order to format correctly in the documentation.

If you want to look at the documentation locally, by the way:
```
bash> pip install -r docs/requirements.txt
bash> mkdocs serve
bash> mkdocs serve  # again
```
and then navigate to `localhost:8000` in a web browser.

**(C)** This isn't actually unique to the `eqx.nn.Linear` layer. All Equinox operations operate on a single element, and need a `jax.vmap` to work over a batch. Perhaps it would be worth adding this to the FAQ? (Either as well as or instead of, whichever you think best.)

jaschau: Hi, thanks for the comments. 
(A) good catch, that is potentially confusing, and I will change it accordingly.
(B) oops, I checked and actually didn't realize it didn't look the way it should. Will fix that.
(C) I guess this would make sense. I'm pretty sure you will have many users like me coming to equinox as users of diffrax who don't have a strong background in JAX.
So, summing up I'll make the suggested edits and update the PR.

jaschau: I have added the fixes described above. Note that I made a small change to the pre-commit hooks as well to address https://github.com/psf/black/issues/2964.

patrick-kidger: Thanks! This basically LGTM; I'll handle merging this in a bit.
Regarding click -- lower down the thread you link, it looks like updating to Black 22.3.0 should fix things, instead of pinning click?

jaschau: Yeah, I was a bit hesitant about this, though, since the formatting rules might have changed in between versions. I've had this problem a couple of times in the past, causing reformatting of the entire codebase. But probably just worth a try...

PR #66, opened 2022-04-15
Added 2d Avg Max Pooling, foward working

### LINKED ISSUE #59: Pooling layer
Hi Patrick,

Just wondering if pooling layers in CNN are available in Equinox?

Thanks,
JW

Sam-Fatehmanesh:
This PR adds 2D Average and Max Pooling at `equinox/nn/pool.py`. I haven't been able to get grad working on pooling because I get the following error when I try to grad the loss function of a simple CNN using pooling. This is in the temporary test program `test_CNN_temp.py` in the tests folder.

```
NotImplementedError: Differentiation rule for 'reduce_window' not implemented
```

I also added a test for pooling to the `test_nn.py` file.

If anyone can help figure out how to resolve my issue, I would appreciate it.

Sorry in advance for any obvious mistakes or errors, this is my first real PR on any open source project.

Fixes #59

### DIFF
--- a/equinox/nn/pool.py
+++ b/equinox/nn/pool.py
@@ -1,0 +1,89 @@
+from typing import Sequence, Tuple, Union
+
+import jax.lax as lax
+import numpy as np
+
+from ..custom_types import Array
+from ..module import Module, static_field
+
+
+class Pool(Module):
+    operation: str = static_field()
+    num_spatial_dims: int = static_field()
+    kernel_size: Union[int, Sequence[int]] = static_field()
+    stride: Sequence[int] = static_field()
+    padding: Union[str, Sequence[Tuple[int, int]]] = static_field()
+
+    def __init__(
+        self,
+        operation: str,
+        num_spatial_dims: int,
+        kernel_size: Union[int, Sequence[int]],
+        stride: Sequence[int] = None,
+        padding: Union[str, Sequence[Tuple[int, int]]] = "SAME",
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+
+        self.operation = operation
+        self.num_spatial_dims = num_spatial_dims
+        self.kernel_size = kernel_size
+        self.stride = stride
+        if stride is None:
+            self.stride = self.kernel_size
+        self.padding = padding
+
+    def __call__(self, x: Array) -> Array:
+        if self.operation == "max":
+            output = lax.reduce_window(
+                x,
+                0.0,
+                lax.max,
+                (1,) + self.kernel_size,
+                (1,) + self.stride,
+                self.padding,
+            )
+            # print(x)
+        elif self.operation == "avg":
+            output = lax.reduce_window(
+                x,
+                0.0,
+                lax.add,
+                (1,) + self.kernel_size,
+                (1,) + self.stride,
+                self.padding,
+            ) / np.prod(self.kernel_size)
+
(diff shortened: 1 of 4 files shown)


equinox/nn/pool.py
```diff
@@ -0,0 +1,89 @@
...
+                lax.add,
+                (1,) + self.kernel_size,
+                (1,) + self.stride,
+                self.padding,
+            ) / np.prod(self.kernel_size)
+
+        return output
+
+
+class Pool2d_Avg(Pool):
```
patrick-kidger, line 60: The appropriate naming convention here would be `AvgPool2d`. (In particular it's generally common in Python to avoid using underscores in class names.)

I think if we do this we should probably add the full gamut of `{Avg,Max}Pool{1,2,3}d`.

equinox/nn/pool.py
```diff
@@ -0,0 +1,89 @@
...
+    def __init__(
+        self,
+        kernel_size,
+        stride=None,
+        padding: Union[str, Sequence[Tuple[int, int]]] = "SAME",
```
patrick-kidger, line 65: Make sure to pass through `**kwargs` as well.

tests/test_nn.py
```diff
@@ -645,3 +645,29 @@ def get_weights(m):
     conv = eqx.nn.Conv3d(5, 4, 3, key=getkey())
     conv = eqx.tree_at(lambda c: c.weight, conv, replace_fn=spectral)
     assert conv(jrandom.normal(getkey(), (5, 8, 8, 8))).shape == (4, 6, 6, 6)
+
+
+def test_pool(getkey):
+    Pool = eqx.nn.Pool("max", 2, (2, 2))
```
patrick-kidger, line 651: As a python convention, variable names (here the `Pool` on the left hand side) should usually be in `snake_case`.

tests/test_nn.py
```diff
@@ -645,3 +645,29 @@ def get_weights(m):
...
+    x = jrandom.normal(getkey(), (1, 15, 15))
+    assert AvgPool(x).shape == (1, 8, 8)
+
+    Pool = eqx.nn.Pool2d_Max((2, 2), padding="VALID")
+    x = jrandom.normal(getkey(), (1, 15, 15))
+    assert Pool(x).shape == (1, 7, 7)
+
+    AvgPool = eqx.nn.Pool("max", 2, (2, 2), padding="VALID")
+    x = jrandom.normal(getkey(), (1, 15, 15))
+    assert Pool(x).shape == (1, 7, 7)
```
patrick-kidger, line 673: Can there be a handful of tests validating that the correct numbers come out as well?

Start with `x = jnp.arange(...).reshape(...)` so there's no randomness, and hardcode the correct output to compare against.

patrick-kidger (commented):
Hey there! I've left a few comments on the PR. Overall it looks excellent; thanks a lot for the contribution.

Could you also add the new pooling layers to the documentation?

Regarding differentiability. Hmm, I can see that Flax has a comment saying that it is only "sometimes" differentiable, and there's also this issue on the main JAX page: https://github.com/google/jax/issues/7815 I've not dug into this myself yet, but as another point of reference can you try investigating to what extent pooling layers can be differentiated in Flax?

patrick-kidger:
Closing this in favour of #77.

PR #77, opened 2022-05-02
Added max and avg pool, alongside tests and docs

Benjamin-Walker:
I currently use einops.rearrange so that the pooling layers and conv layers take the same tensor format as input, BCHW. For some reason lax.conv_general_dilated and lax.reduce_window are BCHW and BHWC respectively. Can do this not using einops if you would prefer to not have the extra dependency.

### DIFF
--- a/equinox/nn/pool.py
+++ b/equinox/nn/pool.py
@@ -1,0 +1,259 @@
+from typing import Callable, Optional, Sequence, Tuple, Union
+
+import jax.lax as lax
+import jax.numpy as jnp
+import jax.random
+import numpy as np
+
+from ..custom_types import Array
+from ..module import Module, static_field
+
+
+class Pool(Module):
+    """General N-dimensional downsampling over a sliding window."""
+
+    init: Union[int, float, Array] = static_field()
+    operation: Callable[[Array, Array], Array]
+    num_spatial_dims: int = static_field()
+    kernel_size: Union[int, Sequence[int]] = static_field()
+    stride: Union[int, Sequence[int]] = static_field()
+    padding: Union[int, Sequence[int]] = static_field()
+
+    def __init__(
+        self,
+        init: Union[int, float, Array],
+        operation: Callable[[Array, Array], Array],
+        num_spatial_dims: int,
+        kernel_size: Union[int, Sequence[int]],
+        stride: Union[int, Sequence[int]] = 1,
+        padding: Union[int, Sequence[int], Sequence[Tuple[int, int]]] = 0,
+        **kwargs,
+    ):
+        """**Arguments:**
+        - `init': The initial value for the reduction.
+        - `operation`: The operation applied to the inputs of each window.
+        - `num_spatial_dims`: The number of spatial dimensions.
+        - `kernel_size`: The size of the convolutional kernel.
+        - `stride`: The stride of the convolution.
+        - `padding`: The amount of padding to apply before and after each
+            spatial dimension.
+
+        !!! info
+
+            In order for `Pool' to be differentiable, `operation(init, x)=x' needs to
+            be true for all finite `x'. For further details see
+            https://www.tensorflow.org/xla/operation_semantics#reducewindow  and
+            https://github.com/google/jax/issues/7718.
+
+        """
+        super().__init__(**kwargs)
+
+        self.operation = operation
+        self.init = init
+        self.num_spatial_dims = num_spatial_dims
+
+        if isinstance(kernel_size, int):
+            self.kernel_size = (kernel_size,) * num_spatial_dims
+        elif isinstance(kernel_size, Sequence):
+            self.kernel_size = kernel_size
+        else:
+            raise ValueError(
+                "`kernel_size` must either be an int or tuple of length "
+                f"{num_spatial_dims} containing ints."
+            )
+
+        if isinstance(stride, int):
+            self.stride = (stride,) * num_spatial_dims
+        elif isinstance(stride, Sequence):
+            self.stride = stride
+        else:
+            raise ValueError(
+                "`stride` must either be an int or tuple of length "
+                f"{num_spatial_dims} containing ints."
+            )
+
+        if isinstance(padding, int):
+            self.padding = tuple((padding, padding) for _ in range(num_spatial_dims))
+        elif isinstance(padding, Sequence) and all(
+            isinstance(element, Sequence) for element in padding
+        ):
+            self.padding = padding
+        elif isinstance(padding, Sequence) and len(padding) == num_spatial_dims:
+            self.padding = tuple((p, p) for p in padding)
+        else:
+            raise ValueError(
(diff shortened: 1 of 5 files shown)


equinox/nn/pool.py
```diff
@@ -0,0 +1,201 @@
...
+            x = x / np.prod(self.kernel_size)
+
+        x = jnp.squeeze(x, axis=0)
+        x = einops.rearrange(x, '... a -> a ...')
+        return x
+
+
+class AvgPool1D(Pool):
+    """As [`equinox.nn.Pool`][] with `operation=lax.add` and
+    `num_spatial_dims=1`."""
```
patrick-kidger, line 86: I can tell you've copied the style of the convolutional layers here, but I think this docstring is a bit opaque. Something like "downsamples by taking an average over a sliding window" might be a helpful one-sentence summary.

patrick-kidger (commented):
Yes, I'd prefer to avoid the additional dependency on einops.

Can you add some tests for the backward pass? As per #66 it seems like that's a potential issue we may need to try and resolve.

patrick-kidger:
Also: I think the new pooling documentation needs to be added to `mkdocs.yml`.

equinox/nn/pool.py
```diff
@@ -0,0 +1,201 @@
...
+    padding: Union[str, Sequence[Tuple[int, int]]] = static_field()
+    init: float = static_field()
+
+    def __init__(
+        self,
+        operation: Callable[[Array], Array],
+        num_spatial_dims: int,
+        kernel_size: Union[int, Sequence[int]],
+        strides: Sequence[int] = None,
+        padding: Union[str, Sequence[Tuple[int, int]]] = "VALID",
```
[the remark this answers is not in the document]
Benjamin-Walker: Sure, I will also include Sequence[Tuple[int, int]], as you can't have 'SAME' padding with only int and Sequence[int].

equinox/nn/pool.py
```diff
@@ -0,0 +1,238 @@
...
+import numpy as np
+
+from ..custom_types import Array
+from ..module import Module, static_field
+
+
+class Pool(Module):
+    """ General N-dimensional downsampling over a sliding window. """
+
+    operation: Callable[[Array], Array] = static_field()
```
patrick-kidger, line 16: This shouldn't be static. (In principle the operation may be learnt.) Also I think the annotation should be `Callable[[Array, Array], Array]`.

equinox/nn/pool.py
```diff
@@ -0,0 +1,238 @@
...
+    num_spatial_dims: int = static_field()
+    kernel_size: Union[int, Sequence[int]] = static_field()
+    stride: Union[int, Sequence[int]] = static_field()
+    padding: Union[int, Sequence[int]] = static_field()
+    init: float = static_field()
+
+    def __init__(
+        self,
+        operation: Callable[[Array], Array],
```
patrick-kidger, line 25: Likewise annotation here

equinox/nn/pool.py
```diff
@@ -0,0 +1,238 @@
...
+        - `stride`: The stride of the convolution.
+        - `padding`: The amount of padding to apply before and after each
+            spatial dimension.
+        """
+        super().__init__(**kwargs)
+
+        self.operation = operation
+        self.init = 0.0
+        if operation is lax.max:
+            self.init = -jnp.inf
```
patrick-kidger, line 45: Make `init` an input, don't try to guess based on special values. Essentially `Pool` should make sense as a standalone operation.

equinox/nn/pool.py
```diff
@@ -0,0 +1,238 @@
...
+        else:
+            raise ValueError(
+                "`stride` must either be an int or tuple of length "
+                f"{num_spatial_dims} containing ints."
+            )
+
+        if isinstance(padding, int):
+            self.padding = tuple((padding, padding) for _ in range(num_spatial_dims))
+        elif isinstance(padding, Sequence) and \
+            all(isinstance(element, tuple) for element in padding):
```
patrick-kidger, line 71: `isinstance(element, tuple)`
to
`isinstance(element, Sequence)`

tests/test_nn.py
```diff
@@ -6,6 +6,8 @@
 import jax.random as jrandom
 import pytest
 
+import sys
+sys.path.append('/Users/benwalker/PycharmProjects/equinox')
```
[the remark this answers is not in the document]
Benjamin-Walker: Hahahah oops, didn't mean to leave that one in

equinox/nn/pool.py
```diff
@@ -0,0 +1,245 @@
...
+    operation: Callable[[Array, Array], Array]
+    num_spatial_dims: int = static_field()
+    kernel_size: Union[int, Sequence[int]] = static_field()
+    stride: Union[int, Sequence[int]] = static_field()
+    padding: Union[int, Sequence[int]] = static_field()
+    init: float = static_field()
+
+    def __init__(
+        self,
+        init: int,
```
patrick-kidger, line 24: Nit: not necessarily an integer! e.g. `jnp.NINF` is a float.

Admittedly I'm not sure what the correct annotation is here. `Union[int, float]` maybe? Probably `Array`s are valid too?

patrick-kidger (commented):
Yep, good thought; add that somewhere in the documentation for `Pool`. Preferably with a few references to the relevant JAX issues etc.

Benjamin-Walker:
Fixed the remaining typesetting mistake

Benjamin-Walker:
I think it was actually 3.10 where they introduced using | instead of Union. Think I have squashed all the commits together, let me know if it hasn't worked.

tests/test_nn.py
```diff
@@ -653,3 +655,116 @@ def get_weights(m):
...
+    x = jnp.arange(64, dtype=jnp.float32).reshape(1, 4, 4, 4)
+    grad_fn = jax.value_and_grad(max_pool_mean)
+
+    grad_fn(x)
+
+
+def test_poolnetworkbackprop(getkey):
+    class CNN(eqx.Module):
+        conv_layer: list[Union[eqx.nn.Conv2d, eqx.nn.MaxPool2D]]
+        linear_layers: list[eqx.nn.Linear]
```
patrick-kidger, line 742: This test will still fail because `list[...]` is 3.9+; you need `typing.List[...]`.

Benjamin-Walker: Fixed now, sorry about all the little mistakes!

patrick-kidger: @Benjamin-Walker still failing!
This will go a lot faster [if you install pre-commit](https://github.com/patrick-kidger/equinox/blob/main/CONTRIBUTING.md).

Benjamin-Walker: Sorry I looked online for how to run pre-commits locally but I should have just read the docs. Hopefully all fine now, all the tests pass locally.

PR #91, opened 2022-05-19
fix two small typos in documentation

jvmncs:
This link in the `eqx.nn.Pooling` docstring was broken, and then I stumbled on a small typo in the contrib docs while fixing that


patrick-kidger:
Ah, good catches both. Thanks for the fixes!

PR #113, opened 2022-06-14
Fix pmap named axes

### LINKED ISSUE #111: Unbound axis name in filter_pmap + lax.pmean
I tried to parallelize my training code using the standard jax idiom of `pmap`ing the entire `update` function and performing `pmean` reduction on loss and gradients inside `update`. Unfortunately, I get an error about the mapped axis name being unbound. The same code works with `filter_vmap`, which I think is supposed to be API-equivalent, so I believe I am experiencing a bug rather than mere misuse. The reproducing code can be found [here](https://gist.github.com/jatentaki/779c949b1ff4e9ea1828fd7098fb0792).

jatentaki:
Fixes #111

### DIFF
--- a/equinox/vmap_pmap.py
+++ b/equinox/vmap_pmap.py
@@ -436,7 +436,6 @@ class _PmapWrapper(Module):
 @doc_strip_annotations
 def filter_pmap(
     fun: Callable = sentinel,
-    axis_name=None,
     *,
     default: AxisSpec = _zero_if_array_else_none,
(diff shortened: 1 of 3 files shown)


patrick-kidger:
LGTM! Can you bump the version number too? I'll do a new release.

jatentaki:
There you go

patrick-kidger:
Thanks for the contribution!

Issue #115, opened 2022-06-15
Lack of `out=` breaks use of `filter_pmap(fn)` when output of `fn` involves non-JAX types


jatentaki:
The repro is [here](https://colab.research.google.com/drive/101ZT_F2_iCzRnh85Z_3_3hD4zQfJrWci?usp=sharing). In short, functions which return PyTrees involving non-JAX types break `filter_{vmap,pmap}`. With `vmap` this can be fixed by specifying the `out=` kwarg, but it doesn't work for `pmap`. Is this just a TODO or a limitation of the "networks as callable PyTrees" model that equinox takes?

jatentaki:
I just realized this can be solved by making the `non_jax` with `equinox.static_field()`; however this is a technique I only learned from reading the source, I don't think it's discussed in the docs?

jatentaki:
In particular, `pmap` is broken for `eqx.nn.MLP` due to nonlinearities [not having been declared as static](https://github.com/patrick-kidger/equinox/blob/main/equinox/nn/composed.py#L30).

jatentaki:
Thanks for your reply, I think I understand your rationale for this design. Still, it sounds at odds with my mental picture of parallel training. This is my first attempt at that in JAX, but as I mentioned in the previous issue, I believe the way to [train on multiple GPUs](https://jax.readthedocs.io/en/latest/jax-101/06-parallelism.html#example) is to shard-replicate the model and optimizer state across GPUs, shard-split the training batch, insert `grad = jax.lax.pmean(grad, axis_name='device')` inside the `update` function and then `pmap` it. This means the example I linked above only differs from ensambling models by the `pmean` operations inside, which is obviously orthogonal to the issue we're discussing here. Therefore my use-case is indeed similar to ensambling models and since my `update` returns an entire neural network (with all non-array leaves it may have) I need to filter that.

What would be your approach to parallel training with `pmap`? To `pmap`-reduce the gradient computation, perform model update on a single GPU and then broadcast the updated model? I'm not very confident with equinox but I have the impression that I would also run into that kind of problems there.

Since making this issue I noticed that using `static_field` indeed breaks some things downstream, so I experimented with pre-computing a non-callable `out=` value. This should be possible since my output signature is known ahead of time, but for some reason JAX complains about structure mismatch...

```python
_0_if_array = lambda v: 0 if eqx.is_array(v) else None
mask = jax.tree_map(_0_if_array, v_model)
[...]

PR #118, opened 2022-06-16
Fix vmap with pytree out

jatentaki:
Partially addresses https://github.com/patrick-kidger/equinox/issues/115 by fixing a bug in filter_vmap to allow using a pytree as filter spec for the out parameter of filter_vmap. Adding tests for this case for both p- and vmap, just in case :)

Recreating this PR because with the previous I managed to introduce a merge conflict with myself...


patrick-kidger:
LGTM! Thanks for the (many!) fix(es).

PR #124, opened 2022-06-30
eqx.filter_{vmap,pmap}(out=...) not experimental

### LINKED ISSUE #115: Lack of `out=` breaks use of `filter_pmap(fn)` when output of `fn` involves non-JAX types

patrick-kidger:
1. Previously using a callable `out` parameter was experimental for
   `filter_vmap` -- because it monkey-patched JAX internals -- and
   unavailable for `filter_pmap`. It has now been updated to work for both,
   and using only the public JAX API. (Hurrah.)

2. Drive-by: Added `eqx.filter_eval_shape` as it looked like this was
   going to useful as part of implementing the previous feature. In the
   end this wasn't the case, but we get a new feature anyway.

3. Drive-by: Fixed a crash bug when filter-jit'ing a partial-wrapped
   function whilst using a PyTree as its filter spec (`fn`).

Closes #115.

CC @jatentaki WDYT?

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -2,8 +2,6 @@ name: Run tests
 
 on:
   pull_request:
-  schedule:
-    - cron: "0 2 * * 6"
 
 jobs:
   run-test:
--- a/docs/api/filtering/filtered-transformations.md
+++ b/docs/api/filtering/filtered-transformations.md
@@ -27,3 +27,7 @@ Practically speaking these are usually the only kind of filtering you ever have 
 ---
 
 ::: equinox.filter_pmap
+
+---
+
+::: equinox.filter_eval_shape
--- a/equinox/__init__.py
+++ b/equinox/__init__.py
@@ -1,4 +1,5 @@
 from . import experimental, nn
+from .eval_shape import filter_eval_shape
 from .filters import (
     combine,
     filter,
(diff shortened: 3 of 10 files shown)


jatentaki:
Looks good, I just tested with my code and it does the job :) Interesting, I also thought `filter_eval_shape` would be a necessary building block towards the solution.

patrick-kidger:
Excellent, I'm glad that this works for you.

By definition all types are PyTrees -- potentially they're just stumps of a single non-JAX-object. So I don't think there's distinction between "any Python object" and "pytress with potentially non-JAX leaves"?

jatentaki:
I guess, but it would not recurse into a non-pytree even if it holds jax arrays somewhere inside, which sounds like it may be implied by the wording. That's just a nit pick though.

patrick-kidger:
Ah I see. So that's the same as passing a JAX array in via closure. And I think that should be totally fine / won't break.

jatentaki:
In case you're still waiting for some input from me, LGTM :)

patrick-kidger:
Aha ty! Nope I just wasn't in any hurry to merge this -- I'm waiting on seeing how/if #126 plays out and then merging them both together.

patrick-kidger:
Merging this now as I've just discovered a need for `filter_eval_shape` in the upcoming version of Diffrax.

PR #129, opened 2022-07-08
Adaptive avg pool 1d

paganpasta:


equinox/nn/pool.py
```diff
@@ -256,3 +256,35 @@ def __init__(
...
+
+
+class AdaptiveAvgPool1d(Module):
+    """Adaptive 1D downsampling for a target shape."""
+
+    target_size: int = static_field()
+
+    def __init__(self, target_size: int):
+        """**Arguments:**
+        - `target_size`: The target output size.
```
patrick-kidger, line 268: Nit: this will need a blank line between these two lines, in order to render correctly in the docs.

equinox/nn/pool.py
```diff
@@ -256,3 +256,35 @@ def __init__(
...
+        """
+
+        self.target_size = target_size
+
+    def __call__(self, x: Array):
+        assert (
+            x.ndim == 1
+        ), f"Only supports 1D input, received input with {x.ndim} dimensions."
```
patrick-kidger, line 276: The convention in Equinox (and other Python projects) is for `assert` statements to be used for cases which should "never be hit". More precisely to indicate that the developer of the library has made a mistake somewhere, and that an assumed invariant isn't true.

In the case of user error, as here, then a `ValueError` is appropriate.

equinox/nn/pool.py
```diff
@@ -256,3 +256,35 @@ def __init__(
...
+        channels = jnp.size(x)
+        assert (
+            channels >= self.target_size
+        ), f"Final Pooled size {self.target_size} cannot be greater than input size {channels}."
+
+        splits = jnp.array_split(x, self.target_size)
+        num_head_arrays = channels % self.target_size
+        if num_head_arrays:
```
patrick-kidger, line 284: I recommend avoiding implicit conversion to bool. Can this be a check of the form `if num_head_arrays == 0`?

equinox/nn/pool.py
```diff
@@ -256,3 +256,35 @@ def __init__(
...
+            head_mean = jax.vmap(jnp.mean)(jnp.asarray(splits[:num_head_arrays]))
+            tail_mean = jax.vmap(jnp.mean)(jnp.asarray(splits[num_head_arrays:]))
```
patrick-kidger, line 286: This is going to be quite inefficient: `x` is split into several arrays (which can be done as views and is efficient), and then concatenated back together into two groups (in the `jnp.asarray`; and this isn't efficient).

I think it should be possible to first split `x` into just the head and tail piece (which can be done as views and is efficient), and then reshape each piece so that they have a leading dimension of the appropriate size (again, this is efficient), and then vmap'ing the mean operation over it.

In each case above, "efficient" = "not copying the underlying data".

patrick-kidger (commented):
Thanks for the PR!

I've left a few line-by-line comments below. Other than this:

- Would it be possible to add support for 2d and 3d pooling as well?
- And possibly also n-dimensional pooling? Although I think the number of pieces might scale exponentially with the dimension n. Not sure if there's a smart way around that. (Do you know how PyTorch handles this?)
- Can you add some tests verifying that this works as expected?
- Can you add this to the pooling documentation in `/docs/api/nn/pool.md`?

equinox/nn/pool.py
paganpasta: I am not sure if I understood it properly. At the moment, splits is a list of arrays. So I understood that you meant to get rid of it altogether and apply the vmap on the 2 parts of the input array. However, can you elaborate on how to use `view` to do this ? Alternatively, I have tried it with `reshape` and it is definitely faster than the older implementation.

Meanwhile, I tested the older implementation with the new one 
```
`new`: 5 loops, best of 5: 1.98 ms per loop
`old`: 5 loops, best of 5: 2.53 s per loop
```

paganpasta:
> Thanks for the PR!

No problem at all.


>     * Would it be possible to add support for 2d and 3d pooling as well?

Yes, but it will take me some time to work on it. I'll prioritize tests for 1D and documentation first.
 

>     * And possibly also n-dimensional pooling? Although I think the number of pieces might scale exponentially with the dimension n. Not sure if there's a smart way around that. (Do you know how PyTorch handles this?)

I'll take a look at this and start a discussion before working on the 2d and 3d implementations. 

>     * Can you add some tests verifying that this works as expected?

Yes, I'll do that.

>     * Can you add this to the pooling documentation in `/docs/api/nn/pool.md`?

Yes!

patrick-kidger:
Excellent!

Regarding higher dimensions: I think it might be possible to implement this by sequentially performing 1d averages over each dimension in turn?

I speculate that this isn't optimal speed-wise for smaller dimensions (as the operations are sequential, not in parallel), but is better for large dimensions (no exponential explosion in the number of pieces). In either case it would be easier to implement though.

paganpasta:
I tried the sequential 1D adaptive calls. For this, probably best to have functional form of the adaptive_avg_pool1d? 
Speed wise, pytorch adaptive1d is `100x` faster than the current eqx version (tried on colab). Any idea on how the speed can be boosted? For the 2D variants, speed difference is `4x`. 

Pytorch AdaptiveAvgPool1,2,3d are implemented as calls to their respective `F.` method. [link](https://pytorch.org/docs/stable/_modules/torch/nn/modules/pooling.html#AdaptiveAvgPool2d)

A general `Nd` adaptive pooling is a base class without a forward call.

docs/api/nn/pool.md
```diff
@@ -53,3 +53,25 @@
         members:
             - __init__
             - __call__
+
+---
+
+::: equinox.nn.adaptive_avg_pool1d
```
patrick-kidger, line 59: I think we can leave this out of the docs, it's really just a private helper function.

equinox/nn/pool.py
```diff
@@ -392,3 +392,106 @@ def __call__(
...
+                "`target_size` cannot be larger than the input channels."
+                f"Expected atleast {self.target_size} but received {channels}x{dims}."
+            )
+        target_channels, target_dims = self.target_size
+
+        x = jax.vmap(adaptive_avg_pool1d, in_axes=(0, None))(x, target_dims)
+        x = jax.vmap(adaptive_avg_pool1d, in_axes=(0, None))(
+            jnp.transpose(x), target_channels
+        )
+        return jnp.transpose(x)
```
patrick-kidger, line 497: I think the transposing can be skipped here, in favour of just setting `in_axes=(1, None)`?

patrick-kidger (commented):
- Yep, agreed that having a functional form for `adaptive_avg_pool1d` is a good way to go. In terms of it apparently being slower -- do you have a benchmark script you can share?

- PyTorch's calls: with a bit of spelunking through the C++ source code I found [this](https://github.com/pytorch/pytorch/blob/5dbec7c07c5eedd748fd56359c2d1b980dfa1037/aten/src/ATen/native/cpu/AdaptiveAvgPoolKernel.cpp#L16), which looks like the 2d implementation. Basically they're just manually iterating.

- `AvgPool1d` expecting 2d input: ack, that's annoying. That's a mistake on my part, I think. (FWIW I do think this might be slightly easier to use inside a `nn.Sequential` this way.)

  The adaptive and non-adaptive versions should definitely be consistent. If you want to make the adaptive ones accept n+1 dimensions that would be fine.
  
  I'd also be happy to go back and change the behaviour of the non-adaptive pooling, since it'd be easy to check for the having a dimension one higher, and then raising a nice error message warning a downstream user of the change in behaviour. (Equinox is still early enough in its life that I feel comfortable making backward-incompatible changes provided the change is loud.)
  
   Which do you think is best?
   
- Would it be reasonable to add an `AdaptivePool` base class, on top of which we implement `AdaptiveAvgPool1d`, `AdaptiveMaxPool2d` etc? I think this should be possible/straightforward, and then we'd have complete parity between adaptive and non-adaptive pooling.

paganpasta:
> * I think the functional form for `adaptive_avg_pool1d` is a great way to implement things. In terms of it apparently being slower -- do you have a benchmark script you can share?

I crudely compute the performance (ONLY CPU) using 
```
key = jax.random.PRNGKey(0)
arr = jax.random.uniform(key, (1,10000))
final_size = 500

adptv_pooler = AdaptiveAvgPool1D(final_size)#eqx

avg_pooler = nn.AvgPool1D(kernel_size=20, stride=20)#eqx

torch_pooler = torch.nn.AdaptiveAvgPool1d(final_size)#torch
torch_tensor = torch.tensor(np.asarray(arr))

%timeit -n 25 jax.vmap(pooler)(arr).block_until_ready()
%timeit -n 25 avg_pooler(arr).block_until_ready()
%timeit -n 25 torch_pooler(torch_tensor)

"""output
25 loops, best of 5: 1.86 ms per loop
25 loops, best of 5: 367 µs per loop
25 loops, best of 5: 16.1 µs per loop
"""

```

> * PyTorch's calls: with a bit of spelunking through the C++ source code I found [this](https://github.com/pytorch/pytorch/blob/5dbec7c07c5eedd748fd56359c2d1b980dfa1037/aten/src/ATen/native/cpu/AdaptiveAvgPoolKernel.cpp#L16), which looks like the 2d implementation. Basically they're just manually iterating.

Yes, eqx.AdaptivePoolAvg1D (version 0) was based on this. Like you mentioned it before, looping in python is perhaps not the best way to reproduce what pytorch does.

> * `AvgPool1d` expecting 2d input: ack, that's annoying. That's a mistake on my part, I think. (FWIW I do think this might be slightly easier to use inside a `nn.Sequential` this way, without needing to wrap in a vmap.) 
> ...  Which do you think is best?

Pytorch supports both 3D (batch, channels, dims) and 2D(channels, dims) inputs for [adaptive1D](https://pytorch.org/docs/stable/generated/torch.nn.AdaptiveAvgPool1d.html#torch.nn.AdaptiveAvgPool1d). Using this as a reference, it makes sense to support 2D input for adaptive1D. This becomes consistent with the other pooling methods implemented in equinox.     


> * Would it be reasonable to add an `AdaptivePool` base class, on top of which we implement `AdaptiveAvgPool1d`, `AdaptiveMaxPool2d` etc? I think this should be possible/straightforward, and then we'd have complete parity between adaptive and non-adaptive pooling.

Yes, it makes sense to have one.

equinox/nn/pool.py
paganpasta: Thanks for the suggestion! Still getting a hang of vmaps.

patrick-kidger:
Regarding the speed: you need to wrap the JAX operations inside a `jax.jit`, i.e. `jax.jit(jax.vmap(pooler))(arr).block_until_ready()`. The first time this is run this will be compiled. On subsequent runs this will then hit the compiled version of that function.

(There are three rules for using JAX that I wish I could emblazon across the sky: 1. have a single `jax.jit` (or `equinox.filter_jit`) decorator surrounding all your code. 2. don't use `lax.cond` unless you know what you're doing 3. don't use `.at[]` unless you know what you're doing.)

tests/test_nn.py
```diff
@@ -778,6 +778,15 @@ def test_avgpool1d():
     assert jnp.all(output == answer)
 
 
+def test_adaptivepool1d():
+    x = jnp.arange(14).reshape(1, 14)
+    adaptive_pool = eqx.nn.AdaptiveAvgPool1D(4)
+    output = adaptive_pool(x)
+    answer = jnp.array([[1.5, 5.5, 9.0, 12.0]])
```
paganpasta, line 785: **Inconsistency** 

For an input `(1, dim)`, other 1D methods return output of shape `(new_dim, )`. Adaptive1d returns `(1, new_dim)`.
What's the best output shape for this scenario?

patrick-kidger:
Okay, we're getting there! I think it should be possible to have an `AdaptivePool` base class rather than just `AdaptiveAvgPool`, by simply passing in the operation to be vmap'd inside `_adaptive_avg_pool1d`: `jnp.max` instead of `jnp.mean`, etc. This would then be consistent with the current pooling operations.

Likewise I don't think we need `_adaptive_avg_pool{2d,3d}` operations. Rather we should be able to do something like:

```python
class AdaptivePool(eqx.Module):
    target_shape: Tuple[int]
    operation: Callable

    def __call__(self, x, *, key=None):
        if x.ndim != len(target_shape):
            raise ValueError(...)
        for i, target_size in enumerate(self.target_shape):
            x = jax.vmap(_adaptive_pool1d, (i, None, None))(x, target_size, self.operation)
        return x

class AdaptiveAvgPool1d(eqx.Module):
    def __init__(self, target_size):
        self.target_shape = (target_size,)
        self.operation = jnp.mean

...  # avg-2d, avg-3d, max-1d, max-2d, max-3d
```

tests/test_nn.py
paganpasta: That's my mistake. I saw the test case: [link](https://github.com/patrick-kidger/equinox/blob/df5538f90c328100097a66eea1944051b2d684fc/tests/test_nn.py#L776) 
and the corresponding shape of the `answer`.

paganpasta:
I agree that a general `AdaptivePool` is much better and having `operation` also allows for easier extension to other pooling methods.
My concern is the `__call__` method of the `AdaptivePool`. Assuming that `_adaptive_avg_pool_1d` is the current implementation then the line below
```
for i, target_size in enumerate(self.target_shape):
            x = jax.vmap(_adaptive_pool1d, (i, None, None))(x, target_size, self.operation)
```
will only work for `AdaptiveAvgPool1d`. The `_adaptive_avg_1d` at the moment assumes the input to be `(dim, )` and all the others are built on top of it. That's why I added `_adaptive_{2d,3d}` to iterate over first `N` dimensions before calling `_adaptive_1d`.

patrick-kidger:
Ah right. Yeah I think what's needed is something like:

```python
for i, target_size in self.target_shape:
  op = _adaptive_pool1d
  for j in range(len(self.target_shape)):
    if i == j:
      continue
    op = jax.vmap(op, in_axes=(j, None, None))
  x = op(x)
```

paganpasta:
Alternatively, I think we can try to solve it recursively inside the `_adaptive_pool1d` with an extra `_adaptive_pool` method. Something like
```
def _adaptive_pool1d(x: Array, target_size: int, operation: Callable) -> Array:
    if x.ndim == 1:
        dims = jnp.size(x)
        num_head_arrays = dims % target_size
        if num_head_arrays != 0:
            head_end_index = num_head_arrays * (dims // target_size + 1)
            head_mean = jax.vmap(operation)(x[:head_end_index].reshape(num_head_arrays, -1))
            tail_mean = jax.vmap(operation)(
                x[head_end_index:].reshape(-1, dims // target_size)
            )
            mean = jnp.concatenate([head_mean, tail_mean])
        else:
            mean = jax.vmap(operation)(
                jax.vmap(operation)(x.reshape(-1, dims // target_size))
            )
        return mean
    else:
        return jax.vmap(_adaptive_pool1d, (0, None, None))(x, target_size, operation)


def _adaptive_pool(x: Array, target_shape: tuple, operation: Callable) -> Array:
    dims = x.ndim
    for i in range(dims):
            x = jnp.moveaxis(x, i, -1)
            x = jax.vmap(_adaptive_pool1d, (0, None, None), 0)(x, target_shape[i], operation)
            x = jnp.moveaxis(x, -1, i)
    return x
```
Then the base class can simply call the `_adaptive_pool` batched over `channels` and the methods can take care of batching over the `dims`?

patrick-kidger:
@paganpasta just checking whether this is ready for review or not?

patrick-kidger:
Alright, merged via #140. Thanks a lot for contributing -- I know we went back-and-forth on the design a bit!

PR #148, opened 2022-07-23
Fix GroupNorm channels argument and docstring

### LINKED ISSUE #127: `channels` may not be unspecified in `GroupNorm`, despite what the docs say
The [docs](https://github.com/patrick-kidger/equinox/blob/792484183969c4ea8632986f5e262b947b4276fa/equinox/nn/normalisation.py#L147) say that `channels` may be equal to `None` but [this](https://github.com/patrick-kidger/equinox/blob/792484183969c4ea8632986f5e262b947b4276fa/equinox/nn/normalisation.py#L152) and [this](https://github.com/patrick-kidger/equinox/blob/792484183969c4ea8632986f5e262b947b4276fa/equinox/nn/normalisation.py#L175) lines use the parameter in a way that breaks in that case.

I'm not attaching a proper repro, I'll probably just fix it myself when I find a moment.

jenkspt:
- Removes erroneous `shape` parameter from docstring
- Fixes #127 
- Adds tests for various argument combinations

### DIFF
--- a/equinox/nn/normalisation.py
+++ b/equinox/nn/normalisation.py
@@ -134,7 +134,7 @@ class GroupNorm(Module):
     def __init__(
         self,
         groups: int,
-        channels: Optional[int]=None,
+        channels: Optional[int] = None,
         eps: float = 1e-5,
         channelwise_affine: bool = True,
         **kwargs,
@@ -150,7 +150,9 @@ class GroupNorm(Module):
         if (channels is not None) and (channels % groups != 0):
             raise ValueError("The number of groups must divide the number of channels.")
         if (channels is None) and channelwise_affine:
-            raise ValueError("The number of channels should be specified if `channelwise_affine=True`")
+            raise ValueError(
+                "The number of channels should be specified if `channelwise_affine=True`"
+            )
         super().__init__(**kwargs)
         self.groups = groups
         self.channels = channels
--- a/equinox/nn/normalisation.py
+++ b/equinox/nn/normalisation.py
@@ -134,23 +134,23 @@ class GroupNorm(Module):
     def __init__(
         self,
         groups: int,
-        channels: int,
+        channels: Optional[int]=None,
         eps: float = 1e-5,
         channelwise_affine: bool = True,
         **kwargs,
     ):
         """**Arguments:**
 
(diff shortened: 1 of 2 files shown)


equinox/nn/normalisation.py
```diff
@@ -134,23 +134,23 @@ class GroupNorm(Module):
...
         - `groups`: The number of groups to split the input into.
         - `channels`: The number of input channels. May be left unspecified (e.g. just
             `None`) if `channelwise_affine=False`.
         - `eps`: Value added to denominator for numerical stability.
         - `channelwise_affine`: Whether the module has learnable affine parameters.
         """
-        if channels % groups != 0:
+        if (channels is not None) and (channels % groups != 0):
             raise ValueError("The number of groups must divide the number of channels.")
+        if (channels is None) and (channelwise_affine==True):
```
patrick-kidger, line 152: Nit: just `if (channels is None) and channelwise_affine` will suffice.

equinox/nn/normalisation.py
```diff
@@ -134,23 +134,23 @@ class GroupNorm(Module):
...
+            raise ValueError("The number of channels should be specified if `channel_wise_affine=True`")
```
patrick-kidger, line 153: Nit: `channel_wise_affine -> channelwise_affine`.

equinox/nn/normalisation.py
```diff
@@ -172,7 +172,8 @@ def __call__(
 
         A JAX array of shape `(channels, ...)`.
         """
-        y = x.reshape(self.groups, self.channels // self.groups, *x.shape[1:])
+        channels = x.shape[0] if self.channels is None else self.channels
```
patrick-kidger, line 175: I think just `channels = x.shape[0]` should suffice here?

patrick-kidger (commented):
This looks great -- thanks for getting involved. I've left a few nitty comments, and then I'd be very happy to merge this.

patrick-kidger:
Looks like the pre-commit checks are failing -- have a look at CONTRIBUTING.MD.

jenkspt:
My bad. Passed pre-commit, so should work now.

patrick-kidger:
Looks good! Thanks for contributing.

PR #155, opened 2022-07-27
Adds utils for calculating attention weights and multihead attention without projection layers.

### LINKED ISSUE #142: Feature Request: add attention utility functions like flax
It would be nice to split up the the [MultiheadAttention](https://github.com/patrick-kidger/equinox/blob/main/equinox/nn/attention.py#L14) Module code into utility functions for calculating attention weights and multihead attention without projection layers. Flax does this -- but would be nice to have it here as well.

[https://github.com/google/flax/blob/main/flax/linen/attention.py#L40-L187](https://github.com/google/flax/blob/main/flax/linen/attention.py#L40-L187)

anshradh:
Fixes #142


equinox/nn/attention.py
```diff
@@ -11,6 +11,183 @@
 from .linear import Linear
 
 
+def dot_product_attention_weights(
+    query: Array,
+    key_: Array,
+    bias: Optional[Array] = None,
+    mask: Optional[Array] = None,
+    dropout_p: float = 0.0,
```
patrick-kidger, line 19: For simplicity let's only have `dropout` and not `dropout_p`.

equinox/nn/attention.py
```diff
@@ -11,6 +11,183 @@
...
+    inference: Optional[bool] = None,
+    deterministic: Optional[bool] = None,
+):
+
+    r"""Computes dot-product attention weights given query and key.
+    Used by :func:`dot_product_attention`, which is what you'll most likely use.
+    But if you want access to the attention weights for introspection, then
+    you can directly call this function and call einsum yourself.
+    This also supports multi-query attention (https://arxiv.org/pdf/1911.02150).
+    **Arguments:**
```
patrick-kidger, line 32: Nit: new line before arguments,

equinox/nn/attention.py
```diff
@@ -11,6 +11,183 @@
...
+        - `dropout_p`: Dropout probability on attention weights.
+        - `dropout`: Already initialized Dropout module. Unused if `dropout_p != 0.`
+          (Keyword only argument).
+        - `key`: A `jax.random.PRNGKey` used for dropout. Unused if `dropout_p = 0.`.
+            (Keyword only argument.)
+        - `inference`: As [`equinox.nn.Dropout.__call__`][]. (Keyword only
+            argument.)
+        - `deterministic`: (Deprecated in favour of `inference`.)
+
+        **Returns:**
```
patrick-kidger, line 49: Nit: dedent

equinox/nn/attention.py
```diff
@@ -11,6 +11,183 @@
...
+
+        A JAX array of shape `(batch..., num_heads, query_seq_length, kv_seq_length)`.
```
patrick-kidger, line 51: The pattern in Equinox so far has been to not support arbitrary prepending dimensions, and let the end user `jax.vmap` if appropriate. Can this function be changed to match this convention? (i.e. have no batch dimension at all)

equinox/nn/attention.py
```diff
@@ -11,6 +11,183 @@
...
+        query.shape[-1],
+    )
+    kv_seq_length = key_.shape[-2] if key_.ndim < query.ndim else key_.shape[-3]
+
+    query = query / jnp.sqrt(depth)
+    if query.ndim == key_.ndim:
+        attn_weights = jnp.einsum("...qhd,...khd->...hqk", query, key_)
+    else:
+        # multi-query attention
+        attn_weights = jnp.einsum("...qhd,...kd->...hqk", query, key_)
```
patrick-kidger, line 76: I think multi-query attention is something you've added in here, i.e. it wasn't supported by the original implementation? I'm a bit skeptical about merging two different operations together like this. I think this might be possible to support more neatly by writing this function as operating on single-headed attention, and then either `vmap`'ing over both query and key (=normal multihead attention) or just `vmap`'ing over query (=multi-query attention)?

equinox/nn/attention.py
```diff
@@ -11,6 +11,183 @@
...
+        ), "q, k, v num_heads must match."
+        assert key_.shape[-3] == value.shape[-3], "k, v lengths must match"
+    elif query.ndim > key_.ndim and query.ndim > value.ndim:
+        # support for multi-query attention
+        assert (
+            query.shape[:-3] == key_.shape[:-2] == value.shape[:-2]
+        ), "q, k batch dims must match."
+        assert key_.shape[-2] == value.shape[-2], "k, v lengths must match"
+    else:
+        raise ValueError("q must have equal or more dimensions than k, v.")
```
patrick-kidger, line 170: Probably it won't matter after the simplification from the above comments anyway, but I think these checks are pretty much duplication the ones in `dot_product_attention_weights`, so things can be simplified.

anshradh: Yeah I simplified to only check that key and value shapes match (since query and key are checked in the weights function).

equinox/nn/attention.py
anshradh: Yeah it actually probably doesn't make sense to add support for multiquery attention in this PR - I can make a follow-up that adds it in so that this one only handles straightforward multihead attention utilities.

anshradh: Went ahead and removed this for now.

anshradh:
@patrick-kidger Gentle ping for another review when you have time :)

equinox/nn/attention.py
```diff
@@ -11,6 +11,146 @@
...
+        assert query.shape[-1] == key_.shape[-1], "q, k depths must match."
+    else:
+        raise ValueError("q must have equal dimensions to k.")
+
+    query_seq_length, num_heads, depth = query.shape
+    kv_seq_length = key_.shape[-3]
+
+    query = query / jnp.sqrt(depth)
+
+    attn_weights = jnp.einsum("...qhd,...khd->...hqk", query, key_)
```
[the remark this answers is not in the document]
patrick-kidger: (+`num_heads` dimensions as above)

equinox/nn/attention.py
```diff
@@ -160,6 +300,7 @@ def __init__(
         self.output_proj = Linear(
             num_heads * vo_size, output_size, use_bias=use_output_bias, key=okey
         )
+
```
patrick-kidger, line 303: Not that this change in and of itself matters, but can spurious changes be kept to a minimum?

patrick-kidger:
Closing in favour of #181.

PR #156, opened 2022-07-28
replace tree_* with tree_util.tree_* to avoid jax warning messages

amir-saadat:


patrick-kidger:
Looks like the pre-commit checks failed. See CONTRIBUTING.md.

I'd suggest doing `import jax.tree_util as jtu` for brevity.

amir-saadat:
sure.

docs/all-of-equinox.md
```diff
@@ -48,7 +49,7 @@ y = jax.random.normal(y_key, (100, 2))
 model = MyModule(model_key)
 grads = loss(model, x, y)
 learning_rate = 0.1
-model = jax.tree_map(lambda m, g: m - learning_rate * g, model, grads)
+model = jtu.tree_map(lambda m, g: m - learning_rate * g, model, grads)
 ```
```
patrick-kidger, line 53: Sorry, exception to the above: in the intro docs I just use `jax.whatever`, to avoid a long import list of `jax.random`, `jax.numpy` etc.

patrick-kidger (commented):
Only the one comment, otherwise LGTM.

docs/all-of-equinox.md
amir-saadat: sure, makes sense :).

Issue #102, opened 2022-06-04
Pull Requests - Activation Layers


conceptofmind:
Hi Patrick,

I wanted to get your thoughts on adding activation layers to Equinox's API. Similar to how Pytorch has torch.nn.ReLU().

For example a basic GELU class example:

```python
class GELU(Module):
    approximate: bool = static_field()

    def __init__(self, approximate: bool = False):
        super().__init__()
        self.approximate = approximate

    def __call__(self, x, *, key = None):
        if self.approximate:
            sqrt_2_over_pi = jnp.sqrt(2 / jnp.pi).astype(x.dtype)
            cdf = 0.5 * (1.0 + jnp.tanh(sqrt_2_over_pi * (x + 0.044715 * (x ** 3))))
            return x * cdf
        else: 
            return jnp.array(x * (lax.erf(x / jnp.sqrt(2)) + 1) / 2, dtype=x.dtype)
```

If this is something you think would be beneficial I can open up a pull request with added documentation.

Thank you,

Enrico

patrick-kidger:
Have discussed this separately. Posting the solution for anyone else coming across this:

```python
from typing import Callable
import equinox as eqx
import jax.nn as jnn

class Lambda(eqx.Module):
    fn: Callable

def __call__(self , x, *, key=None):
        return self.fn(x)

...

eqx.nn.Sequential([
    eqx.nn.Linear(...),
    Lambda(jnn.gelu),
    eqx.nn.Linear(...),
])
```

mfouesneau:
I followed your example and when I try
`loss, grads = value_and_grad(loss_fn)(model, x)`
I get a TypeError:
`TypeError: Argument '<function gelu at 0x7b51a1fd23b0>' of type <class 'function'> is not a valid JAX type.`

(I can evaluate the `loss_fn(model, x)` without error)

patrick-kidger:
If your `model` includes non-JAX-types (does it?) then try either using `eqx.filter_value_and_grad`, or partitioning/combining. See [here](https://docs.kidger.site/equinox/all-of-equinox/#2-filtering).

mfouesneau:
yes thanks. answered in  #706

PR #158, opened 2022-07-28
added lambda_layer to composites

paganpasta:
- Added Lambda layer to wrap any callable for `nn.Sequential`. (courtesy of https://github.com/patrick-kidger/equinox/issues/102#issuecomment-1148330393)


patrick-kidger:
Can you include in `Lambda.__doc__` a bit of explanation for why this exists? And in particular make clear that it's only useful for use with `Sequential`.

Conversely can you also add a reference to this in `Sequential.__doc__`?

patrick-kidger:
Alright, merged!

Issue #136, opened 2022-07-15
Recommended way to partially load weights into a model ?


paganpasta:
Hi,

Is there a recommended way to partially load weights to a new model ? 

At the moment I have an inept solution which uses the fact that `fc` attribute is the only `eqx.nn.Linear` instance in my model

```
is_leaf = lambda x : True if isinstance(x, eqx.nn.Linear) else False
new_model = tree_deserialise_leaves('model.eqx', CNN(num_classes=26, key=key), is_leaf=is_leaf)
```
This approach won't work where there are multiple `Linear` layers. Is there a recommended way to perform such loading? 

A `name` parameter for modules might be useful here.

patrick-kidger:
Yep -- the `filter_spec` argument to `tree_deserialise_leaves` may be a PyTree that is a prefix of the one being loaded, and so you can pass a custom `filter_spec` that does/doesn't load the weights for each part of the model.

Admittedly this does mean you probably want to end up constructing a PyTree with the current `_default_deserialise_filter_spec` on the leaves where you want to load stuff, which is of course a private function. Having ways to more easily extend this default, rather than just overwriting it entirely, would be good.

paganpasta:
Thanks for the suggestion.

Possibly a method which works on a  `PyTree`, `list of leaves` to be ignored and `list of overrides` can be used by the user for deserialization/serialization ?

paganpasta:
@patrick-kidger  having a `strict=True/False` like functionality akin to Pytorch's `load_state_dict` can also be useful in loading partial weights which returns the mismatched leaves along with partially loaded model?

patrick-kidger:
Yeah, that sounds reasonable. Although PyTorch and Equinox serialise in pretty different ways, I'm not sure what a partial match might mean here. That one PyTree is a prefix of the other?

paganpasta:
Yes.

patrick-kidger:
This is surprising tricky. I've not managed to figure out a way to obtain the largest common prefix, using just JAX's public API.

paganpasta:
I see, thanks. 
Happy to work on `strict` and/or extending `_default_deserialise_filter_spec` if you see the advantage of adding them.

patrick-kidger:
Yep, go for it. I think both partial deserialisation and user extensibility are worth handling.

See also the discussion in https://github.com/google/jax/issues/11210 for context.

[...]

PR #165, opened 2022-08-02
Ordered tree map

paganpasta:
Added as per discussion https://github.com/patrick-kidger/equinox/issues/136
- Ordered Tree map
- Doc changes and test cases

### DIFF
--- a/equinox/__init__.py
+++ b/equinox/__init__.py
@@ -19,7 +19,7 @@ from .serialisation import (
     tree_deserialise_leaves,
     tree_serialise_leaves,
 )
-from .tree import tree_at, tree_equal, tree_inference
+from .tree import _ordered_tree_map, tree_at, tree_equal, tree_inference
 from .update import apply_updates
 from .vmap_pmap import filter_pmap, filter_vmap
 
--- a/equinox/serialisation.py
+++ b/equinox/serialisation.py
@@ -2,11 +2,11 @@ import pathlib
 from typing import Any, BinaryIO, Callable, Union
 
 import jax.numpy as jnp
-import jax.tree_util as jtu
 import numpy as np
 
 from . import experimental
 from .custom_types import PyTree
+from .tree import _ordered_tree_map
 
 
 def default_serialise_filter_spec(f: BinaryIO, x: Any) -> None:
@@ -23,7 +23,7 @@ def default_serialise_filter_spec(f: BinaryIO, x: Any) -> None:
 
     !!! info
 
-        This function can be extended to customise the serialization behaviour for leaves.
+        This function can be extended to customise the serialisation behaviour for leaves.
 
     !!! example
 
@@ -67,7 +67,7 @@ def default_deserialise_filter_spec(f: BinaryIO, x: Any) -> Any:
 
     !!! info
 
-        This function can be extended to customise the serialization behaviour for leaves.
+        This function can be extended to customise the deserialisation behaviour for leaves.
 
     !!! example
 
@@ -178,9 +178,9 @@ def tree_serialise_leaves(
             def __serialise(y):
                 spec(f, y)
 
-            jtu.tree_map(__serialise, x, is_leaf=is_leaf)
+            _ordered_tree_map(__serialise, x, is_leaf=is_leaf)
 
-        jtu.tree_map(_serialise, filter_spec, pytree)
+        _ordered_tree_map(_serialise, filter_spec, pytree)
 
 
 def tree_deserialise_leaves(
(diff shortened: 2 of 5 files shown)


equinox/__init__.py
```diff
@@ -19,7 +19,7 @@
     tree_deserialise_leaves,
     tree_serialise_leaves,
 )
-from .tree import tree_at, tree_equal, tree_inference
+from .tree import _ordered_tree_map, tree_at, tree_equal, tree_inference
```
patrick-kidger, line 22: Let's not expose `_ordered_tree_map` publicly. (The underscore indicates this.)

equinox/serialisation.py
```diff
@@ -2,11 +2,11 @@
 from typing import Any, BinaryIO, Callable, Union
 
 import jax.numpy as jnp
-import jax.tree_util as jtu
 import numpy as np
 
 from . import experimental
 from .custom_types import PyTree
+from .tree import _ordered_tree_map
```
patrick-kidger, line 9: Likewise we shouldn't be importing anything private (with the underscore). This indicates that it's private to the module that defined it.

equinox/serialisation.py
```diff
@@ -194,17 +194,16 @@ def tree_deserialise_leaves(
     **Arguments:**
 
     - `path`: The file location to load values from.
-    - `like`: A PyTree of the same structure, and with leaves of the same type, as the
-        PyTree being loaded. Those leaves which are loaded will replace the
-        corresponding leaves of `like`.
+    - `like`: A PyTree of same, prefix or compatible-partial structure
+        of the saved PyTree. The matching leaves are required to be of the same type.
+        Those leaves which are loaded will replace the corresponding leaves of `like`.
```
patrick-kidger, line 199: I seem to recall from #136 that we *can't* deserialise prefixes at all. Indeed we iterate over `like` in a depth-first order, which I would have thought means we end up deserialising leaves in the wrong order.

I note that the test and example you've written for this isn't a PyTree prefix -- it's a PyTree of the same structure (an `eqx.nn.MLP` with leaves of a different shape (the weight/bias for the final layer).

equinox/serialisation.py
```diff
@@ -218,11 +217,17 @@ def tree_deserialise_leaves(
...
-        model = eqx.nn.MLP(2, 2, 2, 2, key=jr.PRNGKey(0))
-        eqx.tree_serialise_leaves("some_filename.eqx", model)
-        model2 = eqx.tree_deserialise_leaves("some_filename.eqx", model)
-        ```
+        model_original = eqx.nn.MLP(2, 2, 2, 2, key=jr.PRNGKey(0))
+        eqx.tree_serialise_leaves("some_filename.eqx", model_original)
+        model_loaded = eqx.tree_deserialise_leaves("some_filename.eqx", model_original)
 
+        # (Recommended) To partially load weights
+        model_partial = eqx.tree_at(lambda mlp: mlp.layers[-1], model_loaded, model_original)
```
patrick-kidger, line 225: I like this example.

equinox/serialisation.py
```diff
@@ -240,8 +246,8 @@ def _deserialise(spec, x):
             def __deserialise(y):
                 return spec(f, y)
 
-            return jtu.tree_map(__deserialise, x, is_leaf=is_leaf)
+            return _ordered_tree_map(__deserialise, x, is_leaf=is_leaf)
 
-        out = jtu.tree_map(_deserialise, filter_spec, like)
-    jtu.tree_map(_assert_same, out, like, is_leaf=is_leaf)
+        out = _ordered_tree_map(_deserialise, filter_spec, like)
+    _ordered_tree_map(_assert_same, out, like, is_leaf=is_leaf)
```
patrick-kidger, line 252: We don't need the `_assert_same` to occur in a guaranteed order. Not that it really matters, but it's not a necessary change.

equinox/tree.py
```diff
@@ -304,3 +304,37 @@ def where(pytree):
...
     return tree_at(_inferences, pytree, replace_fn=lambda _: value)
+
+
+def _ordered_tree_map(
+    f: Callable[..., Any],
+    tree: Any,
+    *rest: Any,
+    is_leaf: Optional[Callable[[Any], bool]] = None
+) -> Any:
+    """In an orderly fashion, maps a multi-input function over pytree args to produce a new pytree.
```
patrick-kidger, line 315: I think this docstring is way too unwieldy. Something like "Like `jax.tree_util.tree_map`, but guaranteed to iterate over the tree in fixed order. (Namely depth-first left-to-right.)"

paganpasta:
Umm should I reopen one with required changes?

patrick-kidger:
Looks like GitHub auto-closed this because the `dev` branch got merged. Yes, please reopen!

equinox/serialisation.py
paganpasta: But then while deserialising it would become an issue if leaves are saved in arbitrary order but accessed in DFS?

equinox/serialisation.py
paganpasta: Yes, my bad. I'll try with the Prefix but I think it won't work, like you said. 
For the MLP I wanted to check the `DFS` matching substructure bit.

equinox/serialisation.py
paganpasta: Sorry ignore this.

patrick-kidger: Indeed, the `out = _ordered...` line is correct. It's the line afterwards (`_ordered_tree_map(_assert_same, ...`) that doesn't need to be ordered.

Like I say this doesn't really matter, but IMO it's a bit neater not to make an unnecessary change.

equinox/serialisation.py
paganpasta: Defining it now in `serialisation.py`.

equinox/serialisation.py
paganpasta: Should I remove the alternative one? Just wanted to somehow capture the DFS sub-structure bit for partial loading.

patrick-kidger: yep

PR #174, opened 2022-08-08
Add len and iter methods to nn.Sequential

jenkspt:
Adds the `__len__` and `__iter__` methods to `eqx.nn.Sequential`

It isn't that hard to write `len(seq.layers)` or `zip(seq.layers, ...)`, but this does make it just slightly cleaner and also mimics the behavior of pytorch Sequential module.


patrick-kidger:
LGTM!

Issue #175, opened 2022-08-09
Adding ceil_mode to nn.Pool?


paganpasta:
Hi,

Is it possible to support torch's `ceil_mode` like functionality in Pooling? 
There is a one sided padding for each dimension as a result of using `ceil` instead of `floor`.
 
Thanks.

patrick-kidger:
Sounds good to me. As usual, I'd be happy to accept a PR on this.

PR #176, opened 2022-08-10
added use_ceil to pooling

paganpasta:
Corresponding issue: https://github.com/patrick-kidger/equinox/issues/175

- Added `use_ceil` parameter to `Pool`
- Added a check to make sure garbage values are not pooled if `padding > kernel/2`.

### DIFF
--- a/equinox/nn/pool.py
+++ b/equinox/nn/pool.py
@@ -18,6 +18,7 @@ class Pool(Module):
     kernel_size: Union[int, Sequence[int]] = static_field()
     stride: Union[int, Sequence[int]] = static_field()
     padding: Union[int, Sequence[int], Sequence[Tuple[int, int]]] = static_field()
+    use_ceil: bool = static_field()
 
     def __init__(
         self,
@@ -27,6 +28,7 @@ class Pool(Module):
         kernel_size: Union[int, Sequence[int]],
         stride: Union[int, Sequence[int]] = 1,
         padding: Union[int, Sequence[int], Sequence[Tuple[int, int]]] = 0,
+        use_ceil: bool = False,
         **kwargs,
     ):
         """**Arguments:**
@@ -38,6 +40,9 @@ class Pool(Module):
         - `stride`: The stride of the convolution.
         - `padding`: The amount of padding to apply before and after each
             spatial dimension.
+        - `use_ceil`: If `True`, then `ceil` is used to compute the final output
+            shape instead of `floor`. For `ceil`, if required, extra padding is added.
+            Defaults to `False`.
 
         !!! info
 
@@ -51,6 +56,7 @@ class Pool(Module):
         self.operation = operation
         self.init = init
         self.num_spatial_dims = num_spatial_dims
+        self.use_ceil = use_ceil
 
         if isinstance(kernel_size, int):
             self.kernel_size = (kernel_size,) * num_spatial_dims
@@ -85,6 +91,26 @@ class Pool(Module):
                 f"{num_spatial_dims} containing ints or tuples of length 2."
             )
 
+        for padding, kernel_size in zip(self.padding, self.kernel_size):
+            if kernel_size / 2 < padding[0] or kernel_size / 2 < padding[1]:
+                raise ValueError(
+                    "Padding should be less than or equal to half of the kernel size."
+                    f" Received kernel_size {kernel_size} and padding {padding}."
+                )
+
+    def _update_padding_for_ceil(self, input_shape):
+        new_padding = []
+        total_padding = np.sum(self.padding, axis=1)
+        for idx, (input_dim, padding, kernel_size, stride) in enumerate(
+            zip(input_shape[1:], total_padding, self.kernel_size, self.stride)
+        ):
+            target_out_dim = np.ceil(
+                (input_dim + padding - 1 * (kernel_size - 1) - 1) / stride + 1
+            ).item()
+            padding = int((target_out_dim - 1) * stride + kernel_size - input_dim)
+            new_padding.append((self.padding[idx][0], padding - self.padding[idx][0]))
+        return tuple(new_padding)
+
     def __call__(
         self, x: Array, *, key: Optional["jax.random.PRNGKey"] = None
     ) -> Array:
@@ -104,6 +130,11 @@ class Pool(Module):
             f"but input has shape {x.shape}"
         )
 
+        if self.use_ceil:
+            padding = self._update_padding_for_ceil(x.shape)
+        else:
+            padding = self.padding
+
         x = jnp.moveaxis(x, 0, -1)
         x = jnp.expand_dims(x, axis=0)
         x = lax.reduce_window(
@@ -112,7 +143,7 @@ class Pool(Module):
             self.operation,
             (1,) + self.kernel_size + (1,),
             (1,) + self.stride + (1,),
-            ((0, 0),) + self.padding + ((0, 0),),
+            ((0, 0),) + padding + ((0, 0),),
         )
 
         x = jnp.squeeze(x, axis=0)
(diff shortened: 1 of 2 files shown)


paganpasta:
Saw that `dilation` is not supported at the moment. However, decided against pushing the change here. Will open a new request corresponding to it in few days.

equinox/nn/pool.py
```diff
@@ -27,6 +28,7 @@ def __init__(
         kernel_size: Union[int, Sequence[int]],
         stride: Union[int, Sequence[int]] = 1,
         padding: Union[int, Sequence[int], Sequence[Tuple[int, int]]] = 0,
+        use_ceil: bool = False,
```
patrick-kidger, line 31: I think for readability's sake this argument needs to be propagated through to all of `MaxPool2d` etc., so that it's in their signature and docstring. (At the moment it's just being passed through as a `**kwarg`.)

equinox/nn/pool.py
```diff
@@ -85,6 +91,26 @@ def __init__(
                 f"{num_spatial_dims} containing ints or tuples of length 2."
             )
 
+        for padding, kernel_size in zip(self.padding, self.kernel_size):
+            if kernel_size / 2 < padding[0] or kernel_size / 2 < padding[1]:
+                raise ValueError(
+                    "Padding should be less than or equal to half of the kernel size."
+                    f" Received kernel_size {kernel_size} and padding {padding}."
```
patrick-kidger, line 98: I don't think this check is technically necessary. If someone really wanted they might want to add loads of padding, knowing they'd just get zero out.

equinox/nn/pool.py
```diff
@@ -18,6 +18,7 @@ class Pool(Module):
     kernel_size: Union[int, Sequence[int]] = static_field()
     stride: Union[int, Sequence[int]] = static_field()
     padding: Union[int, Sequence[int], Sequence[Tuple[int, int]]] = static_field()
+    use_ceil: bool = static_field()
```
patrick-kidger, line 21: In passing, I've just noticed that some of these type annotations are too broad. It should be:
```python
kernel_size: Sequence[int]
stride: Sequence[int]
padding: Sequence[Tuple[int, int]]
```
since we convert them from these other forms during `__init__`.

equinox/nn/pool.py
```diff
@@ -85,6 +91,26 @@ def __init__(
...
+        total_padding = np.sum(self.padding, axis=1)
+        for idx, (input_dim, padding, kernel_size, stride) in enumerate(
+            zip(input_shape[1:], total_padding, self.kernel_size, self.stride)
+        ):
+            target_out_dim = np.ceil(
+                (input_dim + padding - 1 * (kernel_size - 1) - 1) / stride + 1
+            ).item()
+            padding = int((target_out_dim - 1) * stride + kernel_size - input_dim)
+            new_padding.append((self.padding[idx][0], padding - self.padding[idx][0]))
+        return tuple(new_padding)
```
patrick-kidger, line 112: This feels overly-complicated. If `k % n == 0` then `floor(n/k) == ceil(n/k)`, otherwise `floor(n/k) + 1 == ceil(n/k)`.

Thus I think something like this should suffice:
```python
def _update_padding_for_ceil(self, input_shape):
  new_padding = []
  for input_size, (left_padding, right_padding), kernel_size, stride in zip(
                                    input_shape, self.padding, self.kernel_size, self.stride):
    if stride % (input_size + left_padding + right_padding - kernel_size + 1) == 0:
      # the sliding kernel exactly matches up with the edge of the input, with no gaps,
      # if there exists x such that
      # x * stride + kernel_size -1 == input_size + left_padding + right_padding
      new_padding.append((left_padding, right_padding))
    else:
      new_padding.append((left_padding, right_padding + stride))
  return tuple(new_padding)
```
I think?

equinox/nn/pool.py
paganpasta: I see. With a case where `padding == kernel`, I was getting garbage values at the start and sometimes at the end. Do you think raising a warning here would be useful ? 
```python
x = jnp.arange(14).reshape(1, 14)
max_pool = eqx.nn.MaxPool1d(3, 3, padding=3)
print(max_pool(x))

#[[-2147483648 , 2 , 5 , 8, 11, 13]]
```

equinox/nn/pool.py
paganpasta: > This feels overly-complicated. If k % n == 0 then floor(n/k) == ceil(n/k), otherwise floor(n/k) + 1 == ceil(n/k)

The solution was also based on it, but yeah yours does look simpler.

paganpasta: Wanted to cross-check some corrections to your solutions. 

- A minor correction: `input_shape` in `for` should be `input_shape[1:]` but can be ignored as it can be handled by calling the function with the required values
- `stride % (input_size + left_padding + right_padding - kernel_size + 1)` should be flipped around and `1` should be dropped. This is based on the formulae mentioned in [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.MaxPool1d.html?highlight=maxpool#torch.nn.MaxPool1d). 
- Shouldn't the new padding to be added be something like `right_padding + stride - (remainder from the % operation)`?

patrick-kidger: - Agreed
- Would need to think about the `1`; otherwise agreed;
- I don't think it matters. We don't need to perfectly align the sliding window with the boundary of the input; we just need to add enough padding that we take one more step.

equinox/nn/pool.py
patrick-kidger: So the "garbage values" are the identity value for the operation in question. At a conceptual level it makes sense that they're there.

On a practical level though, I can believe that this would probably surprise most folks, and may be a silent source of bugs for them. So I'm happy to restrict padding if you prefer.

equinox/nn/pool.py
paganpasta: 3) Agreed, but adding additional padding than required can potentially output `identity` values in the output as per the discussion above.
I'm in favour of suppressing the padding and also matching the correct padding `right_padding + stride - (remainder from the % operation)` . Thoughts?

patrick-kidger: So I'm saying that I don't think "the remainder from the % operation" shouldn't affect things.

By adding on `stride` extra padding here, we should obtain precisely one more step, of length `stride`, than before?

equinox/nn/pool.py
patrick-kidger: If we're keeping this check, should it actually be `if kernel_size > left_padding + right_padding`?

patrick-kidger: Or actually perhaps `if kernel_size >= max(left_padding, right_padding)`?

Since as long as the padding on each side is smaller than the window, eaxh window must necessarily include at least one non-padding value.

paganpasta: > `kernel_size >= max(left_padding, right_padding)`

This seems like a reasonable check.

equinox/nn/pool.py
paganpasta: I was thinking of a scenario when `kernel_size < right_padding + stride` as many times `kernel_size = stride` is used. 
However, I think the failure will happen regardless of using `stride` or `remainder`. This will happen when with `left_padding`, the `kernel` and `stride` exactly covers the input but `right _padding + _pad > kernel_size`. We will have identity values introduced into the computation. 

Example: 
```
x = jnp.arange(14).reshape(1, 14)
max_pool = eqx.nn.MaxPool1d(kernel_size=3, stride=3, padding=1, use_ceil=True)
max_pool(x)  # [[ 1, 4, 7, 10, 13, -2147483648 ]]
```

paganpasta: Maybe we can simply use `+stride` and simply raise a warning at runtime with a check to let the user know that `identity` have crept in with `ceil`?

patrick-kidger: Perhaps let's move the current is-padding-sensible check to after `use_ceil` is resolved here, and raise a `RuntimeError` if it fails.

patrick-kidger:
Alright, LGTM! I'll merge this along with all the other ready-to-go PRs and do a new release shortly.

PR #181, opened 2022-08-22
Add attention functions and tests

jenkspt:
Adds `dot_product_attention_weights` and `dot_product_attention` functions and tests

**Design considerations**:
- `dot_product_attention` and `dot_product_attention_weights` don't take multi-head inputs -- instead attention heads are vmap'd over in `MultiheadAttention`. This allows for greater flexibility when creating other types of attention modules
- To simplify the `dot_product_attention` signature -- `dropout_fn` is added as a single argument callable, which should close over the dropout arguments like `key` and `inference`. The alternative I think would be to add a functional version of dropout and add its arguments to `dot_product_attention`, however this would make changing the dropout rate after initializing the module less intuitive -- since dropout rate would have to be an attribute of `MultiheadAttention`.
- `mask` shape check is kept inside `dot_product_attention_weights`. The downside to this is that errors raised inside vmap'd functions are less obvious -- i.e. if the heads don't match then `vmap` function will raise an error. The alternative is to pull the shape check out and put it back in `MultiheadAttention`

### DIFF
--- a/equinox/nn/attention.py
+++ b/equinox/nn/attention.py
@@ -40,7 +40,7 @@ def dot_product_attention(
 ) -> Array["query_seq_length", "value_size"]:  # noqa: F821
 
     weights = dot_product_attention_weights(query, key, mask)
-    if dropout_fn:
+    if dropout_fn is not None:
         weights = dropout_fn(weights)
     attn = jnp.einsum("sS,Sd->sd", weights, value)
     return attn
@@ -259,7 +259,7 @@ class MultiheadAttention(Module):
         )
         # vmap over attention heads
         in_axes = (1, 1, 1, None if mask is None else 1, None)
-        attn = jax.vmap(dot_product_attention, in_axes)(
+        attn = jax.vmap(dot_product_attention, in_axes, out_axes=1)(
             query_heads, key_heads, value_heads, mask, dropout_fn
         )
         attn = attn.reshape(query_seq_length, -1)
--- a/equinox/nn/attention.py
+++ b/equinox/nn/attention.py
@@ -1,5 +1,5 @@
 import math
-from typing import Optional
+from typing import Callable, Optional
 
 import jax
 import jax.numpy as jnp
@@ -9,6 +9,41 @@ from ..custom_types import Array
 from ..module import Module, static_field
 from .dropout import Dropout
 from .linear import Linear
+
+
+def dot_product_attention_weights(
+    query: Array["query_seq_length", "qk_size"],  # noqa: F821
+    key: Array["kv_seq_length", "qk_size"],  # noqa: F821
+    mask: Optional[Array["query_seq_length", "kv_seq_length"]] = None,  # noqa: F821
+) -> Array["query_seq_length", "kv_seq_length"]:  # noqa: F821
+
+    logits = jnp.einsum("sd,Sd->sS", query, key)
+    logits = logits / math.sqrt(query.shape[-1])
+    if mask is not None:
+        if mask.shape != logits.shape:
+            raise ValueError(
+                f"mask must have shape (query_seq_length, "
+                f"kv_seq_length)=({query.shape[0]}, "
+                f"{key.shape[0]}). Got {mask.shape}."
+            )
+        logits = jnp.where(mask, logits, -jnp.inf)
+
+    return jax.nn.softmax(logits, axis=-1)
+
+
+def dot_product_attention(
+    query: Array["query_seq_length", "qk_size"],  # noqa: F821
+    key: Array["kv_seq_length", "qk_size"],  # noqa: F821
+    value: Array["kv_seq_length", "value_size"],  # noqa: F821
+    mask: Optional[Array["query_seq_length", "kv_seq_length"]] = None,  # noqa: F821
+    dropout_fn: Optional[Callable[[Array], Array]] = None,
+) -> Array["query_seq_length", "value_size"]:  # noqa: F821
+
+    weights = dot_product_attention_weights(query, key, mask)
+    if dropout_fn:
+        weights = dropout_fn(weights)
+    attn = jnp.einsum("sS,Sd->sd", weights, value)
+    return attn
 
 
 class MultiheadAttention(Module):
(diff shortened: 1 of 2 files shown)


equinox/nn/attention.py
```diff
@@ -11,6 +11,41 @@
...
+def dot_product_attention(
+    query: Array["query_seq_length", "qk_size"],  # noqa: F821
+    key: Array["kv_seq_length", "qk_size"],  # noqa: F821
+    value: Array["kv_seq_length", "value_size"],  # noqa: F821
+    mask: Optional[Array["query_seq_length", "kv_seq_length"]] = None,  # noqa: F821
+    dropout_fn: Optional[Callable[[Array], Array]] = None,
+) -> Array["query_seq_length", "value_size"]:  # noqa: F821
+
+    weights = dot_product_attention_weights(query, key, mask)
+    if dropout_fn:
```
patrick-kidger, line 43: Can this be `if dropout_fn is not None`? (I'm not a fan of implicit-cast-to-bool.)

equinox/nn/attention.py
```diff
@@ -219,22 +254,14 @@ def __call__(
...
-            logits = jnp.where(mask, logits, -jnp.inf)
-
-        weights = jax.nn.softmax(logits, axis=-1)
-        weights = self.dropout(
+        dropout_fn = lambda weights: self.dropout(
             weights, key=key, inference=inference, deterministic=deterministic
         )
-        attn = jnp.einsum("hsS,Shd->shd", weights, value_heads)
+        # vmap over attention heads
+        in_axes = (1, 1, 1, None if mask is None else 1, None)
```
patrick-kidger, line 261: I think we need to specify `out_axes` as 1 as well. Otherwise I think the output of the `MultiheadAttention` layer will be different as compared to the previous version of Equinox.

tests/test_nn.py
```diff
@@ -524,6 +524,29 @@ def test_convtranspose3d(getkey):
     assert jnp.all(conv(data) == answer)
 
 
+def test_dot_product_attention_weights(getkey):
+    q = jnp.array([[0.0, 2**0.5]])
+    k = jnp.array([[1.0, 0.0], [0.0, 1.0]])
+    weights = eqx.nn.attention.dot_product_attention_weights(q, k)
```
patrick-kidger, line 530: Can these new functions be added to `nn/__init__.py`, so that the lookup here can be `eqx.nn.dot_product_attention_weights`? (It's not really communicated at the moment but individual modules should be treated as private API.)

patrick-kidger (commented):
Nits aside, this LGTM!

tests/test_nn.py
patrick-kidger: Likewise, give these functions a docstring and add them to the documentation?

Alternatively, if you prefer, we can treat these as "semi-public" -- there for power users only, not necessarily with full docs etc. In which case let's expose these under `eqx.experimental`.

jenkspt: I like the "semi-public" approach to start with -- It's the more conservative approach that lets us to make api changes if there are any oversights.

jenkspt: Does this mean just add the imports in `eqx/experimental/__init__.py`?

jenkspt: Actually what makes sense to me is to just keep the namespace in the private attention module -- and so users can use at their own risk until it does get added to the public api

patrick-kidger: Sounds good to me. I can see that you were the one to originally open #142 so that'll suffice for now until someone else wants the same functionality.

patrick-kidger:
Have you checked (with the `out_axes=1` change) that the results of `MultiheadAttention` really do remain unchanged after this update? If so I'm happy to merge this.

jenkspt:
> Have you checked (with the `out_axes=1` change) that the results of `MultiheadAttention` really do remain unchanged after this update? If so I'm happy to merge this.

I checked the outputs and they match with `out_axes=`. The tests should probably be updated to catch this (since using `out_axes=0` passes the tests -- but doesn't match the outputs of the current `MultiheadAttention`. Writing these tests isn't easy however -- maybe just copying a small input/output from the current `MultiheadAttention` and adding it to the tests would suffice?

Also while doing this I noticed that the outputs do not match when using dropout -- I'm not sure yet if this is because the order that dropout is being applied is different, or whether there is a problem with vmap'ing when the `key` is closed over in the `dropout_fn`.

jenkspt:
Yea the problem is vmap'ing with the `dropout_fn`

patrick-kidger:
Just checking that you're not expecting any input from me on this, at this stage?

jenkspt:
The most recent changes fix the problem with closing over the key in `dropout_fn` by adding explicit `key` and `inference` arguments.

However vmap'ing dropout doesn't match the un-vmap'ed version.
For example:

```python
import jax
import jax.numpy as jnp
from equinox import nn

key = jax.random.PRNGKey(41)

x = jnp.arange(4*5).reshape(4, 5)
dropout = nn.Dropout(.5)

y1 = dropout(x, key=key)
y2 = jax.vmap(dropout, in_axes=0, out_axes=0)(x, key=jax.random.split(key, 4))
y3 = jax.vmap(dropout, in_axes=1, out_axes=1)(x, key=jax.random.split(key, 5))

jnp.allclose(y1, y2)    # --> False
jnp.allclose(y1, y3)    # --> False
```

I may be fundamentally misunderstanding how the PRNG works, but my expectation is that y1 and y2 are equal. Any insight into why these don't match? Otherwise I can create a jax issue.

A simpler MWE:

```python
import jax
key = jax.random.PRNGKey(41)

def test(key):
    return jax.random.normal(key, ())

y1 = jax.random.normal(key, (4,))
y2 = jax.vmap(test)(jax.random.split(key, 4))
jnp.allclose(y1, y2)    # --> False
```

jenkspt:
Ok that sounds reasonable. Pending any additional feedback I think this is ready to merge.

patrick-kidger:
Excellent. Thanks for contributing!
(I'll be doing a new release shortly.)

Issue #182, opened 2022-08-26
Couple of suggestions for `nn.Sequential`.


paganpasta:
Hi @patrick-kidger , 

I was wondering if the following modifications are worth adding to `nn.Sequential`

1. Supporting OrderedDicts. They have an ordering but also make it easier to access specific indices based on names.
**Possible bug**: At the moment it is possible to only initialize a `sequential` with an `OrderedDict` (but it shouldn't?). 

2. Supporting `delete`, `insert`, `update`, `append` to edit the state of `layers`. Given that `__getitem__` already exists, these would also be qol additions.

Thanks.

patrick-kidger:
Hmm. I don't think we should add special-casing to `Sequential` to support dictionaries in particular. It's quite important to me that the library code be simple, and in particular not have `if special-case` branches.

If someone really wants to, that at the moment I think they could do something like this:
```python
class IterValueDict(dict):
    def __iter__(self):
        return self.values()

Sequential(IterValueDict(a=..., b=...))
```

Regarding these other methods. Don't forget that we want Equinox modules to be immutable. If we were to offer these methods then we'd have to return the modified module out-of-place, which might be a bit surprising given their current name. Is it common to need to change which layers you have in a module after-the-fact?

paganpasta:
Thanks for the suggestion on the `IterValueDict` ! 

Regarding the additional methods, I only had use-case for `append` functionality and suggested the others based on differences from `torch.Sequential`. For `append`, It is mostly the case of establishing base layers and adding more as per the configuration of a specific model. A workaround would be to simply pass around a list and define a Module at the end.
As per my experience, I suppose the other methods are not that common in practice.

jenkspt:
Currently the `layers` attribute of `Sequential` can be a list -- which of course supports insert, append, ect ...
However it may make sense to force this to be a tuple to keep it immutable.

jenkspt:
> Is it common to need to change which layers you have in a module after-the-fact?

This is pretty common with transfer learning -- i.e. doing "model surgery" on a pretrained model to adapt it to a new task. This is where the mutability of PyTorch Modules is really convenient. It would be pretty cool if equinox Modules could be unfrozen and refrozen to allow PyTorch-like manipulation

patrick-kidger:
> Currently the `layers` attribute of `Sequential` can be a list -- which of course supports insert, append, ect ...
> However it may make sense to force this to be a tuple to keep it immutable.

Yeah, that'd make sense.

[...]

PR #195, opened 2022-09-12
make sequential module immutable

jenkspt:
As discussed in #182 
- Changes Sequential `layers` attribute to be tuple to enforce immutability of equinox modules.
- Constructor can still take any `Sequence` type


patrick-kidger:
LGTM!

PR #197, opened 2022-09-14
Add support for asymmetric padding in Conv and ConvTransposed.

Gurvan:
- Adds asymmetric padding to convolutions. The asymmetric padding is defined as a `Sequence` of `int` of length `2 * num_spatial_dim`, analog to `torch.nn.functional.pad`.

### DIFF
--- a/equinox/nn/conv.py
+++ b/equinox/nn/conv.py
@@ -345,7 +345,9 @@ class ConvTranspose(Module):
             the sequence should be of length equal to `num_spatial_dims`, and specify
             the value of each property down each spatial dimension in turn.. If they
             are an integer then the same kernel size / stride / padding / dilation will
-            be used along every spatial dimension.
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Nice, thank you. For asymmetric padding , can we make the API such that the user just directly passes e.g. `((1,2), (3,4), ...)` and we simply do `self.padding = padding`?

Gurvan:
Oh I didn't see you already made the design choice in Pool.

Changes wrt the previous commit:
- made the `padding` API for Conv and ConvTranspose the same as the Pool one.
- added documentation about `padding` in the info section, rather than in the arguments description.
- added tests for padding with sequence of tuples for Conv and ConvTranspose

patrick-kidger:
And merged! Thank you for contributing.

Gurvan:
Thank you for taking the time!

PR #204, opened 2022-09-25
Add type hints to specialized convolution modules.

jondeaton:


equinox/nn/conv.py
```diff
@@ -221,16 +221,16 @@ class Conv2d(Conv):
...
-        stride=(1, 1),
-        padding=(0, 0),
-        dilation=(1, 1),
-        groups=1,
-        use_bias=True,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: Union[int, Sequence[int]],
+        stride: Union[int, Sequence[int]] = (1, 1),
+        padding: Union[int, Sequence[int]] = (0, 0),
```
patrick-kidger, line 228: And I think this can be a `Sequence[Tuple[int, int]]`.

patrick-kidger (commented):
Thanks for contributing! I've left a couple of nitty comments. These annotations would be great to have.

equinox/nn/conv.py
```diff
@@ -189,16 +189,16 @@ class Conv1d(Conv):
...
-        out_channels,
-        kernel_size,
-        stride=1,
-        padding=0,
-        dilation=1,
-        groups=1,
-        use_bias=True,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: int,
```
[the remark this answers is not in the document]
jondeaton: Good point- fixed here and also on `ConvTranspose1d`.

equinox/nn/conv.py
jondeaton: Are you sure about this? It seems to me the type can definitely be `Tuple[int, int]` but thats already covered by `Sequence[int]` (unless maybe we want to include `Tuple[int, int]` it to give the hint that the sequence must be of length 2?). Also the the documentation on `Conv` says "All of `kernel_size`, `stride`, `padding` and `dilation` can be either an integer or a sequence of integers" so I'm not sure `Sequence[Tuple[int, int]]` is valid?

patrick-kidger: This is used to indicate asymmetric padding.

From the docs: "`padding` can alternatively also be a sequence of 2-element tuples, each representing the padding to apply before and after each spatial dimension."

jondeaton: Thanks for pointing this out– fixed here and also for transpose convolutions.

patrick-kidger:
LGTM! Thanks for contributing.

PR #215, opened 2022-10-22
Use bias (GRUCell)

IsaacBreen:
When you see it... :)

Also made the docstring example more easily runnable.


patrick-kidger:
Ach. Thank you for the fix!

PR #229, opened 2022-11-15
Allow passing file-like objects to eqx.serialise/deserialise

jatentaki:
As above. My use-case is to serialize/deserialize into `zipfile.ZipFile` sub-files which are available as file-like objects cannot be accessed by paths. I am not claiming this is the best possible implementation, treat it like a more constructive feature request :)


patrick-kidger:
This LGTM as written! Thanks for the PR.

PR #233, opened 2022-11-18
fix tree_inference

uuirs:
https://github.com/patrick-kidger/equinox/blob/253522cd23b0522e57088d4c886a753b2594e0d7/equinox/tree.py#L259-L260

https://github.com/patrick-kidger/equinox/blob/253522cd23b0522e57088d4c886a753b2594e0d7/equinox/tree.py#L268-L269

The `out.extend` inside the for loop of `out` will cause the same node to be added many times. 

For example:
```python

import equinox as eqx
import time
class Model(eqx.Module):
  sub_module: eqx.Module
  inference: bool
  def __init__(self, sub_module):
    self.sub_module=sub_module
    self.inference=False

model = Model(None)
for i in range(11):
  model = Model(model)

now = time.time()
eqx.tree_inference(model, True)
print(time.time()-now) # 5s+

def _inferences(pytree):
  out = eqx.tree._inferences(pytree)
  print(len(out), len(set(out))) # 82500 12
  return out

eqx.tree_at(_inferences, model, replace_fn=lambda _: True)
```

### DIFF
--- a/equinox/tree.py
+++ b/equinox/tree.py
@@ -253,21 +253,17 @@ def tree_equal(*pytrees: PyTree) -> Union[bool, np.bool_, Bool[Array, ""]]:
     return out
 
 
-def _subinferences(pytree):
-    is_leaf = lambda x: hasattr(x, "inference") and x is not pytree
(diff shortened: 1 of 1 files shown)


patrick-kidger:
LGTM! Thank you for the fix.

PR #235, opened 2022-11-23
Support Buffer Donation in filter_jit

### LINKED ISSUE #3: Support a `donate_spec` argument in `filter_jit`
As an analogue for `donate_argnums`.

uuirs:
Try to close  #3 

Hi @patrick-kidger 

Jax supports buffer donation on CPU after jaxlib >0.3.22, it is necessary for many in-place updation.

I've drafted a proposal in here.

**Details:**
* Adds `donate_default`, `donate_args`, `donate_kwargs`, and `donate_fn` for control of buffer donation;
* Adds Tests.

**Considerations:**
* It would be trivial to make these arugments work only with parameters marked as traced.
* We need the ability to control buffer donation in function, args and kwargs, which is similar to handling Tracing.

**More(not in this PR):**
Supports buffer donation in filter_pmap


patrick-kidger:
This is a great feature to have.

This has prompted me to think about making some larger-scale changes to Equinox. So besides simply accepting this PR, I'd be curious to know what you think about each of the following other options.

1. Right now we specify a spec based around the dichotomy of `True` and `False`, for array-vs-non-array. Should we extend this to a trichotomy `True`, `False`, `donate`? Since a donated argument is necessarily also traced. Otherwise, the API proposed in this PR makes it possible to have a donated static arg, which is meaningless. (`donate` would be some special sentinel value.)

2. Should we make it so that all arrays are donated by default, unless specified otherwise? (So a trichotomy of `True`, `False`, `no_donate`.) This is probably a silent performance improvement for most users, and will be a loud error (i.e. not a silent error) for those rare cases in which an array really is used after JIT'ing.

3. Should we make it so that all arrays are always traced, and all non-arrays are always static? Essentially, simplify the `filter_jit` API by removing a lot of its complexity. I think it's really rare to want to deviate from the default. (Something I didn't appreciate when I first wrote it -- if I could start from scratch then I think I would simply not give it any filtering arguments at all.)

4. If we take option 3: do we either (a) keep arguments that specify whether to donate an array, or (b) forcibly donate, and simply expect a user to perform a manual copy (before the `filter_jit` call) in the rare case they want to keep a copy around?

For the sake of API elegance I quite like the approach of donating all arrays and static'ing all non-arrays, but that might be a mildly breaking change for some small number of use cases.

Other musings:

- I'd be very happy to delete the old-style `filter_spec` argument either way, that's long since deprecated.

- Should we consider removing the filter arguments from all the other filtered transformations? Once again it's very rare to need a non-default for any of these, and in these cases then it can easily be replicated using `eqx.partition` and `eqx.combine`. We then get the benefits of a much simpler API throughout all of Equinox: nice docs, easier new user onboarding, code maintainability, etc. Indeed several of the more unusual filtered transformations that I added later (`filter_{make_jaxpr, closure_convert, vjp, ...}`) already don't take any filter arguments at all.

Since you've taken the time to implement donations for `filter_jit`, I'm guessing you may have a stake in the answers to some of these questions. WDYT?

uuirs:
Thanks for your input! Sorry for late reply.

1

> the API proposed in this PR makes it possible to have a donated static arg

In this implementation, all donated pytree are partitioned from dynamic pytree. As mentioned, it only works with parameters marked as traced, so it doesn't seems like it will be a donated static arg.

a test: 
``` python
import jax
import jax.numpy as jnp
import equinox as eqx
@eqx.filter_jit(donate_args=(False,True)) # True on static
def f(x, y):
  return x+1
x = jnp.zeros((100,))
y = 10
closed_jaxpr = eqx.filter_make_jaxpr(f)(x,y)[0]
xla_call = closed_jaxpr.eqns[0]
assert xla_call.params['donated_invars'] == (False,)
```

> Should we extend this to a trichotomy True, False, donate?

I did consider adding a enum(static, dynamic, donate), and adding chain_partition or something to return several pytree. 

But also, it seems to work that just making donated partition have a lower priority than the original partition and based on dynamic pytree. I think usually we just need to donate model without any additional specs(args, kwargs, etc.).

I personally prefer
`eqx.filter_jit(fun, donate_args=(False, False, True,))` (`args[0:3]` of `fun` all can be a normal module)
`eqx.filter_jit(fun, donate_kwargs=dict(model=True))` (`model` of `fun` can be a normal module)
over 
`eqx.filter_jit(fun, args=(False, True, eqx.donate))` (I don't care if `args[0:2]` are static or dynamic, and usually would not use like this)
`eqx.filter_jit(fun, args=(eqx.is_array, eqx.is_array, eqx.is_array_donated))` (So it might become like this, something I guess would be)
`eqx.filter_jit(fun, kwargs=dict(model=eqx.is_array_donated))` (works)

From this point of view, I would like to keep these `donate_*` and remove the other arguments(kwargs, args, fn, etc.)

2
> Should we make it so that all arrays are donated by default, unless specified otherwise?

I totally agree, if the default behavior can be changed easily via `donate_default=False` or something.

Some small issues:
1. Usually the pytree in input is mixed(dynamic part, static part, donated part, something to reuse, etc.), because we need to return those arrays that need to be reused, for convenience, we may end up returning the entire pytree. Then there are some DCE in jaxpr such as pruning unused input(a related resolved issue, https://github.com/google/jax/issues/5145), now we return all inputs which may make these DCE not work properly.
2. There may be some annoying warning when compiling, complaining that some donated buffers were not usable.

3
> Should we make it so that all arrays are always traced, and all non-arrays are always static?

I think it really makes things easier. Usually I'm use `filter_jit` without any arugment and it works fine. But sometimes I need it to make sure the paramters are treated as dynamic, which is easier than to checking when calling. many existing python functions do not return jnp arrays. We might add another wrapper to just avoid wrong static input in case.


4
> (a) keep arguments that specify whether to donate an array, or (b) forcibly donate, and simply expect a user to perform a manual copy

I prefer to choose (a), just similar to the case above about type wrapper, I like to determine this when defining jitting, no when calling jitted.
In a rare case like large device arrays without ownership(which means I can't easily assign back the reference ), copying is expensive.


IMO, if big changes are allowed, I'd like to remove all the filtered argument and keep donate_* in jit and pmap, and add a type wrapper for users to specify argument as dynamic.

patrick-kidger:
These are some really interesting thoughts.

> which may make these DCE not work properly.

I'm not sure it will. I can believe that passing array in-and-out will prevent them from getting DCE'd, but (a) isn't that already happening independent of this change, and (b) wouldn't this change actually improve things here, because we would be donating and thus reusing the input buffer?

> There may be some annoying warning when compiling

Good point here. We could possibly suppress these warnings.

> But sometimes I need it to make sure the paramters are treated as dynamic, which is easier than to checking when calling. many existing python functions do not return jnp arrays.

That's an interesting use-case. How frequently does that happen? It would be easy enough to make this happen with a wrapper function instead, though:
```python
@eqx.filter_jit
def foo_impl(x, y):
    ...

def foo(x, y):
    x, y = jax.tree_util.tree_map(jax.numpy.asarray, (x, y))
    return foo_impl(x, y)
```

> I like to determine this when defining jitting, no when calling jitted.

Again, I think this could be handled with a wrapper function:
```python
@eqx.filter_jit
def foo_impl(x, y):
    ...

def foo(x, y):
    x = jnp.array(x)  # copy
```

uuirs:
Refactored, seems to be ok.
Todo: documentation and example fixes.

uuirs:
Hi @patrick-kidger, how does it look now?  I've checked the examples on colab and it seems to be ok. 
Is there anything else I haven't considered?

patrick-kidger:
(I'm at NeurIPS at the moment, I'll give a review next week :) )

uuirs:
Thanks, have a nice day!

equinox/jit.py
```diff
@@ -147,43 +117,18 @@ def filter_jit(
...
-        g(1, 2)  # both args static
-
-        h(1)  # traced
-        h(jnp.array(1))  # traced
-        h("hi")  # not a trace-able JAX type, so error
-
-        apply(lambda x: x + 1, jnp.array(1))  # first arg static, second arg traced.
+        x = jnp.array(2)
+        new_x = f(x, 1) #  donated
+        x.is_deleted() # True
```
patrick-kidger, line 126: Can we add examples demonstrating how to:
- trace ints/floats/bools/complexes, by using `jnp.asarray`?
- how to avoid donating just a single array, by using `jnp.copy`?

Something like the examples I gave in my previous message.

examples/frozen_layer.ipynb
```diff
@@ -106,18 +106,21 @@
     "\n",
     "    # Step 3\n",
     "    @eqx.filter_jit\n",
-    "    @ft.partial(eqx.filter_value_and_grad, arg=filter_spec)\n",
+    "    @ft.partial(eqx.filter_value_and_grad, arg=filter_spec, has_aux=True)\n",
```
patrick-kidger, line 109: This is actually a nice example of the sort of silent performance error that it was possible to make before this PR.

In this example, the JIT'd region is currently only covering the computation of the gradients. But it should actually also include the `optim.update` and `eqx.apply_updates` lines as well, since these are also JAX operations. See the RNN example for the correct way to do it. Do you want to apply that fix to this example? This would mean that it isn't necessary to use `has_aux` here, since we won't need to preserve `model` to use after the JIT region.

tests/test_jit.py
```diff
@@ -30,133 +35,44 @@ def test_filter_jit1(api_version, getkey):
...
-        True,
-        True,
-        False,
-        {"a": True, "tuple": (False, True)},
-        True,
-        eqx.is_inexact_array,
-    ]
+    array_tree_ = jax.tree_map(
+        lambda x: jnp.copy(x) if eqx.is_array(x) else x, array_tree
+    )
```
patrick-kidger, line 42: Nit: `jtu.tree_map(jnp.copy, array_tree)` suffices as all elements are arrays.
Also note the use of `jax.tree_util` (`jtu`) rather than `jax`. (The `tree_map` utility is being moved.)

tests/test_jit.py
```diff
@@ -420,5 +317,114 @@ def f(x, y):
...
-    fn = jtu.Partial(f, True)
-    eqx.filter_jit(g, fn=fn)
+    eqx.filter_jit(g)
+
+
+def test_buffer_donation(getkey):
+    def f(x):
+        return x + 1
+
+    def _test(jit_fun, x):
```
patrick-kidger, line 327: This function is only used once so it could be inlined where it is called.

tests/test_jit.py
```diff
@@ -420,5 +317,114 @@ def f(x, y):
...
+        old_p = x.unsafe_buffer_pointer()
+        new_x = jit_fun(x)
+        assert new_x.unsafe_buffer_pointer() == old_p
+        assert x.is_deleted()
+        return new_x
+
+    f_jit = eqx.filter_jit(f)
+
+    assert _eq(_test(f_jit, jnp.array(0)), jnp.array(1.0))
```
patrick-kidger, line 336: The `_eq` function should probably be removed and `helpers.shaped_allclose` used instead. (Which is basically the same and handles pytrees and floating-point arrays.)

tests/test_jit.py
```diff
@@ -420,5 +317,114 @@ def f(x, y):
...
+    _, new_m = m(jnp.ones((10,)))
+    _, new_m = new_m(jnp.ones((10,)))
+    _, new_m = new_m(jnp.ones((10,)))
+    assert num_traces == 1
+
+    assert _eq(new_m.buffer[0], jnp.array(3.0))
+    assert m.buffer.is_deleted()
+    assert new_m.buffer.unsafe_buffer_pointer() == old_m_p.buffer
+
+    old_m_deleted_flag = jax.tree_map(
```
patrick-kidger, line 374: Another use of `jax` instead of `jtu`.

patrick-kidger (commented):
This looks excellent. I've gone through; my comments are just small nits.

Other thoughts:
- Is there anywhere in the documentation that needs updating to reflect this change? I don't think so but I thought I'd double-check with you?
- We should think about similarly updating the other filtered transformations. Don't feel obliged to do this if you don't want to; if not then I'll get that done at some point and then do a new release. (In separate PRs, either way.)

Thank you for making this change happen!

equinox/jit.py
```diff
@@ -92,50 +86,26 @@ def __get__(self, instance, owner):
...
-        `__call__`, and doesn't have to be a normal Python function.)
+    - `donate` indicates whether the buffers of JAX arrays are donated or not, it
+        should either be:
+        - 'arrays': the default, donate all arrays and suppress all warnings about
+            unused buffers;
+        - 'warn': as above, but don't suppress unused buffer warnings;
+        - 'none': disables buffer donation.
     - `**jitkwargs` are any other keyword arguments to `jax.jit`.
 
-    When `args`, `kwargs`, `out`, `fn` are prefixes of the corresponding input, their
```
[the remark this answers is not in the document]
uuirs: I'm not quite sure if I should remove `jitkwargs`, such as https://github.com/google/jax/issues/12714 (but in `pjit`)still need add `keep_unused=True` as a workaround.
And `device` or `backend` can specify the device to perform some operations like allreduce, or something like https://github.com/google/jax/issues/1899 (old but recently mentioned).
But at least users can downgrade to use `jax.jit`.
I want to confirm your decision on this matter.

patrick-kidger:
Alright, merged! Thank you for the PR, this is a great simplification to have. It's on the `dev` branch for now -- once we've unified things (with the other filtered transformations) then I'll merge it and do a new release.

If you are happy to handle the other filtered transformations -- in other PRs -- then I'd be very happy to accept them. I'm thinking that we could:
- Remove the `arg` parameter from `filter_{grad, value_and_grad}`.
- Perform the same simplification to `filter_pmap` as we've done here.
- Figure out if there's a way to simplify the `filter_{vmap,pmap}` interfaces, since they still remain quite complicated.

uuirs:
Thank you for your patience and support! I'll try to submit them next week.

Issue #149, opened 2022-07-24
Updating `operation` in `equinox.nn.Pool` to `static_field`?


paganpasta:
Hi @patrick-kidger,

Using Pool modules with `jax.jit` raises `TypeError: Argument '<function max at 0x7ffb8ca7b940>' of type <class 'function'> is not a valid JAX type.` 

It can be circumvented by decorating the `forward pass` with `@eqx.filter_jit` instead. 
However, for a uniform behaviour, the `operation` attribute can be defined as `static_field()` to fix this difference. Is this difference intended? 

A script to reproduce the behaviour
```python
from equinox.custom_types import Array

import jax
import jax.random as jrandom
import equinox as eqx
import equinox.nn as nn


class Net(eqx.Module):
  maxp: eqx.Module

  def __init__(self):
      super().__init__()
      self.maxp = nn.MaxPool2d(kernel_size=3, stride=2)

  def __call__(self, x: Array) -> Array:
      return self.maxp(x)


@jax.jit  #breaks
# @eqx.filter_jit  #works
def forward(model, x):
    return jax.vmap(model)(x)

model = Net()
random_image = jax.random.uniform(key=jrandom.PRNGKey(0), shape=(1, 3, 224, 224))
output = forward(model, random_image)
print(output.shape)
```

patrick-kidger:
This intended, and is actually pretty much the purpose of filtering: filtering is used to handle the fact that when treated as a PyTree, your model may have leaves that are not JAX types.

The thing is that we can't unconditionally make callables (like `operation`) a static field, because in principle we may wish to use a learnt callable (i.e. the callable may itself be a PyTree with parameters).

Variations of this come up every now and again, most frequently wrt the activation function of `MLP` (#69, #105, #108). This definitely seems to be the biggest hurdle people have with getting up to speed with Equinox, so I'd welcome thoughts on any way the docs might be improved.

paganpasta:
Thanks a lot for the response.  

Just listing down here few things I think may help (hopefully).

- Currently, there is no indication in the docs as to which `nn.Module`s will not play well with `jax.jit` and which one will require filtering. Every time going through the source code to understand which Module might not work is cumbersome. This information can be highlighted in the doc? 

[...]

PR #239, opened 2022-11-26
Added notes about 'TypeError: not a valid JAX type' to FAQ and nn.Lambda documentation.

carlosgmartin:
For https://github.com/patrick-kidger/equinox/issues/149#issuecomment-1328088282.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -76,16 +76,22 @@ import jax
 import equinox as eqx
 
 def loss_fn(model, x, y):
-    return (model(x) - y) ** 2
+    return ((model(x) - y) ** 2).mean()
 
 model = eqx.nn.Lambda(lambda x: x)
+model = eqx.nn.MLP(2, 2, 2, 2, key=jax.random.PRNGKey(0))
+
+x = jax.numpy.arange(2)
+y = x * x
 
 try:
-    jax.jit(loss_fn)(model, 0, 0) # error
+    jax.jit(loss_fn)(model, x, y) # error
 except TypeError as e:
     print(e)
 
-eqx.filter_jit(loss_fn)(model, 0, 0) # ok
+eqx.filter_jit(loss_fn)(model, x, y) # ok
 ```
+
+This error happens because a model, when treated as a PyTree, may have leaves that are not JAX types (such as functions). It only makes sense to trace arrays. Filtering is used to handle this.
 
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -63,3 +63,29 @@ linear = eqx.nn.Linear(input_size, output_size, key=key)
...
+```
+Example:
+```python3
+import jax
+import equinox as eqx
+
+def loss_fn(model, x, y):
+    return (model(x) - y) ** 2
+
+model = eqx.nn.Lambda(lambda x: x)
```
patrick-kidger, line 81: I suspect `Lambda` and `MLP` are going to be the two most frequent culprits here. Can we add both to the example here? E.g:
```python
# Use a model that includes a non-array in its pytree structure
model = eqx.nn.Lambda(jax.nn.relu)
model = eqx.nn.MLP(2, 2, 2, 2, key=jax.random.PRNGKey(0))
```
(or similar)

docs/faq.md
```diff
@@ -63,3 +63,29 @@ linear = eqx.nn.Linear(input_size, output_size, key=key)
...
+
+try:
+    jax.jit(loss_fn)(model, 0, 0) # error
+except TypeError as e:
+    print(e)
+
+eqx.filter_jit(loss_fn)(model, 0, 0) # ok
+```
+
+Instead of [`jax.jit`](https://jax.readthedocs.io/en/latest/_autosummary/jax.jit.html), use [`equinox.filter_jit`](https://docs.kidger.site/equinox/api/filtering/filtered-transformations/#equinox.filter_jit). Likewise for [other transformations](https://docs.kidger.site/equinox/api/filtering/filtered-transformations/).
```
patrick-kidger, line 91: Can we say a couple of words about why this is? That is, explaining that it only makes sense to try and trace arrays.

equinox/nn/composed.py
```diff
@@ -160,6 +160,8 @@ class Lambda(Module):
     Precisely, this just adds an extra `key` argument (that is ignored). Given some
     function `fn`, then `Lambda` is essentially a convenience for `lambda x, key: f(x)`.
 
+    If you get a TypeError saying the function is not a valid JAX type, see the FAQ.
+
```
patrick-kidger, line 164: 1. Can you add a hyperlink to the FAQ entry?

2. I think it might be worth wrapping this into an admonition:
```
!!! faq

    If you get...
```

3. Do you think it would be worth adding the same note in the docstring for `MLP`?

patrick-kidger (commented):
Nice, thankyou! Nitty comments aside this LGTM.

patrick-kidger:
Merged! Thank you for contributing.

PR #274, opened 2023-03-06
Changes attention so that (1) the scaling is applied to the query bef…

j5b:
…ore computing the product for improved numerical stability and (2) the smallest float is used for masking instead of negative infinity to avoid NaNs.

### DIFF
--- a/equinox/nn/attention.py
+++ b/equinox/nn/attention.py
@@ -19,8 +19,8 @@ def dot_product_attention_weights(
     mask: Optional[Bool[Array, "q_seq kv_seq"]] = None,
 ) -> Float[Array, "q_seq kv_seq"]:
 
+    query = query / math.sqrt(query.shape[-1])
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Do you have a MWE that generates NaNs here?

j5b:
Sure! After running

```
query = jnp.asarray([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
attn_mask = jnp.asarray([[[1, 1, 0], [1, 1, 0], [0 , 0, 0]]])

attention = eqx.nn.MultiheadAttention(
    num_heads=1, query_size=3, key=jax.random.PRNGKey(0))
result = attention(query, query, query, mask=attn_mask, inference=True)
```

`result` contains the following:

```
Array([[ 0.25281057,  0.33115894, -0.14003609],
       [ 0.2174798 ,  0.29444888, -0.12778063],
       [        nan,         nan,         nan]], dtype=float32)
```

PR #275, opened 2023-03-07
added simple mnist example

Artur-Galstyan:
A few things:

Running the tests and the pre-commit hook (PCH) doesn't work on MacOS out of the box. This is because the PCH tries to install tensorflow via "pip install tensorflow" but on MacOS you install it via "pip install tensorflow-macos". 

Therefore, I used --no-verify on this PR to skip the PCH. I highly doubt that this notebook would introduce any breaking changes, so it should be fine?


patrick-kidger:
Thanks for the contribution! It'd be great to get this cleaned up and merged.

Some initial things that need tidying up:

- Can you add the example to `mkdocs.yml`, so that it appears in the documentation?
- You will still need to run the pre-commit hooks for formatting. It looks like e.g. this notebook doesn't have sorted imports. I'd recommend just switching the `tensorflow` for `tensorflow-macos` locally. (But don't submit that change here.)
- For the notebook itself: can you submit a clean run? For example I notice that the first cell is [152] in the sequence.
- Likewise some text talking through what we're doing -- follow the format of the existing examples.
- Let's drop the `sklearn` dependency.
- Is there any reason you've defined your own `NeuralNetwork` instead of using `eqx.nn.MLP`? I appreciate that maybe it's worthwhile including a larger version for pedagogical reasons, but if so I think it'd be better to make the example be something that isn't already in the base library -- e.g. a small CNN.
- Let's use `eqx.filter_jit` etc. rather than `jax.jit`.
- JIT'ing the loss function isn't necessary; the `step` function is already JIT'd.

Artur-Galstyan:
Hi Patrick!

I incorporated your remarks:

- The example is now in the docs
- dropped sklearn one_hot encoding in favour of Jax one_hot
- created CNN instead of NN
- used (and tried to explain) when and why to use the `filter_[...]` methods

Hope it's better now :)

BTW I'm a Git Newbie, hope everything is in order 😄  Cheers.

patrick-kidger:
Alright, taking another look at this again. Various notes:

- Note that `isinstance(model, PyTree)` will unconditionally return `True`. By definition everything is a PyTree. (It might just be a stump containing a single leaf. E.g. try doing `jax.tree_util.tree_flatten(object())`.)

- I think using `eqx.tree_pprint` is unnecessary, just printing `model` directly should produce the same output. (As `eqx.Module.__repr__ == eqx.tree_pformat`.)

- Try to avoid calling JAX operations outside of JIT. (E.g. `preds = jax.vmap(model)(x)` inside `evaluate`.) This will be very slow. In the case of this example, you want something like
    ```python
    @eqx.filter_jit
    def compute_accuracy(model, x, y):
        preds = jax.vmap(model)(x)
        targets = jnp.argmax(y, axis=1)
        preds = jnp.argmax(preds, axis=1)
        return sum(targets == preds) / len(x)
        
    avg_acc += compute_accuracy(model, x, y)
    ```
    (Likewise prefer using `np.round` instead of `jnp.round` inside `fit`.)

All-in-all I think this looks very cleanly done, and a worthy addition to the documentation. I'd even suggest we bump it up (in `mkdocs.yml)` to being the first example. Once you've made the above changes then I'll merge this.

Artur-Galstyan:
Hi Patrick!

I've tried to incorporate your suggestions. I've changed the evaluate method such that we have a separate `compute_accuracy` method. I've also tried to use it as an example to use `filter_jit` instead of `jax.jit` (since `jax.jit` throws an error that functions can't be interpreted as arrays, which I think is a good example use case for `filter_jit`). 

I've also added some typing annotations from jaxtyping and added a link to the jaxtyping repo. 

`jnp.round` has been replaced by `np.round`. I've also added a little hint, saying that one should generally avoid using Jax operations outside of `jit` (I didn't know that one, good to know). 

Regarding your first comment about `jax.tree_util.tree_flatten(object())`. In the end, I decided to remove that part altogether since the output of `tree_flatten` was very cluttered and I didn't think that it would contribute much to the overall example (which is essentially a "hello world" example to quickly get started with Equinox).

Artur-Galstyan:
Hey Patrick, no worries, I'm glad I could help 👍

PR #277, opened 2023-03-07
Added U-Net example

patrick-kidger:
CC @Benjamin-Walker -- I thought I'd put this somewhere public so that it can be used.


Benjamin-Walker:
Sure, sounds good to me.

PR #285, opened 2023-03-09
Make eqx.filter_vmap more robust to None in_axes and small bugfix.

RaderJason:


patrick-kidger:
Closing in favour of #286. There's some odd issue with pulling on old code here, so I've replayed your changes on top of the current main as a new commit.

Issue #287, opened 2023-03-11
`deterministic=True` ignored in MultiHeadAttention


mk-0:
I see that `deterministic` is deprecated, but while `Dropout` itself handles it and warns, `MultiHeadAttention` simply ignores it:

```python
from jax import random as jrandom
import equinox as eqx

qk, kk, vk, ak, dk1, dk2 = jrandom.split(jrandom.PRNGKey(0), 6)
attn = eqx.nn.MultiheadAttention(num_heads=1, query_size=1, dropout_p=0.1, key=ak)
q = jrandom.uniform(qk, (5, 1))
k = jrandom.uniform(kk, (5, 1))
v = jrandom.uniform(vk, (5, 1))

assert (
    attn(q, k, v, key=dk1, inference=True) == attn(q, k, v, key=dk2, inference=True)
).all()  # this works

assert (
    attn(q, k, v, key=dk1, deterministic=True) == attn(q, k, v, key=dk2, deterministic=True)
).all(), "Not actually deterministic"
```
I'm just learning jax and equinox, but if this seems right, should I submit a PR?

patrick-kidger:
Yup, that's a bug. Very happy to take a PR on this, thank you.

PR #288, opened 2023-03-12
Fixed deterministic=True being ignored in MultiheadAttention

mk-0:
Handled the argument, added a warning and 2 simple test cases, all similar to how Dropout does it

Relates: #287


patrick-kidger:
Merged! Thank you for the fix.

PR #299, opened 2023-03-24
`cls.__new__` -> `object.__new__` in `tree_unflatten`

### LINKED ISSUE #296: Custom new method throws error when using `tree_unflatten`
Hello Patrick,

```python
import equinox as eqx 
import jax.tree_util as jtu

class Tree(eqx.Module):
    x : int = 1
    def __new__(cls, x):
        print("x =", x)
        self = super().__new__(cls)
        return self
    
tree = Tree(1)
# new
```

```python
leaves, treedef = jtu.tree_flatten(tree)
jtu.tree_unflatten(treedef, leaves)

    262 @classmethod
    263 def tree_unflatten(cls, aux, dynamic_field_values):
--> 264     self = cls.__new__(cls)
    265     dynamic_field_names, static_field_names, static_field_values = aux
    266     for name, value in zip(dynamic_field_names, dynamic_field_values):

TypeError: Tree.__new__() missing 1 required positional argument: 'x'

```

To mitigate this bug, I suggest using `object.__new__` in the `tree_unflatten` rule .

ASEM000:
Fix #296

### DIFF
--- a/equinox/module.py
+++ b/equinox/module.py
@@ -295,7 +295,7 @@ class Module(metaclass=_ModuleMeta):
 
     @classmethod
     def _tree_unflatten(cls, aux, dynamic_field_values):
-        self = cls.__new__(cls)
+        self = object.__new__(cls)
         dynamic_field_names, static_field_names, static_field_values = aux
         for name, value in zip(dynamic_field_names, dynamic_field_values):
             object.__setattr__(self, name, value)


patrick-kidger:
LGTM!
(The test failure looks like an unrelated flake, I'll look into that.)

Thank you for contributing.

PR #303, opened 2023-03-27
Make `depth` more explicit

RaderJason:
It's not immediately clear what convention `depth` is using. For example, a depth two network could either correspond to a network of shape: (in layer, hidden layer 1, hidden layer 2, output layer) or of shape (in layer, hidden layer, output layer). This makes it clear exactly which convention equinox uses.

### DIFF
--- a/equinox/nn/_composed.py
+++ b/equinox/nn/_composed.py
@@ -68,7 +68,7 @@ class MLP(Module):
         - `out_size`: The output size. The output from the module will be a vector
             of shape `(out_features,)`.
         - `width_size`: The size of each hidden layer.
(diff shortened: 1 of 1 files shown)


equinox/nn/_composed.py
```diff
@@ -68,7 +68,9 @@ def __init__(
         - `out_size`: The output size. The output from the module will be a vector
             of shape `(out_features,)`.
         - `width_size`: The size of each hidden layer.
-        - `depth`: The number of hidden layers.
+        - `depth`: The number of hidden layers including the output layer. ie. depth=2
```
patrick-kidger, line 71: Nits:

- `layers, including`
- ``For example, `depth=2` ``

PR #314, opened 2023-04-12
Clip variance in normalization

anh-tong:
Hi Patrick,

I encountered some numerical instability in training with `LayerNorm` as losses return `NaN`. When turning on `jax_debug_nans`, it showed that `LayerNorm` failed to compute `jax.lax.rsqrt`.

I refer to Flax about this issue. That is, it is better to clip variance (this [line](https://github.com/google/flax/blob/660fd6c16df3a5fa2b5b0c2eaf9e2ca4de40bea4/flax/linen/normalization.py#L101)) to enable numerical stability for float32.

I don't have a certain test case for this.


patrick-kidger:
Interesting! LGTM, thanks for the fix.

Issue #310, opened 2023-04-09
Layernorm optional bias


lockwo:
In some cases, we want to make the bias of layernorm optional (but keep the weights, so we can't just use elementwise_affine = False, e.g. https://github.com/karpathy/nanoGPT/blob/master/model.py#L27), however in eqx my thinking is that this would require me to basically make a new LayerNorm class (similar to that link) without the bias, just to change this attribute. Would it be possible to make `use_bias` a flag for the layernorm layer? I would be happy to make this PR if so, and interested in hearing any opposition to this idea.

PR #315, opened 2023-04-12
Change LayerNorm to allow weights or biases

lockwo:
Addresses https://github.com/patrick-kidger/equinox/issues/310. I handled the deprecation of `elementwise_affine` in as similar a manner as I could to `dropout` which also has a deprecated argument.


patrick-kidger:
Thanks for the fix! This looks good to me.
The test failure looks like a flake, I'll fix that separately.

PR #316, opened 2023-04-13
Initialize dropout differently across batch and across sequence in BE…

j5b:
…RT example.

### DIFF
--- a/examples/bert.ipynb
+++ b/examples/bert.ipynb
@@ -50,7 +50,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": null,
    "metadata": {
     "cellView": "form",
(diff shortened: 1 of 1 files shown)


patrick-kidger (commented):
This bit:
```python
        ff_keys = (
            [None] * seq_len
            if ff_key is None
            else jax.random.split(ff_key, num=seq_len)
        )
```
looks wrong to me. I think it should be
```python
        ff_keys = (
            None
            if ff_key is None
            else jax.random.split(ff_key, num=inputs.shape[0])
        )
```

Other than that LGTM!

j5b:
Good catch! This should be fixed now.

patrick-kidger:
Great! Thanks for the fix.

Issue #289, opened 2023-03-15
Multigpu training?


rdilip:
So I know for a while there was an issue with equinox not playing well with pmap. With the advent of pjit (and I think now just jit), are there examples of using equinox for multigpu training?

patrick-kidger:
I think what you're referring to is #192: for a while `filter_pmap` had quite high overhead. That should now be substantially reduced, though.

Equinox should play well with both pmap and (p)jit. I don't think the documentation has any examples, it's true. (I'd be happy to take a PR on this.)

rdilip:
If you have an easy example, that'd be great! I'm hoping to start a small new project in equinox, but I'm checking to make sure there's nothing important that I need currently missing.

ChenkaiMao97:
Hi Patrick, thanks for the awesome repo, I really enjoyed learning and using it. I am trying to add distributed training with pmap, and currently I'm having issues with UnexpectedTracerError. I'm not very familiar with JAX so most times it's trial and error. I would also appreciate it if we could have some example showing how to use pmap (together with vmap) for distributed training. Thanks!

agpcd:
+1

patrick-kidger:
I've just added a new [parallelism example](https://docs.kidger.site/equinox/examples/parallelism/) to the docs. :)

What do people think?

agpcd:
Also worth mentioning that the BERT example does multigpu training with einops (very neat by the way)

https://docs.kidger.site/equinox/examples/bert/#bert-implementation

PR #334, opened 2023-05-02
Add support for pjit

ZacCranko:
This PR addresses #289 by patching `filter_jit` with a switch to call `jax.experimental.pjit.pjit` instead of `jax.jit`. I've been using this patch extensively and seems to be completely seamless. 

Also @patrick-kidger, I'm just wondering what the practice of  importing everything with the `thing as thing` convention does? I've seen this in a few places elsewhere.

### DIFF
--- a/equinox/__init__.py
+++ b/equinox/__init__.py
@@ -21,7 +21,7 @@ from ._filters import (
     is_inexact_array_like as is_inexact_array_like,
     partition as partition,
 )
-from ._jit import filter_jit as filter_jit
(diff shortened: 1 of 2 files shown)


patrick-kidger:
JAX recently unified pjit and jit into a single operation. I don't think there's a difference between them any more, and this toggle shouldn't be necessary.

The `import foo as foo` convention is used to declare to static analysis tools that this is a re-export, i.e. that `foo` isn't private to the module.

ZacCranko:
Ah that's neat. I didn't see that. Thanks.

PR #343, opened 2023-05-08
Lots of improvements to attention

patrick-kidger:
- Support for autoregressive attention;
    - Includes support for zero-length queries, e.g. when populating the caches for the prompt.
- Causal masking available by passing mask="causal";
- Support for multi-query attention.

Still TODO:
- support biases, not just masks.
- interpolate between MHA and MQA
- have KV caching not push elements backwards at the end.
- ~cast softmax to float32~ [Done elsewhere!]

### DIFF
--- a/docs/api/nn/attention.md
+++ b/docs/api/nn/attention.md
@@ -1,5 +1,9 @@
 # Attention
 
+::: equinox.nn.self_attention
+
+---
+
 ::: equinox.nn.MultiheadAttention
     selection:
         members:
--- a/equinox/nn/__init__.py
+++ b/equinox/nn/__init__.py
@@ -1,5 +1,8 @@
 from ._activations import PReLU as PReLU
-from ._attention import MultiheadAttention as MultiheadAttention
+from ._attention import (
+    MultiheadAttention as MultiheadAttention,
+    self_attention as self_attention,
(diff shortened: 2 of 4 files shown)


Artur-Galstyan:
Hi Patrick, 

regarding the interpolation between MHA and MQA: I mentioned the double VMAP strategy (see https://www.jitx.io/posts/multihead-attention) and finally got around to benchmarking it, see:
image

This is just trained on the TinyShakespeare dataset, but it seems as though - performance wise - it's close or _very, very slightly ahead_ of the normal MHA implementation. I'll do some more experiments later, but the preliminary results look promising that the double-vmap strategy works

marcelroed:
What happened to this effort, @patrick-kidger? Would you be interested in contributions to finish this?

patrick-kidger:
So we came to the conclusion that most 'serious' users of attention are probably writing their own implementations. This has been quite a common trend for a while.

As such it made most sense for Equinox to essentially just provide a 'default' version of attention that covers most introductory-to-intermediate use-cases, rather than a complicated API -- easy to get lost in! -- that tries to also cover everything else.

If there are small extensions to the existing API that you'd like, then I'd be willing to consider those though :)

PR #358, opened 2023-05-17
Add support for bias to MLP

### LINKED ISSUE #290: Configure "bias" in the MLP
Amazing package.  Would it be possible to extend the `MLP` type in https://github.com/patrick-kidger/equinox/blob/main/equinox/nn/composed.py  to support passing on to the `use_bias` argument for the `Linear` layers?  In many cases, I have found that things work better if there is a bias on only the last output layer, on only the internal layers, and vice versa so separating them would be nice.  I haven't seen cases where you want only the first layer to have a bias, so leaving that one out.

To be more specific, I would imagine,
- In https://github.com/patrick-kidger/equinox/blob/main/equinox/nn/composed.py#L44-L50  add in two arguments: 
```python
use_bias: bool = static_field(),
last_use_bias: bool = static_field(),
```
- And then the code becomes something like
```python
        if depth == 0:
            layers.append(Linear(in_size, out_size, last_use_bias, key=keys[0]))
        else:
            layers.append(Linear(in_size, width_size, use_bias, key=keys[0]))
            for i in range(depth - 1):
                layers.append(Linear(width_size, width_size, use_bias, key=keys[i + 1]))
            layers.append(Linear(width_size, out_size, last_use_bias, key=keys[-1]))
```
etc.  With a few other small changes like defaults for the bias.

Does this sound feasible?  If so, and you are amenable I could create a PR to implement it as practice for working with JAX?  If so and you roughly agree with the approach, just tell me your preferred names for the two arguments.  Alternatives might be `use_hidden_bias` and `use_last_bias` for example.

jlperla:
Closes #290 

@patrick-kidger   Follows your suggested naming convention.  Please tell me if there is anything else I should add.

### DIFF
--- a/equinox/nn/_composed.py
+++ b/equinox/nn/_composed.py
@@ -60,7 +60,7 @@ class MLP(Module):
         activation: Callable = jnn.relu,
         final_activation: Callable = _identity,
         use_bias: bool = True,
-        use_final_bias: bool = True,        
+        use_final_bias: bool = True,
         *,
         key: PRNGKey,
         **kwargs,
@@ -81,9 +81,9 @@ class MLP(Module):
         - `final_activation`: The activation function after the output layer. Defaults
             to the identity.
         - `use_bias`: Whether to add on a bias to internal layers. Defaults
-            to the True.            
+            to the True.
         - `use_last_bias`: Whether to add on a bias to to the final layer. Defaults
-            to the True.            
+            to the True.
         - `key`: A `jax.random.PRNGKey` used to provide randomness for parameter
             initialisation. (Keyword only argument.)
 
--- a/equinox/nn/_composed.py
+++ b/equinox/nn/_composed.py
@@ -44,8 +44,8 @@ class MLP(Module):
     layers: Tuple[Linear, ...]
     activation: Callable
     final_activation: Callable
-    use_bias: bool = static_field(),
-    use_final_bias: bool = static_field(),
+    use_bias: bool
+    use_final_bias: bool
     in_size: Union[int, Literal["scalar"]] = static_field()
     out_size: Union[int, Literal["scalar"]] = static_field()
     width_size: int = static_field()
--- a/equinox/nn/_composed.py
+++ b/equinox/nn/_composed.py
@@ -44,8 +44,8 @@ class MLP(Module):
     layers: Tuple[Linear, ...]
     activation: Callable
     final_activation: Callable
-    use_bias: bool
-    use_final_bias: bool
+    use_bias: bool = static_field()
+    use_final_bias: bool = static_field()
     in_size: Union[int, Literal["scalar"]] = static_field()
     out_size: Union[int, Literal["scalar"]] = static_field()
     width_size: int = static_field()
--- a/equinox/nn/_composed.py
+++ b/equinox/nn/_composed.py
@@ -44,6 +44,8 @@ class MLP(Module):
     layers: Tuple[Linear, ...]
     activation: Callable
     final_activation: Callable
+    use_bias: bool = static_field(),
+    use_final_bias: bool = static_field(),
     in_size: Union[int, Literal["scalar"]] = static_field()
     out_size: Union[int, Literal["scalar"]] = static_field()
     width_size: int = static_field()
@@ -57,6 +59,8 @@ class MLP(Module):
         depth: int,
         activation: Callable = jnn.relu,
         final_activation: Callable = _identity,
+        use_bias: bool = True,
+        use_final_bias: bool = True,        
         *,
         key: PRNGKey,
         **kwargs,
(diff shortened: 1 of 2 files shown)


jlperla:
Sorry I somehow missed the contributing.md.  I think I have now installed and applied the precommit scripts so hopefully CI passes.

equinox/nn/_composed.py
```diff
@@ -44,6 +44,8 @@ class MLP(Module):
     layers: Tuple[Linear, ...]
     activation: Callable
     final_activation: Callable
```
jlperla, line 46: @patrick-kidger i wasn't sure if these should have a `= static_field()` or not.  Tests pass, but I  don't have a mental model of what those are for.

patrick-kidger: Hey there! Thanks for the PR.

So `static_field` is used to mark anything that should never interact with JAX transforms, like jit, grad etc. This is why it's on `in_size` for example -- this is just a bit of metadata, which we don't want to get promoted to a JAX array.

These deliberately don't have a `static_field`, as it is possible that someone may pass in a learnt activation function (e.g. `eqx.nn.PReLU`). In this case we do want JAX to be able to see whatever parameters are part of the activation function.

On which note -- you probably *do* want to put `static_field` on the extra bias values, which again are just extra metadata. (Other than that this PR LGTM.)

jlperla: I will add those then.

I think this is coming together in my head a bit. Thanks for your patience as I learn JAX. So these tag something so you can either put the equivalent to a `ZeroTangent`  (https://juliadiff.org/ChainRulesCore.jl/stable/api.html#ChainRulesCore.ZeroTangent ) either directly or by just filtering those out and stopping the recursion when building the tangent type? And hence callables like activation functions (or equinox modules) may themselves have nodes?

patrick-kidger: A little bit, yes!

So the thing that's really going on here is what JAX calls "pytrees". These are any nested containers of dictionaries, tuples, lists, or user-registered custom nodes.

For example:
```python
pytree = [0, (1, 2)]
treedef = jax.tree_util.tree_structure(pytree)
print(treedef)  # PyTreeDef([*, (*, *)])
```
Likewise if you have no containers, then you just get a trivial pytree:
```python
pytree = 0
treedef = jax.tree_util.tree_structure(pytree)
print(treedef)  # PyTreeDef(*)
```
And in fact the types of the "leaves" of the tree is irrelevant to the structure:
```python
pytree = ["some string", object(), {"a": 0}]
treedef = jax.tree_util.tree_structure(pytree)
print(treedef)  # PyTreeDef([*, *, {"a": *}])
```

Every time you pass an object across any JAX API boundary, it will look over the leaves of the pytree structure -- e.g. when you do
```python
@jax.grad
def f(x):
    ...

x = (1.0, {"a": 2.0})
f(x)
```
then `x` is iterated over as a pytree, and gradients computed with respect to every leaf.

Alright, how does Equinox fit into this?

In Equinox, every module is *also* a pytree node. (One of those "user-registered custom nodes" from earlier.) For example `mlp = eqx.nn.MLP(...)` is a pytree -- this is what makes it possible to pass Equinox models across JAX's JIT/grad/etc. API boundaries, and get gradients computed wrt its leaves, for example.

And indeed, whether something is a pytree node and whether something is callable are unrelated concepts. For example this `mlp` is both a node and it is callable. (Contrast e.g. a dictionary, which is a node but not callable, or a regular function, which is callable but pytree-wise is "trivial" without subnodes.)

And finally, what does `static_field` do? It simply marks that this attribute is part of the structure of the pytree, and not a node:
```python
class Foo(eqx.Module):
    attribute_node: int
    attribute_structure: int = eqx.static_field()

foo = Foo(attribute_node=1, attribute_structure=2)
leaves, treedef = jax.tree_util.tree_flatten(foo)
print(leaves)  # [1]
print(treedef)  # PyTreeDef(CustomNode(Foo[('attribute_node',), ('attribute_structure',), (2,)], [*]))
```
Thus, it is "hidden" from any JAX transformations.

patrick-kidger: One final addendum -- in an effort to provide a slightly more user-friendly API, Equinox provides some operations like `equinox.filter_{jit, grad, ...}` that are analogous to the original `jax.{jit, grad, ...}`. The difference is simply that Equinox automatically looks over all the leaves of the pytree, and only performs the transformation wrt the ones that are JAX or NumPy arrays. (And basically auto-static's everything else.)

As long as you're only using the `filter_*` verisons of transforms, this actually makes it possible to leave off `static_field` basically all of the time. (I still include the `static_field`s here just for the sake of anyone who happens not to be using these versions of the transforms.)

Tbh I think the fact that `jax.jit` does not already act like `equinox.filter_jit` (or `grad` like `filter_grad` etc.) was just a small mistake made early in JAX's API design, when things hadn't settled as much as they have now. In practice I usually just recommend that end users use `filter_{jit, grad, ...}` everywhere so that they don't need to worry about the details.

jlperla:
@patrick-kidger   I think this is good to merge now.  I ran into a different issue but it appears to be unrelated to my change.

patrick-kidger:
Alright, this looks good to me! Thank you for the contribution.
I'll merge this shortly, and include it in the next release.

PR #363, opened 2023-05-19
Added flattening with key paths

patrick-kidger:
- Fixed a docstring
- Added support for flattening with keypaths.


sachith-gunasekara:
Could you please clarify how this could be used?

Issue #370, opened 2023-05-26
Possible improvement to serialization documentation


bytbox:
I'm not completely sure this is a good idea, but... for me, a common desire is "package up, in one file, both the pytree leaves, and instructions for creating the pytree skeleton". Concretely, I might have a function
```python
def make_model(key, **hyperparameters):
    return MyModel(depth=hyperparameters['depth'], key=key)
```
and then I would like to save in a single file both the dict `hyperparameters` and the weights of `MyModel`.

This is of course easy, but it's not immediately obvious from the documentation that it's possible at all (at least it wasn't to me). I think it's worth adding to the "Tricks" page (or perhaps FAQ) a section explaining that it _is_ possible (because `tree_serialise_leaves` is happy to dump to any `BinaryIO`), with an example like:
```python
def save_model(filename, model, hyperparameters)
  leaves = io.BytesIO()
  eqx.tree_serialise_leaves(leaves, model)
  leaves.seek(0)
  with open(filename, 'wb') as f:
    pickle.dump((hyperparameters, leaves), f)
```

patrick-kidger:
This makes sense to me! I'd be happy to take a pull request on this -- I could see a more fully-featured example being a good way to demonstrate how to make this pattern work in full. FWIW this is the approach I take in my own code. The only requirement is that the hyperparameters themselves must be serialisable, e.g. not a Python function.

For completeness I note that there is also the [Orbax](https://github.com/google/orbax) library for serialisation. They support asynchronous serialisation, and sharded (multi-device) JAX arrays, which are useful for bigger models. It looks like their `orbax.checkpoint.PyTreeCheckpointer()` operates on essentially the same premise as `eqx.tree_{de,}serialise_leaves` in that it expects a PyTree structure at deserialisation time, so probably the pattern used here would apply to them too.

patrick-kidger:
Closing as done!

PR #374, opened 2023-05-27
Add example of serialisation of hyperparameters (#370)

bytbox:
A first draft, I guess. As discussed here: https://github.com/patrick-kidger/equinox/issues/370#issuecomment-1564513099. I tried to clean things up a bit, and avoid encouraging bad practices like overuse of pickle.

One warning: I couldn't get the pre-commit hooks to run correctly, so I just disabled them. I'm assuming they don't operate on ipynb files! If they do, let me know and I'll go back and try harder.


patrick-kidger (commented):
Thanks for the contrbution!

The pre-commit hooks do run on `.ipynb` files. What kind of issue are you getting with them?

Reading over the example, it looks really good! The trick with `\n` is pretty neat.

Typo: `make(key=jr.PRNGKey(0), **params)` inside `load` should have a `model = ...`.

Nit: consistently used "Equinox" over "equinox" or "`equinox`".

Suggestion: the approach you're using here relies on the hyperparameters themselves being serialisable. It'd be worth explicitly mentioning that this is a requirement.

bytbox:
I added a commit, with everything fixed. I don't remember what error I was getting installing the pre-commit hooks, but I think it was related to using a virtual environment. Just installing everything globally worked fine.

Thanks!

bytbox:
Done and pushed!

patrick-kidger:
Marvellous -- merged! Thank you for contributing.

PR #383, opened 2023-06-07
Added an optional dtype argument to BatchNorm.

Benjamin-Walker:
Was encountering the error "Old and new values have different structures." when using BatchNorm with complex valued inputs. Added the ability to specify the dtype of the buffers passed to StateIndex. 

I am slightly unsure on the typing to use for dtype (as using jnp.dtype raises a style error during pyright's pre-commit hook). I have currently opted for Num from jaxtyping.

### DIFF
--- a/equinox/nn/_batch_norm.py
+++ b/equinox/nn/_batch_norm.py
@@ -3,7 +3,7 @@ from typing import Hashable, Optional, Sequence, Tuple, Union
 import jax
 import jax.lax as lax
 import jax.numpy as jnp
-from jaxtyping import Array, Bool, Float
+from jaxtyping import Array, Bool, Float, Num
 
 from .._module import Module, static_field
 from ._stateful import State, StateIndex
@@ -60,7 +60,7 @@ class BatchNorm(Module):
         channelwise_affine: bool = True,
         momentum: float = 0.99,
         inference: bool = False,
-        dtype: jnp.dtype = jnp.float64,
+        dtype: Num = jnp.float64,
         **kwargs,
     ):
         """**Arguments:**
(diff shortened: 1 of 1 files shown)


equinox/nn/_batch_norm.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         channelwise_affine: bool = True,
         momentum: float = 0.99,
         inference: bool = False,
+        dtype: Num = jnp.float32,
```
patrick-kidger, line 63: `Num` is only allowed to be used in conjunction with an array and shape, e.g. `Num[Array, "foo"]`.

Surprisingly, there actually isn't a valid dtype annotation that I know of. For now I'd just leave this unannotated.

equinox/nn/_batch_norm.py
```diff
@@ -139,7 +144,7 @@ def __call__(
             def _stats(y):
                 mean = jnp.mean(y)
                 mean = lax.pmean(mean, self.axis_name)
-                var = jnp.mean((y - mean) ** 2)
+                var = jnp.mean(abs(y - mean) ** 2)
```
patrick-kidger, line 147: Slightly better would be `(y - mean) * jnp.conj(y - mean)`.

Benjamin-Walker:
Hey! No worries, made the changes and rebased into a single commit locally, but still not quite got the hang of doing so such that it updates the pull request properly! Played about a bit too much and going to need to open a new pull request now, apologies!

patrick-kidger:
Haha sure. FWIW the commands for this are something like `git reset --soft HEAD~5; git commit; git push -f`

Benjamin-Walker:
Thank you, will keep a note of that somewhere. Was trying to use `git rebase -i HEAD~5`, squashing all my commits into one, and then `git push -f`. Updated my forked branch but didn't transfer across to this pull request.

PR #407, opened 2023-07-04
better custom autodiff

patrick-kidger:
- Now using symbolic zeros in `filter_custom_{jvp,vjp}`. This also brings with it an API tweak for backward compatibility. (E.g. `filter_custom_jvp.defjvp -> filter_custom_jvp.def_jvp` for the new behaviour.)
- Removed unnecessary `lax.select` in `eqxi.while_loop` when not vmap'ing. This should improve speed slightly when not vmap'ing.
- Improved speed of backprop through `eqxi.while_loop` buffers, in particular avoiding https://github.com/google/jax/issues/10197. In particular this can be used to reduce some programs with quadratic runtime down to just linear runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ _In other words, why should you care? Because Equinox is really simple to learn,
 pip install equinox
 ```
 
-Requires Python 3.9+ and JAX 0.4.11+.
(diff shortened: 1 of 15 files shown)


patrick-kidger:
TODO: benchmark effect on Diffrax.

PR #409, opened 2023-07-05
Apply functools.wraps to Module methods to preserve docstrings

bowlingmh:
Methods of `equinox.Module` lose their `__doc__`, `__name__`, `__module__`, etc.

The methods all get wrapped with`jax.treel_util.Partial`, which doesn't use `functools.wraps` likely due to the signature change making the docstring usually inapplicable.  In this case we're just doing instance-binding so these attributes remain relevant.  Simply wrapping `Partial` does not work alone as it sets the `__wrapped__` attribute, which causes inspection of the signature to think a `self` argument is still expected.

### DIFF
--- a/equinox/_module.py
+++ b/equinox/_module.py
@@ -101,7 +101,9 @@ class _wrap_method:
     def __get__(self, instance, owner):
         if instance is None:
             return self.method
-        return jtu.Partial(self.method, instance)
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Thanks for the contribution!
What's the reason that `__wrapped__` is deleted?

PR #445, opened 2023-08-12
Remove deprecated diffrax.NoAdjoint from Score-Based Diffusion example

homerjed:
Remove `diffrax.NoAdjoint` from the [Score-Based Diffusion](https://github.com/patrick-kidger/equinox/blob/main/examples/score_based_diffusion.ipynb) example.


patrick-kidger:
LGTM! Thank you for the fix.

Issue #448, opened 2023-08-15
Passing on `State` to BatchNorm when it is integrated with `nn.Squential`


paganpasta:
Hi,

This in continuation to the effort of bringing eqxvision up to date with changes in equinox (https://github.com/paganpasta/eqxvision/issues/72).

One of the changes required now are to pass the state explicitly to the Normlayer's call. 
Quite often the models are recursive collections of bunch of layers with BatchNorm integrated in them somewhere. For example,
ResNet utilises mainly 4 blocks where each block is implemented as a Sequential and within each sequential block there is a BatchNorm.

One solution I see is to either modify the `equinox.nn.Sequential` to accomodate an optional State input or simply implement a modified `Sequential` in the Eqxvision repository.

@patrick-kidger, I wanted to seek your advice on the best way to go about and supporting such scenarios before making any drastic changes.

patrick-kidger:
Heyhey! I think we could probably do something like:
```python
class StatefulLayer(eqx.Module):
    @abc.abstractmethod
    def __call__(self, x, state):
        ...

class BatchNorm(StatefulLayer):
    ...

class Sequential(eqx.Module):
    ...

    def __call__(self, x, state=sentinel):
        for layer in self.layers:
            if isinstance(layer, StatefulLayer) and state is not sentinel:
                x, state = layer(x, state)
            else:
                x = layer(x)
        if state is sentinel:
            return x
        else:
            return x, state
```

In other words: (a) use the same trick that `LayerNorm` uses to optionally pass through stateful input, and (b) add an explicit way of checking whether a layer should take an extra state input.

I'd be happy to take a PR on this to Equinox if you want. (I'd also be happy to see this appear in Eqxvision; I don't have strong feelings either way.)

patrick-kidger:
Yup! Problem solved. :)

PR #449, opened 2023-08-15
first draft for BN support in nn.Sequential

paganpasta:
Added the StatefulLayer as per discussion https://github.com/patrick-kidger/equinox/issues/448

One doubt I have is the `inference` parameter, it is not utilised by other norms. 
If BN needs to be easily interchangable with other norms, we might want to either remove `inference` parameter from BN and only rely on the self.inference flag or
add `inference` params to other normalisation layers as well. 

Also, do all Stateful layers utilise the `inference` flag?

### DIFF
--- a/equinox/nn/__init__.py
+++ b/equinox/nn/__init__.py
@@ -34,4 +34,8 @@ from ._pool import (
 )
 from ._rnn import GRUCell as GRUCell, LSTMCell as LSTMCell
 from ._spectral_norm import SpectralNorm as SpectralNorm
-from ._stateful import State as State, StateIndex as StateIndex
+from ._stateful import (
+    State as State,
+    StatefulLayer as StatefulLayer,
+    StateIndex as StateIndex,
+)
--- a/equinox/nn/_batch_norm.py
+++ b/equinox/nn/_batch_norm.py
@@ -6,11 +6,11 @@ import jax.lax as lax
 import jax.numpy as jnp
 from jaxtyping import Array, Bool, Float
 
-from .._module import field, Module
-from ._stateful import State, StateIndex
+from .._module import field
+from ._stateful import State, StatefulLayer, StateIndex
 
 
-class BatchNorm(Module):
+class BatchNorm(StatefulLayer):
     r"""Computes a mean and standard deviation over the batch and spatial
     dimensions of an array, and uses these to normalise the whole array. Optionally
     applies a channelwise affine transformation afterwards.
--- a/equinox/nn/_composed.py
+++ b/equinox/nn/_composed.py
@@ -3,8 +3,7 @@ from typing import (
     Any,
     Literal,
     Optional,
-    Protocol,
-    runtime_checkable,
+    overload,
     Union,
 )
 
@@ -12,9 +11,11 @@ import jax.nn as jnn
 import jax.random as jrandom
 from jaxtyping import Array, PRNGKeyArray
 
+from .._custom_types import sentinel
 from .._doc_utils import doc_repr
 from .._module import field, Module
 from ._linear import Linear
+from ._stateful import State, StatefulLayer
 
 
 _identity = doc_repr(lambda x: x, "lambda x: x")
@@ -122,12 +123,6 @@ class MLP(Module):
         return x
 
 
-@runtime_checkable
-class _SequentialLayer(Protocol):
-    def __call__(self, x: Array, *, key: Optional[PRNGKeyArray] = None) -> Array:
-        ...
-
-
 class Sequential(Module):
     """A sequence of [`equinox.Module`][]s applied in order.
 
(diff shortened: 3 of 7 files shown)


equinox/nn/_batch_norm.py
```diff
@@ -10,7 +11,29 @@
 from ._stateful import State, StateIndex
 
 
-class BatchNorm(Module):
+class _StatefulLayer(Module):
```
patrick-kidger, line 14: Let's put this in `nn/_stateful.py`, and additionally make it public API. (i.e. provide it in `__init__.py`). This should be something an end user could override in principle.

(I'm happy to hold off on documenting it for now, though.)

equinox/nn/_batch_norm.py
```diff
@@ -10,7 +11,29 @@
...
+    current state as arguments and returns the output data and updated state.
+    """
+
+    @abc.abstractmethod
+    def __call__(
+        self,
+        x,
+        state: State,
+        *,
+        key: Optional["jax.random.PRNGKey"] = None,  # pyright: ignore
```
patrick-kidger, line 30: This should now be `Optional[jaxtyping.PRNGKeyArray]`, see the other examples elsewhere in Equinox.

equinox/nn/_composed.py
```diff
@@ -136,32 +140,62 @@ class Sequential(Module):
         Activation functions can be added by wrapping them in [`equinox.nn.Lambda`][].
     """
 
-    layers: tuple[_SequentialLayer, ...]
+    layers: tuple[Union[_SequentialLayer, _StatefulLayer], ...]
```
patrick-kidger, line 143: I think I'd suggest just removing the `_SequentialLayer` protocol. IMO with this update it becomes more confusing than it is useful.

paganpasta: Wouldn't that imply that layers now only accepts `StatefulLayer` ?

patrick-kidger: I mean to remove both from type annotations: just have `layers: tuple`.

equinox/nn/__init__.py
```diff
@@ -34,4 +34,8 @@
 )
 from ._rnn import GRUCell as GRUCell, LSTMCell as LSTMCell
 from ._spectral_norm import SpectralNorm as SpectralNorm
-from ._stateful import State as State, StateIndex as StateIndex
+from ._stateful import (
+    State as State,
+    StatefulLayer,  # noqa: F401
```
paganpasta, line 39: A bit unsure about this.

patrick-kidger: You need `StatefulLayer as StatefulLayer`. The `X as X` pattern is used to include that it's public API, despite being imported into this file.

patrick-kidger:
Also `SpectralNorm` is also stateful (the only other stateful layer built-in to Equinox), so it should be marked as such.

equinox/nn/_composed.py
```diff
@@ -136,32 +139,58 @@ class Sequential(Module):
...
+                x, state = layer(x, state=state, key=key)
+            else:
+                x = layer(x, key=key)
+        if state is sentinel:
+            return x
+        else:
+            return x, state
 
-    def __getitem__(self, i: Union[int, slice]) -> _SequentialLayer:
+    def __getitem__(self, i: Union[int, slice]):
```
paganpasta, line 193: @patrick-kidger Not sure about the return type here.

patrick-kidger: I think it's always going to be a `Module` -- it's indexing or slicing `self.layer: Sequence[Module]`.

paganpasta:
Surprisingly, the pre-commit passed on my local.

patrick-kidger:
Oh sorry, I only just saw your comment here. I think just adding a `# pyright: ignore` is fine here.
(Can you squash your commits together into just one commit please?)

equinox/nn/_composed.py
```diff
@@ -136,32 +139,58 @@ class Sequential(Module):
         Activation functions can be added by wrapping them in [`equinox.nn.Lambda`][].
     """
 
-    layers: tuple[_SequentialLayer, ...]
+    layers: tuple
 
-    def __init__(self, layers: Sequence[_SequentialLayer]):
+    def __init__(self, layers: Sequence[Module]):
```
patrick-kidger, line 144: `Module` isn't quite accurate here -- we can put non-modules here too. I think `Callable` should be used here. (`from collections.abc import Callable`)

patrick-kidger: Also, now that `_SequentialLayer` is unused, can it be deleted?

equinox/nn/_composed.py
```diff
@@ -136,32 +139,58 @@ class Sequential(Module):
...
+        key: Optional[PRNGKeyArray] = None,
+    ) -> Union[Array, tuple[Array, State]]:
         """**Arguments:**
 
-        - `x`: Argument passed to the first member of the sequence.
-        - `key`: A `jax.random.PRNGKey`, which will be split and passed to every layer
-            to provide any desired randomness. (Optional. Keyword only argument.)
+        - `x`: passed to the first member of the sequence.
+        - `state`: If provided, then it is passed on to any layer which implements
+            `equinox._batch_norm._StatefulLayer`.
```
patrick-kidger, line 168: "then it is passed to any layer which subclasses `eqx.nn.StatefulLayer`, which is used purely to indicate that this layer accepts and returns an extra `state` argument"

equinox/nn/_composed.py
```diff
@@ -136,32 +139,58 @@ class Sequential(Module):
...
+                x, state = layer(x, state=state, key=key)
+            else:
+                x = layer(x, key=key)
+        if state is sentinel:
+            return x
+        else:
+            return x, state
 
-    def __getitem__(self, i: Union[int, slice]) -> _SequentialLayer:
+    def __getitem__(self, i: Union[int, slice]) -> Module:
```
patrick-kidger, line 193: Likewise the return annotation here should actually be `Callable`, not `Module`, I've realised.

equinox/nn/_stateful.py
```diff
@@ -145,3 +146,24 @@ def tree_unflatten(cls, keys, values):
...
+
+class StatefulLayer(Module):
+    """An abstract base class representing a stateful layer within a neural network.
+
+    Stateful layers maintain internal state across calls, which can be useful for tasks
+    such as BatchNormalisation.
+
+    Subclasses must implement the `__call__` method that takes input data and the
+    current state as arguments and returns the output data and updated state.
+    """
```
patrick-kidger, line 159: Can you (a) mention that right now this is used solely to declare the API for use with Sequential, and (b) add this base class to the documentation?

patrick-kidger:
And merged! Thank you for the implementation, I'm glad to have it in.

PR #455, opened 2023-08-20
Added an example for a Deep Convolutional Generative Adversarial Network

ahmed-alllam:


patrick-kidger:
Hey there! Thanks for the contribution. This looks very cleanly implemented.

Various nits:
- The authors are currently listed as Alec Radford etc. Whilst they wrote the original paper, "authorship" here usually means the one who wrote the code. Put another way: they haven't given permissions to have their names affiliated with Equinox, so we including their names makes it look like they endorse our work.
- Prefer `import jax.random as jr` rather than using `jax.random` directly.
- It looks like you have some unused code for de/serialising the models. Can this be removed?
- Can we avoid using PyTorch at the very end? (When generating the images.) In particular it's a bit weird that we're going JAX->Python->PyTorch->NumPy

ahmed-alllam:
Hi Patrick!

Thanks for the notes. I have added the BibTeX reference to the original paper, and did some code refactoring to fix the issues you mentioned.

patrick-kidger:
Okay, I'm going over this now.

First of all, I really like this example! It's very well-written, thank you. The following are small nits only.

- Can you style the bibtex reference in the same as the other examples? I.e. with `??? Reference`, ```` ```bibtex ...``` ````, and an arXiv link.
- Why are `Generator.{conv_layers, output_layers}` separate? I think they can be combined into a single list. Likewise in the discriminator.
- `jax.nn.relu` can't be used a type hint, as it's not a type. (Which is unfortunate, as it's quite clear what is meant.) I'd suggest using `(from collections.abc import) Callable` instead.
- Let's not use default parameters for the models (`input_shape=latent_size; key=PRNGKey()` etc.) I quite like the mantra that "default parameters are a root of unexpected behaviour", and as these models are only instantiated once then I don't think they're needed here.
- Best practice is to avoid doing things like `input_shape[2]`, in favour of e.g. `height, width, channels = input_shape`. This serves to assert that `input_shape` really is of length 3, and the semantic meaning of each component. Likewise for e.g. `batch[0]`.
- Why is the discriminator using `input_shape[1]` for the size of many of its channels? This is the width of the image, I think?
- Can `generator_losses`, `discriminator_losses` be moved inside `train`? I'd prefer not to have `train` be side-effecting.
- As `real_labels` and `fake_labels` are constants, would it be worth moving them inside the JIT'd region?
- I'd prefer to train by iterating over steps, rather than epochs. The former is usually better practice. (C.f. some of the other Equinox examples.)

As you can see I'm being fairly pernickity, as I think it's important to get these kinds of pedagogical examples as high-quality as possible!

ahmed-alllam:
Thank you, Patrick, for your valuable feedback!

I've incorporated your suggestions and addressed all the minor issues in the code example. Notably, I've replaced the epoch-based training loop with step-based implementation to align with the conventions followed in other examples. I'd be more interested in hearing your perspective on how this approach is generally perceived as a best practice.

patrick-kidger:
And merged! Thank you for contributing.

"Epochs" depend on the size of your dataset. If you were to use two differently-sized datasets and train a model on both, then expressing "we trained the model for n epochs on each dataset" would not be a fair comparison: it should be " we trained the model for n steps on each dataset", to more accurately reflect that amount of training that has occured + the amount of computational work that was required.

PR #460, opened 2023-08-25
Updating error message for operation on non-pytree leaves

LouisDesdoigts:
Changes error for operation on non-pytree leaves to a `TypeError`, updated the error message.


LouisDesdoigts:
I updated the raised error to a `TypeError` as that seems more appropriate. Let me know if you are happy with that and the error message, otherwise I can change it!

patrick-kidger:
LGTM! Thank you for the contribution :)

PR #463, opened 2023-08-28
Document mask convention

colehaus:
It seems worth documenting the convention explicitly since, unless I'm missing something, it's the opposite of e.g. [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html)'s.


patrick-kidger:
LGTM! Thank you. And you're right, this is the opposite convention of PyTorch's.

Issue #472, opened 2023-09-05
AbstractVar's that are also static


fhchl:
I'd like to indicate to downstream users that a _static_ instance attribute of a abstract base class should be overwritten by its sub-classes, e.g. using `AbstractVar`. However, it's not possible to combine them right now as `AbstractVar`s do not allow default values.

Is there any chance one could do something along the following lines?

```python
import equinox as eqx

class Foo(eqx.Module):
    a: AbstractVar[int] = eqx.field(static=True)
    # or
    a: int = eqx.field(static=True, abstract=True)
````

**Comment by fhchl** [comment]
One related annoyance right now is that when the base class defines a field as static, then subclasses can not define any non-default attributes due to the error

```
TypeError: non-default argument '...' follows default argument
```

**Comment by patrick-kidger** [comment]
Hmm. So right now this is best handled by doing:
```python
class AbstractFoo(eqx.Module):
    a: AbstractVar[int]

class ConcreteFoo(AbstractFoo):
    a: int = eqx.field(static=True)
```
as the intention of a `foo: AbstractVar[int]` is just to be able to communicate that "`self.foo` will exist at runtime", nothing more.

Doing something like this is actually surprisingly complicated. (Inside baseball: abstract variables are a lower-level piece of functionality, below both (a) dataclasses and (b) pytrees. This can basically be reduced to how [this dataclass-making function](https://github.com/patrick-kidger/equinox/blob/8c39b21e16f2b3749df068261df321f5176713dd/equinox/_better_abstract.py#L296) should handle abstract variables.) Off the top of my head I'm not sure how we might implement this, but I'd welcome a PR / further discussion.

fhchl:
I tried some simple (but a tad convoluted) ideas in that Pull request. Mainly keeping an additional list of abstractvars that where declared to be static. When the abstract attribute is finally made concrete, we add a `eqx.field(default=...., static=True)`. 

What do you think?

patrick-kidger:
Thanks for writing the PR!
Taking a look at it: right now, my main issue with this approach is that the `AbstractVar` machinery is now coupled to the static-field machinery. I think it's important that these stay separate.

[...]

PR #473, opened 2023-09-06
Allow Abstract vars to be static

fhchl:
A draft on how to support abstract attributes that are static as discussed in #472 .

Please don't mind the comments which were just there to keep me sane.

I am not quite sure of the  impact of two specific changes:

1. Moving the parent classes `super().__new__` call to the end of the the new `__new__`
2. This required using `if name not in namespace` instead of  `if name not in cls.__dict__` on line 290.

Added some tests. Everything seems to run fine so far.

### DIFF
--- a/equinox/_better_abstract.py
+++ b/equinox/_better_abstract.py
@@ -206,25 +206,52 @@ def _is_concretisation(sub, super):
     return is_subhint(sub_args, sup_args)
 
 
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Closing as per discussion over in #472, that we'd like to pursue a different strategy.

PR #474, opened 2023-09-07
Minor help comment fix

knyazer:
Fix the name of the function in the comment, from the one that was (probably) used in older versions, to the currently existing one.

I was having an issue related to the incorrect construction of PyTree, thus was reading the source code of the implementation. Noticed that the name in the comment was incorrect.


patrick-kidger:
Good catch! Thank you for the fix :)

PR #491, opened 2023-09-11
module_update_wrapper now looks at __wrapped__, rather than requiring an extra argument. It no longer mutates its input.

patrick-kidger:


patrick-kidger:
Closing in favour of #494. This branch had some history-rewriting shenanigans due to #493.

Issue #469, opened 2023-09-02
Q: Idiomatic way to do weight tying/sharing?


colehaus:
A question rather than an issue.

It seems relatively straightforward to mimic the behavior of weight sharing by: `tree_at`ing or `tree_map`ing after initialization to copy initial weights over and then combining gradients to ensure they stay synchronized. But this, AIUI, would be mostly silly because you wouldn't reduce memory usage at all.

It in fact seems there is no realistic post-hoc `tree_map`, etc you can do to turn a non-shared architecture into one with shared weights (in a performance-relevant way)? Because a shared-weight model corresponds to a fundamentally different Pytree and there is no notion of shared reference/alias in the JAX idiom?

The best or only approach is to manually pass shared weights down to the pieces that need them at `__call__` time? e.g. if `Parent` has two children that are supposed to share weights, `Parent` could keep `weights` as a class attribute and pass it to each child when `__call__`ing?

I see some discussion of [how do this in Flax](https://github.com/google/flax/discussions/1264), but it looks like Flax maybe already implements things by registering and threading parameters around so it's more straightforward to do weight sharing in that approach?

(For a little more context, I'm implementing a version of the T5 paper and thinking about sharing weights across the encoder and decoder halves. But it looks like that might entail pretty substantive changes to the model architecture.)

patrick-kidger:
So your approach of storing them on the parent,and passing them at `__call__` time, is the one I use and recommend.

> Because a shared-weight model corresponds to a fundamentally different Pytree and there is no notion of shared reference/alias in the JAX idiom?

Yup, exactly! At that point it's not a tree any more, but a DAG.

That said, at some point I'd like to see if I can find a better way of handling this. (Although I don't have any elegant solutions at the moment; I'd welcome any thoughts.)

You may also like `eqx.tree_check`, which can be used to check that you haven't accidentally got repeated nodes in the tree.

colehaus:
Okay, that makes sense. Thanks!

colehaus:
It's a little gross but preliminary tests suggest that something like this works as a way to avoid duplicating the core `__call__` logic for each `Module` (once in the non-shared version and once in the shared version):

```py
A = TypeVar("A", bound=Callable[..., Any])


class FakeCallable(Generic[A]):
    def __new__(cls, *args: Any, **kwargs: Any) -> A:
        return super().__new__(cls)  # pyright: ignore

    def __init__(self, entries: Mapping[str, Any], call: type[A]):
[...]

PR #497, opened 2023-09-12
add `filter_checkpoint`

dlwh:
As discussed in #469

### DIFF
--- a/equinox/_ad.py
+++ b/equinox/_ad.py
@@ -969,3 +969,82 @@ if getattr(typing, "GENERATING_DOCUMENTATION", False) and not TYPE_CHECKING:
 
     filter_custom_jvp.__doc__ = _filter_custom_jvp_doc
     filter_custom_vjp.__doc__ = _filter_custom_vjp_doc
+
+
+def filter_checkpoint(
+    fun: Callable[_P, _T] = sentinel,
+    *,
+    prevent_cse: bool = True,
+    policy: Optional[Callable[..., bool]] = None,
+) -> Callable[_P, _T]:
+    """Filtered version of `jax.checkpoint`.
+
+    Gradient checkpointing is a technique for reducing memory usage during
+    backpropagation, especially when used with reverse mode automatic differentiation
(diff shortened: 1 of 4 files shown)


equinox/_ad.py
```diff
@@ -940,3 +940,82 @@ def filter_custom_vjp(fn):
...
+        @ft.partial(
+            jax.checkpoint,  # pyright: ignore
+            prevent_cse=self._prevent_cse,
+            policy=self._policy,
+            static_argnums=(0,),
+        )
+        def fun_checkpoint(_static, _dynamic):
+            _args, _kwargs = combine(_static, _dynamic)
+            out = self._fun(*_args, **_kwargs)
+            _dynamic_out, _static_out = partition(out, _is_struct)
```
patrick-kidger, line 1010: Is `_is_struct` really the criterion we need here? (On both the input and the output.) I think just `eqx.is_array` would be correct?

patrick-kidger (commented):
Nits aside, LGTM!

equinox/_ad.py
dlwh: yeah you're right. will fix. do you like is_array or is_array_like here? i feel like probably is_array is fine since you never need to checkpoint scalars (except with some truly massive vmaps or xmaps or something)

dlwh:
rfal

equinox/_ad.py
patrick-kidger: I think `is_array` is probably best here. (Equinox generally prefers to distinguish arrays from arraylikes.)

dlwh: ok, so all done then!

patrick-kidger:
And merged! Thanks for contributing -- this will appear in the upcoming v0.11.0 release.

(What does RFAL mean, by the way?)

dlwh:
Great!

"ready for a look". I thought it was a google-ism, but maybe it was peculiar to my startup's teams of xooglers.

patrick-kidger:
Ah, ty! The one I'm familiar with is "PTAL" -- please take a look :)

PR #500, opened 2023-09-16
Proximal Policy Optimization + Brax

knyazer:
A relatively simple implementation of the PPO for continuous environments. 

I am not sure whether this example qualifies as an "example" since I have used some features that are considered advanced (like eqx.error_if and custom filter_{} implementation) so I am wondering what you think.

I have tried to keep the boundary of correctness/conciseness somewhat on the conciseness side: for example, I don't use jaxtyping, since the function declarations become unreadable. Maybe it makes sense to establish some sort of a guideline on "how to write an example" so that other contributors don't have to make such decisions on their own.

The only dependencies of the notebook are JAX, Brax, and Equinox, and the whole execution takes about 6 minutes on Nvidia T4.


patrick-kidger:
Hey there! Thanks for contributing -- I think this looks really cool. To answer your first question, I'm happy to have examples that use some of the more advanced parts of the API.

My first comment is that this example is really big. An important concern on my end is that anything I host has to be something I can maintain, and I think this is probably a bit beyond that. Would it be possible to winnow this down to the bits that you feel are most important? (E.g. I can see various spots where you've introduced unused optional functionality; I'm guessing this example is adapted from a codebase you have somewhere?)

Alternatively, if you'd prefer to package this up as a standalone GitHub repository, then I'd be happy to add a link to it from the documentation.

Let me know what you think and we'll take it from there.

Other than that -- I can see you've introduced `filter_scan` and `filter_cond`. I'd be happy to take these as PRs against Equinox itself. I've started needing this functionality myself, so I think it's probably time to add all the filtered higher-order-primitives.

knyazer:
I was originally contemplating removing wrappers functionality, but I thought that it would be nice to have it as an example of how wrappers could be implemented. Now, I have tried to reduce the code base, but without hurting the structure significantly. I am wondering if it is still too complex for an example?

I think that PPO might be indeed a bit too complex of a thing to implement as an "introductory" example, since it uses a lot of data manipulation, which does not have high pedagogical value. However, I observed that there are very few self-contained JAX PPO implementations out there, and, in comparison to them, Equinox filtered transformation allow the code to be much more concise. Thus, while the pedagogical value of the example is questionable, this example might be considered as a semi-realistic use case for Equinox, showing that it is better (more readable) than most of the alternatives. 

Since if you add this notebook to the repository, you will have to support it, it seems like a good idea at some point to separate "supported" and "unsupported" notebooks: so that, while having some examples that are guaranteed to work, there are also many more examples, that are of decent quality (like, you looked over them, and approved them) but are not continuously supported. This would probably significantly reduce your load, and allow you to focus more on the library itself. I am not sure though how this could be done.

> Hey there! Thanks for contributing -- I think this looks really cool. To answer your first question, I'm happy to have examples that use some of the more advanced parts of the API.
> 
> My first comment is that this example is really big. An important concern on my end is that anything I host has to be something I can maintain, and I think this is probably a bit beyond that. Would it be possible to winnow this down to the bits that you feel are most important? (E.g. I can see various spots where you've introduced unused optional functionality; I'm guessing this example is adapted from a codebase you have somewhere?)
> 
> Alternatively, if you'd prefer to package this up as a standalone GitHub repository, then I'd be happy to add a link to it from the documentation.
> 
> Let me know what you think and we'll take it from there.
> 
> Other than that -- I can see you've introduced `filter_scan` and `filter_cond`. I'd be happy to take these as PRs against Equinox itself. I've started needing this functionality myself, so I think it's probably time to add all the filtered higher-order-primitives.

patrick-kidger:
Yeah, I'm afraid this is still way too large.

PPO definitely doesn't need this kind of complexity, though.  [Take a look at this example](https://github.com/patrick-kidger/rl-test), which is a PPO example I threw together a few years ago. It's substantially simpler than what you have here. In particular `algorithms.py` manages to implement all of vanilla PG, A2C, and PPO, in <100 lines.

I still think this would be nice to have as a hosted example, essentially by reproducing the abstractions of the linked repository in Equinox. And we could continue to use Brax for the environment, which is would make things even smaller than the linked repo.

knyazer:
So, I hosted the original notebook [here](https://github.com/knyazer/PPO-Brax-Equinox). I don't really want to split it into separate files, since I think it is a really good (though uncommon) practice for not-that-complex-algorithms to have a self-contained implementation. I frequently encounter that researchers try to achieve "good code quality" but along the way the implementation becomes completely unusable by other people.

I am quite a bit inspired by your implementation of the RL algorithms, and I think I am going to dedicate some time to figure out a way to write well-performing JAX/Brax/Equinox PPO implementation, that would take only a few hundred lines. Would be really nice to have it. 

I didn't completely get what you meant in the last paragraph of your comment: do you mean to make a separate self-hosted repository with "tools" so that implementation in Equinox won't have the annoying data-manipulation complexity?

patrick-kidger:
> I am quite a bit inspired by your implementation of the RL algorithms, and I think I am going to dedicate some time to figure out a way to write well-performing JAX/Brax/Equinox PPO implementation, that would take only a few hundred lines. Would be really nice to have it.

Great! If you put something together in this vein then I'd be happy to host it here.

> I didn't completely get what you meant in the last paragraph of your comment: do you mean to make a separate self-hosted repository with "tools" so that implementation in Equinox won't have the annoying data-manipulation complexity?

I was just referring to the fact that my linked example also includes an implement of cart-pole, a general ODE-solving environment, and an abstract environment. All of that could be cut from a hosted example by doing `brax.envs.create` instead.

We probably don't need Brax for anything beyond that; I think the rest of the implementation can be short (as in my linked example), and has pedagogical value.

(And to be clear, I definitely don't want to hide any of the Equinox-owned part of the implementation in a separate repository.)

PR #505, opened 2023-09-21
Statefulness is now propagated through Sequential layers.

patrick-kidger:


patrick-kidger:
The test failure is spurious -- this is due to a regression in jaxlib==0.4.16; it'll be fixed in the next jaxlib release.

Issue #486, opened 2023-09-09
Serialise and deserialise don't round-trip for some types (sort of)


colehaus:
Suppose you have a pytree with a scalar parameter (e.g. `p` on `Dropout`). Now suppose this scalar is a non-standard type like `jax.dtypes.bfloat16`. When you go to serialize this pytree, it's probably already been JAXified (i.e. the scalar is now something like `bf16[]`) (see Fn1). So you'll hit the:

```py
    if isinstance(x, jax.Array):
        jnp.save(f, x)
```
case in `default_serialise_filter_spec` and the saving will work fine. However, when you're deserializing, the `like` pytree will probably still have a scalar (e.g. a raw `bfloat16`). Since `bfloat16` isn't an instance of `float`, you'll hit the final case in `default_deserialise_filter_spec`:

```py
    if isinstance(x, jax.Array):
        return jnp.load(f)
    elif isinstance(x, np.ndarray):
        return np.load(f)
    elif isinstance(x, (bool, float, complex, int)):
        return np.load(f).item()
    else:
        return x
```
But that means the bytes corresponding to the bfloat16 scalar array are still unread on disk and the next `np.load` or `jnp.load` will read from them instead of reading their correct array. Now, the rest of the reads will be off by one until/unless some error occurs later in the deserialization (in my case, calling `item()` on an array with a size greater than zero).

This is fixable in user code. I just ended up doing:

```py
def deserialise_filter_spec(f: BinaryIO, x: A) -> A:
    if isinstance(x, jax.dtypes.bfloat16):
        return jax.dtypes.bfloat16(jnp.load(f).item())
    else:
        return eqx.default_deserialise_filter_spec(f, x)
```

But figuring out what's going on is sort of involved.

Possible improvements:

- Add [`np.generic`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.generic) to the `isinstance` check in the default spec: `isinstance(x, (bool, float, complex, int, np.generic))`
- Do the `_assert_same` inline with the deserialising instead of afterward to help localize errors. In my case, that means I would've gotten an assertion error in the leaf immediately after the erroneous bfloat16 leaf instead of a deserialization error many leaves later.
- I ended up having to fiddle with `_ordered_tree_map` to print out paths to the leaves to figure out what was going on. It might be nice to include the path in the error message on a failed deserialization.

[...]

Issue #507, opened 2023-09-22
`is_array` returns `True` for instances of `np.generic`


colehaus:
I started looking at PR for  https://github.com/patrick-kidger/equinox/issues/486. While doing so, I noticed another potential issue: [`is_array`'s `isinstance` check includes `np.generic`](https://github.com/patrick-kidger/equinox/blob/76a13bd100a8127c7e5176f2dee9d36c837b46ec/equinox/_filters.py#L21). This does not seem to align with the intention for `np.generic` according to [the documentation](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.generic) where it's described as for scalars.

If I remove `np.generic`, the only test that breaks is [this one](https://github.com/patrick-kidger/equinox/blob/76a13bd100a8127c7e5176f2dee9d36c837b46ec/tests/test_callback.py#L11) in `test_callback`. And the "breakage" there is actually revealing an error in the test IMO.

```py
    def f(x, y):
        assert y is sentinel1
        return (x + 1).astype(np.float32), sentinel2

    out_struct = (jax.ShapeDtypeStruct((), jnp.float32), sentinel2)

    out = eqx.filter_pure_callback(
        f, jnp.array(1.0), sentinel1, result_shape_dtypes=out_struct
    )
```
The `out_struct` suggests that the return type will be a zero-dimensional array. That would be true if `x` were a JAX array, but numpy arrays and JAX arrays have a discrepancy here:

```py
>>> jnp.array(1.0) + 1
Array(2., dtype=float32, weak_type=True)
>>> np.array(1.0) + 1
2.0
```

And the numpy behavior is the relevant one here since:

> The input callback will be passed NumPy arrays in place of JAX arrays and should also return NumPy arrays.

according to [the docs](https://jax.readthedocs.io/en/latest/_autosummary/jax.pure_callback.html).

So `f` in the test is actually returning a `float32` scalar instead of a zero-dimensional `float32` array.

All of which is a somewhat long-winded way of saying that `np.generic` can be safely removed as far as Equinox tests are removed with a slight tweak to this one test:

`return (x + 1).astype(np.float32), sentinel2` → `return np.array((x + 1).astype(np.float32)), sentinel2`

How this change would affect user code is, of course, less clear.

patrick-kidger:
This is actually intentional behaviour. As you've noticed, NumPy uses a different type for scalar arrays, whilst JAX doesn't.

[...]

PR #511, opened 2023-09-22
Handle `np.generic` in ser/de and add path info to deserialisation errors

colehaus:
The changes mentioned in https://github.com/patrick-kidger/equinox/issues/486.

(I'm not actually expecting the Poetry and Nix commit to get merged. Just thought I'd include it with the PR so the info is visible and can maybe save some other Nix and Poetry users a little work.)


colehaus:
(The CI failures seem unrelated to my changes and all tests pass when I run them locally.)

equinox/_serialisation.py
```diff
@@ -1,6 +1,7 @@
 import pathlib
 from collections.abc import Callable
 from contextlib import contextmanager
+from functools import wraps
```
patrick-kidger, line 4: `import functools as ft` is usually the style we've been using in Equinox

equinox/_serialisation.py
```diff
@@ -105,6 +116,14 @@ def default_deserialise_filter_spec(f: BinaryIO, x: Any) -> Any:
...
     elif isinstance(x, (bool, float, complex, int)):
         return np.load(f).item()
+    elif isinstance(x, np.generic):
+        y = np.load(f)
+        try:
+            # `item()` converts back to base types like `float`
+            # so we have to use the constructor
+            return type(x)(y.astype(x.dtype).item())
+        except ValueError:
+            return type(x)(jnp.frombuffer(y, dtype=x.dtype)[0].item())
```
patrick-kidger, line 126: What's going on here? This seems like a lot. What goes wrong with e.g. `type(x)(y)`?

tests/test_serialisation.py
```diff
@@ -79,6 +81,61 @@ def test_leaf_serialisation_file(getkey, tmp_path):
...
+    assert eqx.tree_equal(jax_array, loaded_jax_array)
+
+    tree = jnp.array(1), bfloat16(1), np.float32(1), jnp.array(1)
+    like_tree = jnp.array(2), bfloat16(2), np.float32(2), jnp.array(2)
+
+    # Ensure we can round trip when we start with a scalar
+    eqx.tree_serialise_leaves(tmp_path, tree)
+    loaded_tree = eqx.tree_deserialise_leaves(tmp_path, like_tree)
+    # Can't use `tree_equal` because `is_array` treats `bfloat16` as an array and
+    # ends up trying to call `all()` on a `bool`
```
patrick-kidger, line 130: What's going on here? This sounds like a potential bug with `tree_equal`?

tests/test_serialisation.py
```diff
@@ -79,6 +81,61 @@ def test_leaf_serialisation_file(getkey, tmp_path):
...
+    assert loaded_tree == tree
```
patrick-kidger, line 131: FWIW you may prefer to use `from helpers import shaped_allclose; assert shaped_allclose(loaded_tree, tree)` here.

This helper does strict everything-must-match: shape, dtype, array type, etc. In contrast `__eq__` is fairly permissive, e.g. it happily promotes arraylikes.

(Side note, the name `shaped_allclose` is a bit odd -- I've been meaning to rename it to `tree_identical` or something.)

patrick-kidger (commented):
Thanks! Indeed, if you could remove the flake/poetry changes that would be useful.
(Would they be worth placing in an example / FAQ entry / something, maybe?)

Can you rebase from the `dev` branch? This is what's causing the CI failure.

equinox/_serialisation.py
colehaus: Yeah, it's a bit unfortunate.

The immediate error if you just do `type(x)(y)` is: `No cast function available`.

My understanding: The issue is that `type(y.item())` is just `bytes` because when you `np.load` a `bfloat16` (and presumably other types numpy doesn't know about) it just treats them as bytes. So we have to turn the bytes into a JAX array and then interpret them as the proper dtype. `frombuffer` is the only function I know of for converting bytes into an array, but it "Interpret[s] a buffer as a 1-dimensional array." so we then have to `[0]` to turn that into a 0D array. 

One alternative is:

```py
pos = f.tell()
try:
  type(x)(np.load(f).astype(x.dtype).item())
except ValueError:
  f.seek(pos)
  type(x)(jnp.load(f).astype(x.dtype).item())
```
which is a little more straightforward in some ways, but also involves more IO so is presumably slower.

tests/test_serialisation.py
colehaus: This is related to the https://github.com/patrick-kidger/equinox/issues/507 issue.

```py
        for pytree in pytrees[1:]:
            flat_, treedef_ = jtu.tree_flatten(pytree)
            if treedef_ != treedef:
                return False
            for elem, elem_ in zip(flat, flat_):
                if is_array(elem):
                    if is_array(elem_):
                        if (elem.shape != elem_.shape) or (elem.dtype != elem_.dtype):
                            return False
>                       allsame = (elem == elem_).all()
E                       AttributeError: 'bool' object has no attribute 'all'
```
The `bfloat16`s satisfy the `is_array` check so we end up doing something like `(bfloat16(1) == bfloat16(1)).all()`.

colehaus:
Maybe I can just put the Nix/Poetry changes in a separate PR and then immediately close it. People would presumably be able to find that as long as their search included closed issues/PRs.

equinox/_serialisation.py
patrick-kidger: I don't think this branch should trigger for `bfloat16`:
```python
import jax.numpy as jnp

x = jnp.bfloat16(1)
print(isinstance(x, jnp.generic))  # False
print(type(x))  # It's a JAX array
```
That is, I think this case should already be handled by the `isinstance(x, jax.Array)` path.

tests/test_serialisation.py
patrick-kidger: Likewise -- `bfloat16` is a JAX-only type, and is treated as an array. That means that `.all()` should exist.

For that matter `np.generic.all` also exists, so again this should work.

How does one obtain a raw Python `bool` here?

equinox/_serialisation.py
colehaus: Ah, this is is for handling scalars directly (admittedly probably not many bfloat16 scalars but perhaps makes more sense for np.float32 and whatever other `np.generic` scalars people may want to use):
```py
>>> jnp.bfloat16(1)
Array(1, dtype=bfloat16)
>>> jax.dtypes.bfloat16(1)
1
>>> isinstance(jax.dtypes.bfloat16(1), np.generic)
True
```

colehaus: (And I checked that the branch is indeed triggered. The test does not pass unless I add this branch.)

tests/test_serialisation.py
colehaus: See the other comment. `jax.dtypes.bfloat16` is different than `jnp.bfloat16`.

equinox/_serialisation.py
patrick-kidger: Gotcha; I see the distinction between `jax.{numpy,dtypes}.bfloat16`.
So I've just checked, and `jnp.load` will correctly deserialise a `jax.dtypes.bfloat16`. As such I suspect the correct thing to do here is to have both ser+de run `if is_array_like(x): jnp.{save, load}`. Note that `is_array_like(x) == isinstance(x, (bool, int, float, complex, np.generic, np.ndarray, jax.Array))`, so this is just smushing multiple branches together into a single implementation.

tests/test_serialisation.py
patrick-kidger: I think this is indicative of a bug in JAX:
```
x = np.float32(1)
type(x == x)  # np.bool_
y = jax.dtypes.bfloat16(1)
type(y == y)  # Python bool
```
I'd suggest (a) switching `tree_equal` to use `jnp.all` instead of `.all()`; (b) using `tree_equal` here; (c) opening a bug against JAX.

patrick-kidger:
Placing the nix/poetry changes in a closed PR seems like a reasonable place for now!

tests/test_serialisation.py
colehaus: Just filed a bug on the `ml_dtypes` library: https://github.com/jax-ml/ml_dtypes/issues/105.

patrick-kidger:
Alright, this LGTM. Now merged.
Thank you for figuring this one out, this was a subtle one!

colehaus:
Thanks! Looks like the JAX/ml_dtypes bug has now been addressed as well: https://github.com/jax-ml/ml_dtypes/commit/67236bf2d777e8ed923619ef514d5e015e175f1c

PR #542, opened 2023-10-04
Fix scan-over-layers example in docs

mcbal:
The [scan-over-layers example](https://docs.kidger.site/equinox/tricks/#improve-compilation-speed-with-scan-over-layers) in the tricks section of the docs currently outputs just the result of the final MLP acting on the input. Changing `mlp(x)` to `mlp(_x)` inside the scan function `f` leads to the intended behavior.


patrick-kidger:
Whoops! LGTM, and thankyou for contributing the fix.

PR #545, opened 2023-10-08
Added RMSNorm

packquickly:
Added RMS normalisation, a simplified variant of layer norm which computes

$$ \frac{x}{\sqrt{\varepsilon + \frac{1}{n}\Vert x \Vert^2_2}} \gamma + \beta$$

where $n = \dim(x)$ and $\gamma$ is a learnable array if `use_weight=True`, or $\sqrt{n}$ if `use_weight=False`. $\beta$ is an optional bias term. This has become somewhat popular in transformers for being simpler than layer norm but having similar/better performance.

Originally proposed in [this paper](https://arxiv.org/abs/1910.07467), see [Lucidrain's dicussion](https://github.com/lucidrains/x-transformers#root-mean-square-layer-normalization) about this for more details.

The defaults `use_weight=False` and `use_bias=False` are intentional. This is meant to be a faster and simpler version of layer norm, so leaving these off by default made the most sense to me.

### DIFF
--- a/equinox/nn/_normalisation.py
+++ b/equinox/nn/_normalisation.py
@@ -372,18 +372,11 @@ class RMSNorm(Module):
                 "`RMSNorm(shape)(x)` must satisfy the invariant `shape == x.shape`"
                 f"Received `shape={self.shape} and `x.shape={x.shape}`. You might need "
                 "to replace `rms_norm(x)` with `jax.vmap(rms_norm)(x)`.\n"
-                "\n"
-                "If this is a new error for you, it might be because this became "
-                "stricter in Equinox v0.11.0. Previously all that was required is that "
-                "`x.shape` ended with `shape`. However, this turned out to be a "
-                "frequent source of bugs, so we made the check stricter!"
             )
         inv_rms = jax.lax.rsqrt(jnp.sum(x**2) + self.eps)
-        out = inv_rms * x
+        out = jnp.sqrt(self.dim) * inv_rms * x
         if self.use_weight:
             out = self.weight * out
-        else:
-            out = jnp.sqrt(self.dim) * out
         if self.use_bias:
             out = out + self.bias
         if state is sentinel:
--- a/equinox/nn/_normalisation.py
+++ b/equinox/nn/_normalisation.py
@@ -1,6 +1,7 @@
 import functools as ft
 import warnings
 from collections.abc import Sequence
+from math import prod
 from typing import Optional, overload, Union
 
 import jax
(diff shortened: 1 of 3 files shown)


patrick-kidger:
Nice! This would be good to have.
For `use_weight=False`, I don't think your description and the code quite match up -- you have both a `1/n` and a `gamma = sqrt(n)` factor; I think you only want one of the two. (I also checked [both](https://arxiv.org/pdf/1910.07467.pdf) [papers](https://arxiv.org/pdf/2307.14995.pdf)) and only the one scaling factor seems to be consistent with them too; they basically just do `x/rms(x)`).

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,125 @@ def __call__(
...
+        is passed through unchanged. If `state` is not passed, then just the output is
+        returned.
+        """
+        if x.shape != self.shape:
+            raise ValueError(
+                "`RMSNorm(shape)(x)` must satisfy the invariant `shape == x.shape`"
+                f"Received `shape={self.shape} and `x.shape={x.shape}`. You might need "
+                "to replace `rms_norm(x)` with `jax.vmap(rms_norm)(x)`.\n"
+                "\n"
+                "If this is a new error for you, it might be because this became "
```
patrick-kidger, line 376: This part of the message can be removed.

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,125 @@ def __call__(
...
+                "stricter in Equinox v0.11.0. Previously all that was required is that "
+                "`x.shape` ended with `shape`. However, this turned out to be a "
+                "frequent source of bugs, so we made the check stricter!"
+            )
+        inv_rms = jax.lax.rsqrt(jnp.sum(x**2) + self.eps)
```
patrick-kidger, line 381: +1 for using `rsqrt`, it often gets forgotten!

packquickly:
Ah! Good catch on the numerics! If this looks right I'll rebase into 1 commit to merge

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+class RMSNorm(Module):
+    r"""
+    A simplified version of LayerNorm which rescales the inputs, but does not center
+    them. Optionally applies a learned reweighting of the transformed array afterward.
+
+    Given an input array $x$, this layer computes
+
+    $$\frac{x}{\sqrt{\varepsilon + \frac{1}{n}\Vert x \Vert^2_2}} \gamma + \beta$$
+
+    where $\Vert \cdot \Vert_2$ is the 2-norm, $n = \dim(x)$, and $\gamma$ is a
```
patrick-kidger, line 282: I'd probably write it out explicitly, to avoid ambiguity whether "2-norm" is mean or sum.

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+    learned array with the same shape as $x$ if `use_weight=True`, or
+    $\gamma = 1/\sqrt{n}$ if `use_weight=False`, as proposed in
```
patrick-kidger, line 284: Not true as per discussion?

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+        eps: float = 1e-5,
+        use_weight: bool = False,
+        use_bias: bool = False,
+        **kwargs,
+    ):
+        """**Arguments:**
+
+        - `shape`: Shape of the input.
+        - `eps`: Value added to denominator for numerical stability.
+        - `use_weight`: Whether the module has learnable affine weights.
```
patrick-kidger, line 321: `use_bias` is missing?

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+        @article{zhang2019root,
+            title={Root Mean Square Layer Normalization},
+            author={Biao Zhang and Rico Sennrich},
+            year={2019},
+            journal={arXiv:1910.07467}
+    }
+        ```
+    """
+
+    shape: tuple[int] = field(static=True)
```
patrick-kidger, line 301: this annotation should be `tuple[int, ...]`

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+        returned.
+        """
+        if x.shape != self.shape:
+            raise ValueError(
+                "`RMSNorm(shape)(x)` must satisfy the invariant `shape == x.shape`"
+                f"Received `shape={self.shape} and `x.shape={x.shape}`. You might need "
+                "to replace `rms_norm(x)` with `jax.vmap(rms_norm)(x)`.\n"
+            )
+        inv_rms = jax.lax.rsqrt(jnp.sum(x**2) + self.eps)
+        out = jnp.sqrt(self.dim) * inv_rms * x
```
patrick-kidger, line 377: Note that `inv_rms` is badly named here, as it's actually an inv-sum-of-squares right now.

Can you just replace `jnp.sum` with `jnp.mean`?

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+        - `eps`: Value added to denominator for numerical stability.
+        - `use_weight`: Whether the module has learnable affine weights.
+        """
+        super().__init__(**kwargs)
+        if isinstance(shape, int):
+            shape = (shape,)
+        else:
+            shape = tuple(shape)
+        self.shape = shape
+        self.dim = prod(self.shape)
```
patrick-kidger, line 329: As this is easily derivable from existing attributes (and after my suggested change in `__call__`, unused) then I think it can be removed.

equinox/nn/_normalisation.py
```diff
@@ -267,3 +268,118 @@ def __call__(
...
+        self.eps = eps
+        self.use_weight = use_weight
+        self.use_bias = use_bias
+        self.weight = jnp.ones(shape) if use_weight else None
+        self.bias = jnp.ones(shape) if use_bias else None
```
jondeaton, line 334: `bias` should be initialized with `jnp.zeros` so that at initialization the elementwise affine is a no-op.

patrick-kidger: Oh, hey Jon!

It looks like this PR isn't going anywhere at the moment -- if you or anyone else wants to pick it up then I'd still be happy to see this merged.

jondeaton: Hey Patrick :)

I'd be happy to pick this up. Would you/Jason prefer that I make a new PR against `main`, or should I create a PR against `packquickly:rmsnorm`?

patrick-kidger: No strong feelings! As you've done it is fine.

patrick-kidger:
Closing in favour of #629.

PR #547, opened 2023-10-09
create `py.typed` marker file to mark the code as typed.

vidhanio:
without this file, mypy complains:
```
src/***/__main__.py:4: error: Skipping analyzing "equinox": module is installed, but missing library stubs or py.typed marker  [import]
src/***/__main__.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```
adding this file fixes the error.


patrick-kidger:
Whoops - LGTM! Thank you for the contribution.

By the way, I'd recommend trying out [pyright](https://github.com/microsoft/pyright), which is usually faster / more accurate than mypy.

vidhanio:
pylance/pyright also had a similar error, i just assumed mypy errors are more general. i saw that pyright is used in ci in this project, but local development doesn't require a `py.typed`, unlike when you import it as an external package.

thank you for merging! this library looks great. :)

PR #556, opened 2023-10-13
Fix missing index argument in docstring example for `StateIndex`

edwardwli:


patrick-kidger:
Thank you for the fix! LGTM, so merged. :)

PR #557, opened 2023-10-14
filter_value_and_grad: preserve return type of wrapped function

ConnorBaker:
I ran into this while trying to update the RNN tutorial to use `jaxtyping`.

Using a bounded type variable conveys to the type checker that wrapping a function with `filter_value_and_grad` will not change the type of value returned by the original function.

As an example:

Suppose `f` returns a `float`. Let `wrapped = filter_value_and_grad(f)`. Then the return type of `wrapped` is `tuple[_Scalar, PyTree]`, not `tuple[float, PyTree]`, where `_Scalar` is defined here: https://github.com/patrick-kidger/equinox/blob/56bafcb17bbced52974cf8fe0ee84d20cf778e38/equinox/_ad.py#L102

The wrapped function can return a value of a different type so long as it is also a `_Scalar` (like `complex`).

Using the type variable communicates this invariant to the type checker.

I opted not to replace `_Scalar` with `_ScalarTy` throughout the entire file because it doesn't make sense to use a type variable which occurs exclusively arguments or the return value, and there were a few other use sites which do that with `_Scalar`.

### DIFF
--- a/equinox/_ad.py
+++ b/equinox/_ad.py
@@ -100,19 +100,20 @@ class _GradWrapper(Module):
 
 
 _Scalar = Union[float, complex, Float[ArrayLike, ""], Complex[ArrayLike, ""]]
+_ScalarTy = TypeVar("_ScalarTy", bound=_Scalar)
 
 
 @overload
 def filter_value_and_grad(
     *, has_aux: Literal[False] = False
-) -> Callable[[Callable[_P, _Scalar]], Callable[_P, tuple[_Scalar, PyTree]]]:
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Hmm, looks like checks are failing just for formatting. (See [the contributing guide](https://github.com/patrick-kidger/equinox/blob/main/CONTRIBUTING.md) for how to have these checks run automatically locally, including automatically formatting.)

Other than that, this LGTM -- thank you for the fix.

ConnorBaker:
That’s odd, I had installed and run them locally to make sure. I wonder if I have some global config set somewhere…

patrick-kidger:
Alright, LGTM! Merged.

PR #568, opened 2023-10-22
RoPE Embeddings

### LINKED ISSUE #566: RoPE embeddings & Sinusoidal embeddings
Hey Patrick,

I was wondering if we could add RoPE and Sinusoidal embeddings to Equinox. It got a bit tedious to rewrite those every time, so I think it'd be nice to include it. 

I went ahead and provided 2 implementations: one with and one without the use of complex numbers.

Here's the complex numbers version:

```python
def precompute_freqs_cis(
    dim: int, end: int, theta: float = 10000.0
) -> Complex[Array, "end dim/2"]:
    def polar(abs, angle):
        return jnp.array(
            abs * jnp.cos(angle) + abs * jnp.sin(angle) * 1j, dtype=jnp.complex64
        )

    freqs = 1.0 / (theta ** (jnp.arange(0, dim, 2)[jnp.newaxis, :] / dim))
    t = jnp.arange(end)
    freqs_outer = jnp.outer(t, freqs)
    freqs_cis = polar(jnp.ones_like(freqs_outer), freqs_outer)
    return freqs_cis


def get_rope_embeddings_complex(x: Float[Array, "seq_len dim"]):
    def negate_half(x: Float[Array, "seq_len dim"]):
        d_2 = x.shape[-1] // 2
        return jnp.concatenate([x[..., :d_2], -x[..., d_2:]], axis=-1)

    seq_len, dim = x.shape
    freqs = precompute_freqs_cis(dim, seq_len)
    neg_half_x = negate_half(x)
    freqs_real = jnp.tile(freqs.real, (1, 2))
    freqs_imag = jnp.tile(freqs.imag, (1, 2))

    x_rope = (x * freqs_real) + (neg_half_x * freqs_imag)
    return x_rope
```

and this is the real numbers version (what you would fine in the RoFormer paper)
```python
def get_freqs(seq_len: int, dim: int, base: int = 10000):
    theta = 1 / (base ** (jnp.arange(0, dim, 2) / dim))
    t = jnp.arange(seq_len)

    idx_theta = jnp.einsum("i,j->ij", t, theta)
    idx_theta = jnp.concatenate([idx_theta, idx_theta], axis=1)

    freqs_cos = jnp.cos(idx_theta)
    freqs_sin = jnp.sin(idx_theta)

    return freqs_cos, freqs_sin


def get_rope_embedding(x: Float[Array, "seq_len dim"]):
    def negate_half(x: Float[Array, "seq_len dim"]):
        d_2 = x.shape[-1] // 2
        return jnp.concatenate([x[..., :d_2], -x[..., d_2:]], axis=-1)

    seq_len, dim = x.shape
    cos, sin = get_freqs(seq_len, dim)
    neg_half_x = negate_half(x)
    x_rope = (x * cos) + (neg_half_x * sin)
    return x_rope
```
Clearly, this is the not-cached version, i.e. you actually only need to compute the frequencies once and that keep those as a static field (I think as a static field).

I went ahead and performed a performance test on both:

`dim=1024` and `seq_len=4096` on MacBook Pro M1 Base model. Results are in seconds.

|          | Complex | Real      | 
|----------|---------|-----------|
| No JIT   | 299.1253| 193.1669  |
| Avg No Jit| 0.029912 | 0.01931 |
| With JIT | 71.4320 | 83.8433   |
| Avg With Jit| 0.0071 | 0.0083|

As you can see, the "real" version is *much* faster when not jitted, but the complex version is faster when you jit it.

My questions are:
1. Is there any interest in adding RoPE & Sinusoidal Embeddings?
2. If yes, which version should I add?

I think it'd be very helpful to have those as you frequently use those when implementing a transformer.

For reference, here's the entire code that you can copy paste and run (it also contains the sinusoidal embeddings too).

<details>
  <summary>Click me</summary>
  
  ```python
import sys
import time
from typing import Optional, Sequence, Union

import jax
import jax.numpy as jnp
import matplotlib.pyplot as plt
import numpy as np
from icecream import ic
from jaxtyping import Array, Complex, Float, Int, PRNGKeyArray
from matplotlib.animation import FuncAnimation
from matplotlib.patches import FancyArrowPatch

import equinox as eqx


# for reference, this is the ported version of the original code from LLaMA
# which I shamelessly stole, I mean borrowed from
def apply_rotary_emb(
    xq: Float[Array, "seq_len n_heads max_seq_len"],
    xk: Float[Array, "seq_len n_heads max_seq_len"],
    freqs_cis: Complex[Array, "seq_len dim//n_heads"],
) -> tuple[
    Float[Array, "seq_len n_heads max_seq_len"],
    Float[Array, "seq_len n_heads max_seq_len"],
]:
    def reshape_for_broadcast(freqs_cis: Array, x: Array):
        ndim = x.ndim
        assert 0 <= 1 < ndim
        assert freqs_cis.shape == (
            x.shape[1],
            x.shape[-1],
        ), f"{freqs_cis.shape=} != {(x.shape[1], x.shape[-1])}"
        shape = [d if i == 1 or i == ndim - 1 else 1 for i, d in enumerate(x.shape)]
        return freqs_cis.reshape(*shape)

    def reshape_complex_array(x: Array):
        return jax.lax.collapse(x, len(x.shape) - 2, len(x.shape))

    # e.g.: if xq.shape = 1, 32, 128, then xq_reshaped.shape =  1, 32, -1, 2 => 1, 32, 64, 2
    # it's 64 because total_elements = 1 * 32 * 128 = 4096 and 4096 / (1 * 32 * 2) = 64

    # also, because complex does not accept float16, we have to parse it to float32
    xq_reshaped = jnp.array(xq.reshape(*xq.shape[:-1], -1, 2), dtype=jnp.float32)
    xq_ = jax.lax.complex(
        xq_reshaped[..., 0], xq_reshaped[..., 1]
    )  # the last dimension is of size 2, the "..." skip over the first dimensions until we reach the last one
    xk_reshaped = jnp.array(xk.reshape(*xk.shape[:-1], -1, 2), dtype=jnp.float32)
    xk_ = jax.lax.complex(xk_reshaped[..., 0], xk_reshaped[..., 1])

    freqs_cis = reshape_for_broadcast(freqs_cis, xq_)
    tq = xq_ * freqs_cis
    tk = xk_ * freqs_cis
    xk_out = reshape_complex_array(jnp.stack((jnp.real(tk), jnp.imag(tk)), axis=-1))
    xq_out = reshape_complex_array(jnp.stack((jnp.real(tq), jnp.imag(tq)), axis=-1))
    return xq_out, xk_out


# standard positional encoding
def get_positional_encoding(
    n_tokens: int, n_vocab: int
) -> Float[Array, "n_tokens n_vocab"]:
    pos = jnp.arange(n_tokens)[:, jnp.newaxis]
    div_term = jnp.exp(jnp.arange(0, n_vocab, 2) * -(jnp.log(10000.0) / n_vocab))
    # the following expression is closer to the actual notation they used.
    # div_term = 1 / 10000 ** (jnp.arange(0, n_vocab, 2) / n_vocab)
    pos_enc = jnp.zeros((n_tokens, n_vocab))
    pos_enc = pos_enc.at[:, 0::2].set(jnp.sin(pos * div_term))
    pos_enc = pos_enc.at[:, 1::2].set(jnp.cos(pos * div_term))
    return pos_enc


def sinusoidal_animation():
    def update(num, x, pos_enc, arrow, label):
        updated = x + pos_enc[num]
        arrow.set_positions((0, 0), (updated[1], updated[0]))
        label.set_position((updated[1] + 0.1, updated[0] + 0.1))
        label.set_text(str(num))
        return arrow, label

    pos_enc = get_positional_encoding(8, 2)
    x = jnp.array([1, 1])

    xmin, xmax, ymin, ymax = -3, 3, -3, 3
    ticks_frequency = 1

    fig, ax = plt.subplots()
    ax.set_xlim(xmin, xmax)
    ax.set_ylim(ymin, ymax)
    ax.set(xlim=(xmin - 1, xmax + 1), ylim=(ymin - 1, ymax + 1), aspect="equal")

    ax.spines["bottom"].set_position("zero")
    ax.spines["left"].set_position("zero")

    # Remove top and right spines
    ax.spines["top"].set_visible(False)
    ax.spines["right"].set_visible(False)

    # Create 'x' and 'y' labels placed at the end of the axes
    ax.set_xlabel("x", size=14, labelpad=-24, x=1.03)
    ax.set_ylabel("y", size=14, labelpad=-21, y=1.02, rotation=0)

    # Create custom major ticks to determine position of tick labels
    x_ticks = np.arange(xmin, xmax + 1, ticks_frequency)
    y_ticks = np.arange(ymin, ymax + 1, ticks_frequency)
    ax.set_xticks(x_ticks[x_ticks != 0])
    ax.set_yticks(y_ticks[y_ticks != 0])

    # Create minor ticks placed at each integer to enable drawing of minor grid
    # lines: note that this has no effect in this example with ticks_frequency=1
    ax.set_xticks(np.arange(xmin, xmax + 1), minor=True)
    ax.set_yticks(np.arange(ymin, ymax + 1), minor=True)

    # Draw major and minor grid lines
    ax.grid(which="both", color="grey", linewidth=1, linestyle="-", alpha=0.2)

    # Draw arrows
    arrow_fmt = dict(markersize=4, color="black", clip_on=False)
    ax.plot((1), (0), marker=">", transform=ax.get_yaxis_transform(), **arrow_fmt)
    ax.plot((0), (1), marker="^", transform=ax.get_xaxis_transform(), **arrow_fmt)

    # (line,) = ax.plot([], [], "r-")

    arrow = FancyArrowPatch((0, 0), (x[1], x[0]), mutation_scale=5, color="r")
    ax.add_patch(arrow)
    label = ax.text(x[1] + 0.1, x[0] + 0.1, "0")
    ani = FuncAnimation(
        fig,
        update,
        len(pos_enc),
        fargs=[x, pos_enc, arrow, label],
        interval=500,
        blit=True,
    )
    plt.show()


def get_freqs(seq_len: int, dim: int, base: int = 10000):
    theta = 1 / (base ** (jnp.arange(0, dim, 2) / dim))
    t = jnp.arange(seq_len)

    idx_theta = jnp.einsum("i,j->ij", t, theta)
    idx_theta = jnp.concatenate([idx_theta, idx_theta], axis=1)

    freqs_cos = jnp.cos(idx_theta)
    freqs_sin = jnp.sin(idx_theta)

    return freqs_cos, freqs_sin


def get_rope_embedding(x: Float[Array, "seq_len dim"]):
    def negate_half(x: Float[Array, "seq_len dim"]):
        d_2 = x.shape[-1] // 2
        return jnp.concatenate([x[..., :d_2], -x[..., d_2:]], axis=-1)

    seq_len, dim = x.shape
    cos, sin = get_freqs(seq_len, dim)
    neg_half_x = negate_half(x)
    x_rope = (x * cos) + (neg_half_x * sin)
    return x_rope


def precompute_freqs_cis(
    dim: int, end: int, theta: float = 10000.0
) -> Complex[Array, "end dim/2"]:
    def polar(abs, angle):
        return jnp.array(
            abs * jnp.cos(angle) + abs * jnp.sin(angle) * 1j, dtype=jnp.complex64
        )

    freqs = 1.0 / (theta ** (jnp.arange(0, dim, 2)[jnp.newaxis, :] / dim))
    t = jnp.arange(end)
    freqs_outer = jnp.outer(t, freqs)
    freqs_cis = polar(jnp.ones_like(freqs_outer), freqs_outer)
    return freqs_cis


def get_rope_embeddings_complex(x: Float[Array, "seq_len dim"]):
    def negate_half(x: Float[Array, "seq_len dim"]):
        d_2 = x.shape[-1] // 2
        return jnp.concatenate([x[..., :d_2], -x[..., d_2:]], axis=-1)

    seq_len, dim = x.shape
    freqs = precompute_freqs_cis(dim, seq_len)
    neg_half_x = negate_half(x)
    freqs_real = jnp.tile(freqs.real, (1, 2))
    freqs_imag = jnp.tile(freqs.imag, (1, 2))
    # freqs_real = freqs.real[..., jnp.newaxis]
    # freqs_imag = freqs.imag[..., jnp.newaxis]

    x_rope = (x * freqs_real) + (neg_half_x * freqs_imag)
    return x_rope


def rope_animation():
    def update(num, x, pos_enc, arrow, label):
        updated = pos_enc[num]
        arrow.set_positions((0, 0), (updated[1], updated[0]))
        label.set_position((updated[1] + 0.1, updated[0] + 0.1))
        label.set_text(str(num))
        return arrow, label

    x = jnp.ones((8, 2))
    pos_enc = get_rope_embedding(x)

    xmin, xmax, ymin, ymax = -3, 3, -3, 3
    ticks_frequency = 1

    fig, ax = plt.subplots()
    ax.set_xlim(xmin, xmax)
    ax.set_ylim(ymin, ymax)
    ax.set(xlim=(xmin - 1, xmax + 1), ylim=(ymin - 1, ymax + 1), aspect="equal")

    ax.spines["bottom"].set_position("zero")
    ax.spines["left"].set_position("zero")

    # Remove top and right spines
    ax.spines["top"].set_visible(False)
    ax.spines["right"].set_visible(False)

    # Create 'x' and 'y' labels placed at the end of the axes
    ax.set_xlabel("x", size=14, labelpad=-24, x=1.03)
    ax.set_ylabel("y", size=14, labelpad=-21, y=1.02, rotation=0)

    # Create custom major ticks to determine position of tick labels
    x_ticks = np.arange(xmin, xmax + 1, ticks_frequency)
    y_ticks = np.arange(ymin, ymax + 1, ticks_frequency)
    ax.set_xticks(x_ticks[x_ticks != 0])
    ax.set_yticks(y_ticks[y_ticks != 0])

    # Create minor ticks placed at each integer to enable drawing of minor grid
    # lines: note that this has no effect in this example with ticks_frequency=1
    ax.set_xticks(np.arange(xmin, xmax + 1), minor=True)
    ax.set_yticks(np.arange(ymin, ymax + 1), minor=True)

    # Draw major and minor grid lines
    ax.grid(which="both", color="grey", linewidth=1, linestyle="-", alpha=0.2)

    # Draw arrows
    arrow_fmt = dict(markersize=4, color="black", clip_on=False)
    ax.plot((1), (0), marker=">", transform=ax.get_yaxis_transform(), **arrow_fmt)
    ax.plot((0), (1), marker="^", transform=ax.get_xaxis_transform(), **arrow_fmt)

    # (line,) = ax.plot([], [], "r-")

    arrow = FancyArrowPatch((0, 0), (0, 0), mutation_scale=5, color="r")
    ax.add_patch(arrow)
    label = ax.text(x[1] + 0.1, x[0] + 0.1, "0")
    ani = FuncAnimation(
        fig,
        update,
        len(pos_enc),
        fargs=[x, pos_enc, arrow, label],
        interval=1000,
        blit=True,
    )
    plt.show()


if __name__ == "__main__":
    dim = 1024
    max_seq_len = 4096

    x_rope_real = get_rope_embedding(jnp.ones((max_seq_len, dim)))
    x_rope_complex = get_rope_embeddings_complex(jnp.ones((max_seq_len, dim)))

    ic(x_rope_real)
    ic(x_rope_complex)

    ic(jnp.allclose(x_rope_real, x_rope_complex))
    # testing the rope embedding
    test_cases = 1_000_0

    x = jax.random.uniform(
        jax.random.PRNGKey(42), (max_seq_len, dim), dtype=jnp.float32
    )
    start_time = time.time()
    for _ in range(test_cases):
        x_rope_real = get_rope_embedding(x)
    end_time = time.time()
    ic("Real:", end_time - start_time)
    ic("Avg:", (end_time - start_time) / test_cases)
    start_time = time.time()
    for _ in range(test_cases):
        x_rope_complex = get_rope_embeddings_complex(x)
    end_time = time.time()
    ic("Complex: ", end_time - start_time)
    ic("Avg:", (end_time - start_time) / test_cases)

    # rope_animation()
    # sinusoidal_animation()

  ```
</details>

Artur-Galstyan:
Added RoPE embeddings from [the RoFormer paper](https://arxiv.org/pdf/2104.09864.pdf). 

I need to add this to my transformer to perform some tests first before I can mark this as ready. Also if it's not needed/wanted, feel free to close this PR :) 

If merged, closes #566


Artur-Galstyan:
I went ahead and added RoPE to the current MHA implementation, but I'm not **super** happy with it.

I had to add the `state_length` parameter from the [autoregressive-attention branch](https://github.com/patrick-kidger/equinox/tree/autoregressive-attention), which was not yet introduced in the current MHA implementation. Also, currently, we're not enforcing a `state_length`, which means - if we didn't add `state_length` as an additional parameter - we'd have to recompute the frequencies whenever `state_length` changes. The question here would be: do you want to allow that? (I think we shouldn't because in Jax you should predefine all your matrix shapes once and not dynamically change it to avoid re-jitting)

What do you think?

patrick-kidger:
Okay, so I've just rediscovered a rotary emebdding implementation that I had lying around in an old project. This uses a real implementation, although it sounds like the complex implementation is what we really want. I've quickly adapted it (untested) to try and fit what we're doing here.

```python
# rotary implementation

class RotaryEmbedding(eqx.Module):
    max_seq_length: Optional[int] = eqx.field(static=True)
    qk_size: int = eqx.field(static=True)
    cos: Optional[Float[Array, "max_seq_length qk_size"]]
    sin: Optional[Float[Array, "max_seq_length qk_size"]]
    inv_freq: Optional[Float[Array, " half_qk_size"]]

    def __init__(self, max_seq_length: Optional[int], qk_size: int):
        if qk_size % 2 == 1:
            raise ValueError("Rotary embeddings require qk_size to be even.")
        self.max_seq_length = max_seq_length
        self.qk_size = qk_size
        inv_freq = 1.0 / (10000 ** (jnp.arange(0, qk_size, 2) / qk_size))
        if max_seq_length is None:
            self.inv_freq = inv_freq
            self.sin = self.cos = None
        else:
            self.inv_freq = None
            self.sin, self.cos = self._sincos(max_seq_length, inv_freq)

    def _sincos(self, max_seq_length: int, inv_freq: Float[Array, " half_qk_size"]):
        t = jnp.arange(max_seq_length, dtype=inv_freq.dtype)
        freqs = t[:, None] * inv_freq[None, :]
        emb = jnp.concatenate([freqs, freqs], axis=-1)
        assert emb.shape == (max_seq_length, self.qk_size)
        return jnp.sin(emb), jnp.cos(emb)

    def call(self, qk: Float[Array, "seq_length qk_size"]):
        seq_length, qk_size = qk.shape
        if self.max_seq_length is not None:
            assert seq_length <= self.max_seq_length
        assert qk_size == self.qk_size
        qk1 = qk[:, : qk_size // 2]
        qk2 = qk[:, qk_size // 2 :]
        rotate_qk = jnp.concatenate([-qk2, qk1], axis=-1)
        if self.max_seq_length is None:
            assert self.inv_freq is not None
            sin, cos = self._sincos(seq_length, lax.stop_gradient(self.inv_freq))
            assert self.sin is None
            assert self.cos is None
        else:
            assert self.inv_freq is None
            sin, cos = lax.stop_gradient(self.sin), lax.stop_gradient(self.cos)
            assert sin is not None
            assert cos is not None
        return qk * cos[:seq_length] + rotate_qk * sin[:seq_length]

    def __call__(
        self,
        query_heads: Float[Array, "seq_length num_heads qk_size"],
        key_heads: Float[Array, "seq_length num_heads qk_size"],
        value_heads: Float[Array, "seq_length num_heads vo_size"],
    ):
        query_heads = jax.vmap(self.call, in_axes=1, out_axes=1)(query_heads)
        key_heads = jax.vmap(self.call, in_axes=1, out_axes=1)(key_heads)
        return key_heads, query_heads, value_heads

# how this integrates into the existing attention code:

class MultiheadAttention(Module):
    ...
    process_heads: Callable[[Array, Array, Array], tuple[Array, Array, Array]]

    def __init__(..., process_heads=lambda x, y, z: (x, y, z):
        ...
        self.process_heads = process_heads

    def __call__(...):
        ...
        key_heads, query_heads, value_heads = self.process_heads(key_heads, query_heads, value_heads)
        ...

# user code

MultiheadAttention(process_heads=RotaryEmbedding(...))
```

I think there are a couple of things the above sketch does that we can use here:
- it uses of `lax.stop_gradient` so that we don't accidentally train the embeddings!
- it doesn't bake-in rotary embeddings into the multihead attention implementation, and instead does something pretty general. This might be useful if someone invents something even better in a couple of years.
- it means we don't need to worry about explicitly specifying the state length in MHA itself just yet.
- it supports both caching the embeddings (pass `RotaryEmbedding(max_seq_length=5)`) and computing them on-the-fly (pass `RotaryEmbedding(max_seq_length=None)`).

By the way, that second bullet point is a general trick, known as dependency inversion / dependency injection. Those terms are most frequently mentioned in object-oriented programming, but it's actually a useful concept more broadly than that -- in this case for our not-OOP functional JAX+Equinox stuff.

One thing it *doesn't* do, and which we should have a think about it, is reusing the same sines and cosines across multiple `RotaryEmbedding` instances. It'd be much better if we can avoid creating mutiple copies of the sines and cosines. (Maybe a global cache keyed by `max_seq_length` and `qk_size`?)

Artur-Galstyan:
Alright, just a few questions and remarks:

1) I'll perform a full performance comparison between your (real) implementation, my (real) implementation and the (complex) implementation on an RTX 3090 (all I have 😭) once the implementation details are all ironed out (my previous benchmark was on a Macbook)

2) Essentially, the motto is: if `max_seq_len` is given, use the precomputed freqs, else compute them on the fly and then use those. I got that part implemented. 

3) Caching: The goal is that multiple RoPE embeddings have access to a global dictionary with the key `(seq_len, qk_size)`. I agree, this would be best. While browsing the repository to see if something like a global cache already exists, I found the file _caches.py. It has 2 lists in it, which are already used in _compile_utils.py (e.g.). My idea is - to have a nice separation of concerns - to add another internal cache in there, like `internal_rope_embedding_caches = dict()` and store the cache in there. Is that the right place for it? Then - always before computing the freqs - check if already in cache and then proceed.

4) The dependency injection idea had also crossed my mind (although I didn't know the term for it 😄) and I had thought against it because I thought it might confuse the user. But you're right. It's better to have the generality in place. Perhaps a convenience wrapper could be helpful, which passes the RoPE embedding as the `Callable` (e.g. `use_rope_embeddings = True` => callable = RoPEEmbeddings(...))

5) I'm a bit hesitant to perform the `vmap` within the RoPE embeddings directly (instead of letting the user handle that when the user passes the `callable` to the MHA implementation). The reason for that is that we force the user to ensure that the arguments have the shape `(seq_length num_heads qk_size)`. But the user might use those embeddings in a completely different scenario (although I can't think of one at the moment) which is not transformer-related. If kept as is, if the users array is just the shape `(x, y)` we'd force him to add another dimension in the middle `(x, 1, y)`. What do you think?

patrick-kidger:
(By the way -- I haven't forgotten about this PR. Have been a bit snowed under recently; will try to get back to this in a couple of weeks.)

Artur-Galstyan:
Hey Patrick, no worries. I'm in the middle of merging/updating the MHA refactoring anyway. Just need a couple more tests regarding the autoregressive features. Should be done soon.

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
...
+                self.freqs_cis = internal_rope_embedding_cache[
+                    (self.embedding_size, self.max_seq_len)
+                ]
+            else:
+                self.freqs_cis = self.precompute_freqs_cis(
+                    embedding_size, self.max_seq_len
+                )
+                internal_rope_embedding_cache[
+                    (self.embedding_size, self.max_seq_len)
+                ] = self.freqs_cis
```
patrick-kidger, line 139: You know, I think we can probably skip having this as an attribute altogether. Just create the global variable here if required, and then look it up during the forward pass.

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
...
+        return jnp.concatenate([x[..., :d_2], -x[..., d_2:]], axis=-1)
+
+    @staticmethod
+    def precompute_freqs_cis(
+        embedding_size: int, end: int, theta: float = 10000.0
+    ) -> Complex[Array, "end embedding_size/2"]:
+        def polar(abs, angle):
+            return jnp.array(
+                abs * jnp.cos(angle) + abs * jnp.sin(angle) * 1j, dtype=jnp.complex64
+            )
```
patrick-kidger, line 155: I think this function is only used once, so it can be inlined. Moreover I'm guessing the `jnp.array` isn't necessary: probably just `freqs_cis = jnp.cos(freqs_outer) + 1j * jnp.sin(freqs_outer)` will work?

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
...
+            (Keyword only argument.)
+
+        **Returns:**
+
+        A JAX array of shape `(seq_len, embedding_size)`, with the rotary positional
+        encoding applied to the input.
+        """
+
+        seq_len, embedding_size = x.shape
+        assert embedding_size == self.embedding_size, (
```
patrick-kidger, line 185: These should be `ValueError`s rather than `assert` statements. The former is used when a user has gotten something wrong; the latter is used when the developer has gotten something wrong (and an internal invariant no longer holds).

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
...
+
+        if self.max_seq_len is not None:
+            assert seq_len <= self.max_seq_len, (
+                f"x.shape[0] must be <= self.max_seq_len, "
+                f"but {x.shape[0]} > {self.max_seq_len}"
+            )
+
+        neg_half_x = self.negate_half(x)
+
+        if self.max_seq_len is not None:
```
patrick-kidger, line 201: Nittiest of nits: better to do `if self.max_seq_len is None` and flip the branches.
Doesn't really matter much here of course, but I find that this improves readability when applied to more complicated statements, so I like to emphasise it.

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
...
+        assert freqs_cis is not None
+        freqs_cis = jax.lax.stop_gradient(freqs_cis)
+        freqs_real = jnp.tile(freqs_cis.real, (1, 2))
+        freqs_imag = jnp.tile(freqs_cis.imag, (1, 2))
+
+        x_rope = (x * freqs_real) + (neg_half_x * freqs_imag)
+        return x_rope
+
+
+class SinusoidalPositionalEmbedding(Module):
```
patrick-kidger, line 219: (Basically the same comments here.)

patrick-kidger (commented):
Gosh, first of all, *thank you* for being so patient with me. I can't believe I kept this languishing on my to-do list for so many months.

I've finally gone over and done a review. Overall I really like it! I've left some nitty comments but I'd love to get this into the next release of Equinox.

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
                 "`eqx.nn.Embedding()(x)` should be called with a scalar index `x`. "
                 "Use `jax.vmap` if you would like to index with multiple values."
             )
+
+
+class RotaryPositionalEmbedding(Module):
+    embedding_size: int = field(static=True)
+    max_seq_len: Optional[int] = field(static=True)
+    freqs_cis: Optional[Complex[Array, "embedding_size/2"]] = field(static=True)
```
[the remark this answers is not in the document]
Artur-Galstyan: > Also, it shouldn't be static if it will be an array.

I always thought that everything not marked as `static=True` will be affected by the `eqx.filter_grad(...)` function and thus be changed when calling `optax.apply_updates`. In this case, we don't want those to change, right?

equinox/nn/_embedding.py
```diff
@@ -89,3 +93,253 @@ def __call__(
...
+        self,
+        embedding_size: int,
+        max_seq_len: Optional[int] = None,
+        *,
+        key: Optional[PRNGKeyArray] = None,
+        **kwargs,
+    ):
+        """**Arguments:**
+
+        `RotaryPositionalEmbedding` requires:
```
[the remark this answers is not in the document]
Artur-Galstyan: By "docs on the attention page", are you referring to the file docs/api/nn/attention.md? 

If so, sure, I'll add it. But might it not be a bit confusing, since the classes are implemented under `_embedding.py` yet their docs are under `attention.md`?

equinox/nn/_embedding.py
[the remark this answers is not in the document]
Artur-Galstyan: > I think to make the annotation here work, it should be `"{self.embedding_size}/2"`.

The ruff linter gave me trouble since it couldn't find `self` there. I tried this `RotaryPositionalEmbedding.embedding_size / 2` and it didn't complain any more. WDYT?

equinox/nn/_embedding.py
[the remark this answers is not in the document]
patrick-kidger: So the existing embedding docs are for a different kind of embedding, namely lookup tables. We overload the word "embedding" a lot in ML!
So I'm thinking of keeping these together with attention, since that's where they used.

equinox/nn/_embedding.py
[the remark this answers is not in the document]
Artur-Galstyan: > > Also, it shouldn't be static if it will be an array.
> 
> I always thought that everything not marked as `static=True` will be affected by the `eqx.filter_grad(...)` function and thus be changed when calling `optax.apply_updates`. In this case, we don't want those to change, right?

It's been so long that I even forgot my own implementation, lol. I had this in place just for that case:

`freqs_cis = jax.lax.stop_gradient(freqs_cis)`

But for my own understanding: any array marked with `static=True` would - during `jax.grad()` behave the same as if I do `jax.lax.stop_gradient(...)`?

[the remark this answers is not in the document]
patrick-kidger: On `static=True`: it's true that it won't be seen by gradients etc. However, you can basically never use arrays as static fields, due to this:
```python
import equinox as eqx
import jax
import jax.numpy as jnp

class Foo(eqx.Module):
    x: jnp.ndarray = eqx.field(static=True)

foo = Foo(jnp.arange(3))
bar = Foo(jnp.arange(3) + 1)

f = jax.jit(lambda x: x.x + 1)
f(foo)
f(bar)
```
what's going on here is that in the `f(bar)` call, JAX will try to compare static fields for equality, to see whether or not it needs to recompile. And `foo.x = bar.x` returns a nonscalar array, not a bool, so it crashes.

In any case the above should all just be an academic question, since as per below I'm recommending not making this a field at all.

On ruff: prepend a space, i.e. `" {self.embedding_size}/2"`. I think that should make it work.

equinox/nn/_embedding.py
Artur-Galstyan: I'm having some problems understanding and implementing this.

I realise that it makes sense to **not** have `freqs_cis` as an attribute anymore. Instead, it's better (and uses less code - always a win) to simply check the cache during the forward pass. 

Only problem now is that we get leaked tracers!

Essentially, its like this now

```python

cache = {}

class Foo(eqx.Module):
    n: int = field(static=True)

    def __init__(self, n):
        self.n = n

    def __call__(self, x: Float[Array, "n"]):
        n, _ = x.shape

        y = None
        if n in cache:
            y = cache[n]
        else:
            y = jnp.arange(n)
            cache[n] = y #<<<<<<<<< Leaking the tracer here!
        
        return x + y
```

When JITting `Foo`, we get a leak. So, I'm not sure how to get it to work here 😞

Artur-Galstyan: On second thought: do we even _really_ need to cache the `freqs_cis`? If the user calls the function with constantly changing `seq_len`s then won't Jax reJIT the function anyway?

I.e. 

```
import equinox as eqx
import jax.numpy as jnp


class Test(eqx.Module):
    def __init__(self):
        pass

    def __call__(self, x):
        print(f"JIT: {x.shape=}")
        return x + 1


t = Test()
t = eqx.filter_jit(t)


print(t(jnp.array([1])))
print(t(jnp.array([1])))

print(t(jnp.array([1, 2])))
print(t(jnp.array([1, 2, 3])))
print(t(jnp.array([1, 2])))

```

Which outputs

```
JIT: x.shape=(1,)
[2]
[2]
JIT: x.shape=(2,)
[2 3]
JIT: x.shape=(3,)
[2 3 4]
[2 3]
```

Which would suggest that Jax automatically does the caching for us, no?

patrick-kidger: On the tracer leak: ah, try wrapping the setter in `with jax.ensure_compile_time_eval()`.

As for why we should cache this: I *believe* (but have not carefully checked) that we need to do this to work around a quirk of the XLA compiler. Namely, suppose we were to compute the (co)sines as we needed them, inside `__call__`. Now suppose we have many `RotaryPositionalEmbedding` layers, and iterated through them in a `jax.lax.scan` using the [scan-over-layers trick](https://docs.kidger.site/equinox/tricks/#improve-compilation-speed-with-scan-over-layers). Then I belive XLA is not smart enough to identify that the repeated computation can be hoisted out of the scan (i.e. to perform it only once and reuse the result every time). Instead, it will recompute them every time. Thus, it falls to us to do this manually instead.

In particular this is to do with having multiple `RotaryPositionalEmbedding` layers in the same model; it's unrelated to handling different calls with different sequence lengths.

dlwh: @Artur-Galstyan If you want something to crib off of, we have an implementation here https://github.com/stanford-crfm/levanter/blob/main/src/levanter/models/llama.py#L223 I think the only thing missing is the ensure_compile_time_eval bit.

Artur-Galstyan:
Hey Patrick, could you re-review the current changes? I added an example to the sinusoidal class as well as removed the `max_seq_len` entirely in favour of run time checking. I got this idea from David's code in which there is also no `max_seq_len` involved.

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+
+class RotaryPositionalEmbedding(Module, strict=True):
+    """
+    A rotary positional encoding module, as described in the paper
+    "RoFormer: Enhanced Transformer with Rotary Position Embedding". While
+    this module can be used in any context, it is particularly useful for
+    providing positional information to transformer models.
+
+    !!! example
+        The following example demonstrates how to use `RotaryPositionalEmbedding` in
```
patrick-kidger, line 109: Nit: this needs a newline to render correctly in the docs:
```
!!! example

    The following ...
```
(Try running `pip install -r docs/requirements.txt`, and then `mkdocs serve -> Control-C -> mkdocs serve`, to see what the result looks like. Note that you do need two consecutive `mkdocs serve` commands, as the first one compuptes and caches some information needed for the second one.)

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+        a simple transformer model.
+        ```python
+
+        class TransformerBlock(eqx.nn.StatefulLayer):
```
patrick-kidger, line 113: This doesn't need to be a stateful layer.
(That's used just as a marker for `eqx.nn.Sequential` to call the layer as `layer(x, state)` rather than `layer(x)`.)

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+            query_rope_embeddings: RotaryPositionalEmbedding
+
+            def __init__(...):
+                ...
+                self.query_rope_embeddings = RotaryPositionalEmbedding(
+                    embedding_size=n_embd
+                )
+                self.key_rope_embeddings = RotaryPositionalEmbedding(
+                    embedding_size=n_embd
+                )
```
patrick-kidger, line 125: I think we should be able to use just a single `RotaryPositionalEmbedding` layer?

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+
+                    return query_heads, key_heads, value_heads
+
+                mha_output = self.mha_attention(
+                    process_heads=process_heads,
+                    query=jax.vmap(self.rms_norm)(x),
+                    key_=jax.vmap(self.rms_norm)(x),
+                    value=jax.vmap(self.rms_norm)(x),
+                    mask=mask,
+                )
```
patrick-kidger, line 145: This is excellent. I think this is really important to make it clear to a use how can be handled, thank you!

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+                    embedding_size=n_embd
+                )
+                self.key_rope_embeddings = RotaryPositionalEmbedding(
+                    embedding_size=n_embd
+                )
+                ...
+
+            def __call__(...):
+                def process_heads(query_heads, key_heads, value_heads):
+                    query_heads = jax.vmap(self.query_rope_embeddings,
```
patrick-kidger, line 130: Can you add a comment about array shapes, so that what the vmap is doing here can be understood?

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+                                           out_axes=1)(query_heads)
+                    key_heads = jax.vmap(self.key_rope_embeddings,
+                                         in_axes=1,
+                                         out_axes=1)(key_heads)
+
+                    return query_heads, key_heads, value_heads
+
+                mha_output = self.mha_attention(
+                    process_heads=process_heads,
+                    query=jax.vmap(self.rms_norm)(x),
```
patrick-kidger, line 141: I'd probably remove the explicit `query/key/value/mask=...` inputs here. They don't directly affect the purpose of this example so they're just visual noise.
(They're also suggestive of doing a few particular things, like self-attention and using RMSNorm, which may not be needed in general.)

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+        if embedding_size % 2 != 0:
+            raise ValueError(
+                f"x.shape[-1] must be even, but {x.shape[-1]} is not even."
+            )
+
+        neg_half_x = self.negate_half(x)
+
+        with jax.ensure_compile_time_eval():
+            if (embedding_size, seq_len) in internal_rope_embedding_cache:
+                freqs_cis = internal_rope_embedding_cache[(embedding_size, seq_len)]
```
patrick-kidger, line 236: I think we can make this even less memory-intensive: as `seq_len` increases then the cached value becomes a superset of the previous one. As such I think we should be able to keep track of precisely one array for each `embedding_size`, which represents the largest value of `seq_len` that is has seen yet. Smaller values of `seq_len` can be found by slicing this array at `[:seq_len]`.
WDYT?

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+            else:
+                freqs_cis = self.precompute_freqs_cis(embedding_size, seq_len)
+                internal_rope_embedding_cache[(embedding_size, seq_len)] = freqs_cis
+
+        assert freqs_cis is not None, "freqs_cis must not be None."
+        freqs_real = jnp.tile(freqs_cis.real, (1, 2))
+        freqs_imag = jnp.tile(freqs_cis.imag, (1, 2))
+
+        x_rope = (x * freqs_real) + (neg_half_x * freqs_imag)
+        return jax.lax.stop_gradient(x_rope)
```
patrick-kidger, line 246: I don't think we want to stop the gradient flowing through our KVQ heads! I'm guessing this is a holdover from when the frequencies were stored on the layer itself.

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+            def __call__(...):
+                def process_heads(query_heads, key_heads, value_heads):
+                    query_heads = jax.vmap(self.query_embeddings,
+                                           in_axes=1,
+                                           out_axes=1)(query_heads)
+                    key_heads = jax.vmap(self.key_embeddings,
+                                         in_axes=1,
+                                         out_axes=1)(key_heads)
+
+                    return query_heads, key_heads, value_heads
```
patrick-kidger, line 284: Possibly I'm revealing some naivety here, but I don't think I've seen sinusoidal embeddings used to modify embeddings very frequently. (Ever?) I usually think of it as something being provided at the very start of the architecture, not something provided on each layer.
@dlwh -- you probably have a better idea of this than I do?

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+    def __init__(
+        self,
+        embedding_size: int,
+        *,
+        theta: float = 10000.0,
+        key: Optional[PRNGKeyArray] = None,
+        **kwargs,
+    ):
+        """**Arguments:**
+        `SinusoidalPositionalEmbedding` requires:
```
patrick-kidger, line 325: Nit: can we remove the `` `SinusoidalPositionalEmbedding` requires:`` line from the docstring?

equinox/nn/_embedding.py
```diff
@@ -89,3 +96,313 @@ def __call__(
...
+    ) -> Float[Array, "seq_len embedding_size"]:
+        pos = jnp.arange(seq_len)[:, jnp.newaxis]
+        div_term = jnp.exp(
+            jnp.arange(0, embedding_size, 2) * -(jnp.log(theta) / embedding_size)
+        )
+        # the following expression is closer to the actual notation they used.
+        # div_term = 1 / 10000 ** (jnp.arange(0, embedding_size, 2) / embedding_size)
+        pos_enc = jnp.zeros((seq_len, embedding_size))
+        pos_enc = pos_enc.at[:, 0::2].set(jnp.sin(pos * div_term))
+        pos_enc = pos_enc.at[:, 1::2].set(jnp.cos(pos * div_term))
```
patrick-kidger, line 359: I have a pretty good hunch that JAX is going to do a terrible job of interleaving these, and will probably make copies here.

It doesn't *really* matter since we're running this under `jax.ensure_compile_time_eval`, but if I had to guess, then probably
```python
mult = pos * div_term
sines = jnp.sin(mult)
cosines = jnp.cos(mult)
jnp.stack([sines, cosines], axis=2).reshape((seq_len, embedding_size))
```
will be much faster.
(Assuming what I've written about is functionally correct anyway, please double-check...)

By the way, I'd have thought the `pos * div_term` broadcasting should fail when running under `JAX_NUMPY_RANK_PROMOTION=raise`, as [we do](https://github.com/patrick-kidger/equinox/blob/34252989c251c315c6eee5895dd64e9645b350ce/tests/conftest.py#L11). I'm surprised it doesn't seem to here - any ideas?

equinox/nn/_embedding.py
Artur-Galstyan: Oh, I think my brain went afk when I wrote this. You're absolutely right, sinusoidal embeddings happen before the query/key/value projections, making this just plain wrong. Don't know what happened there. I'll come up with a better example. 😄

equinox/nn/_embedding.py
Artur-Galstyan: This sounds interesting. I'll build it in and check if it has any effect on performance.

equinox/nn/_embedding.py
Artur-Galstyan: Regarding your code: it's indeed 2x faster (unjitted) and _a bit_ faster when jitted (outputs are exactly the same). 

Mine: 
```
181 µs ± 945 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Yours:
```
150 µs ± 569 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Since I don't see any downsides here, I'll change it to your implementation.

As for the rank promotions: When I added those into jupyterlab to test everything, it actually does raise issues! But it does that for the RoPE code too. I'll fix those of course but I'm not sure why it's not raised by the GH test runner.

equinox/nn/_embedding.py
Artur-Galstyan: Yes, this was a leftover. Since the layer has no more `Array`, it has no trainable parameters. If it still had the `freqs_cis` as a field - and if we didn't use `jax.lax.stop_gradient` it - probably (I'm a bit shaky on `jax.lax.stop_gradient`) - would have changed the array.

Artur-Galstyan: Interestingly, this makes quite a difference. With `stop_gradient` it's around 20% faster (on my MacBook, so it doesn't mean all that much - basically from 8 it/s to 10it/s). But the overall performance of the trained transformer is basically the same, a tiny bit better when not using `stop_gradient`. As it makes semantically more sense to _not_ stop the gradients, I'll keep it as is. But it's an interesting find.

Artur-Galstyan:
I've incorporated all the feedback. I really like how it all turned out. The docs look good too. I think you can re-review it.

equinox/nn/_embedding.py
```diff
@@ -194,9 +197,14 @@ def precompute_freqs_cis(
         freqs = 1.0 / (
             theta ** (jnp.arange(0, embedding_size, 2)[jnp.newaxis, :] / embedding_size)
         )
-        t = jnp.arange(end)
+
+        t = jnp.arange(end, dtype=float)
```
patrick-kidger, line 201: Heads-up that using `dtype=int` (and so perhaps also `dtype=float`?) can cause funny platform-dependent behaviour -- see https://github.com/google/jax/issues/9574.

If you want "default floating point precision" here then I think `jnp.arange(float(end))` will accomplish that: `jnp.arange` will check whether its argument is exact or inexact, and should then use the appropriate default dtype of that sort.

equinox/nn/_embedding.py
```diff
@@ -194,9 +197,14 @@ def precompute_freqs_cis(
...
         freqs_outer = jnp.outer(t, freqs)
-        freqs_cis = jnp.cos(freqs_outer) + jnp.sin(freqs_outer) * 1j
+        freqs_cis = (
+            jnp.array(jnp.cos(freqs_outer), dtype=jnp.complex64)
+            + jnp.array(jnp.sin(freqs_outer), dtype=jnp.complex64) * 1j
```
patrick-kidger, line 205: Likewise I'm not wild about the explicit dtype casts here, which may do the wrong thing if e.g. float64s (and thus complex128s) are what is being used for some reason. Or probably more likely, if e.g. bfloat16s are being used... are we going to the correct thing in all these cases?

If you're just trying to have the strict-dtype-promotion check not yell at you, then the easiest way to do it is to explicitly disable it locally where you know that what you really want the dtype promotion:
```python
with jax.numpy_dtype_promotion("standard"):
    ...
```

patrick-kidger (commented):
Haha, thank you! Developing all these libraries has given me a lot of practice bumping into edge cases :)

equinox/nn/_embedding.py
Artur-Galstyan: Fixed this for the sinusoidal embeddings, too. There, it was enough to make one of the arguments a float, which makes the whole thing a float, i.e.:
```python
jax.numpy.arange(0.0, 42, 2)
```

Artur-Galstyan:
Alright! I've fixed those `dtype` explicit casts, and things are looking good! I'll be off to work and check back in in the evening. Up for another round of reviews?

equinox/nn/_embedding.py
```diff
@@ -354,8 +352,7 @@ def get_positional_encoding(
         pos = jnp.arange(seq_len, dtype=float)[:, jnp.newaxis]
```
patrick-kidger, line 352: Heads-up that you still seem to have some `dtype=float`s

Artur-Galstyan: Oops, must have overlooked this, my bad. It's fixed now

Artur-Galstyan:
Hey @patrick-kidger, I've added 3 tests, that test:
- that we have the correct shapes
- that the `freqs_cis` are computed correctly
- that the output values are correct

I used code from the original llama implementation as well as huggingface's rope implementation. I documented the scripts which were used to generate the `expected_values` from the tests [here](https://gist.github.com/Artur-Galstyan/d33eda74072fea61545127adb90197b5) and [here](https://gist.github.com/Artur-Galstyan/8d0bb5743f00650aa6c0d7427595a0ff)

patrick-kidger:
Alright -- merged! This LGTM :D
Thank you for putting this together, and thank for your great patience. I'm really glad to have this in.

PR #573, opened 2023-10-24
Module signatures

danielward27:
Assign __signature__ in equinox modules.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ First fork the library on GitHub.
 Then clone and install the library in development mode:
 
 ```bash
-git clone https://github.com/your-username-here/equinox.git
+git clone https://github.com/patrick-kidger/equinox.git
 cd equinox
 pip install -e .
 ```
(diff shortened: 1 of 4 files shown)


patrick-kidger:
Unfortunately I think this still ends up breaking (as you found out with `filter_jit`):
```python
import inspect
import equinox as eqx


class A(eqx.Module):
    x: int

    def __call__(self, y: str):
        pass


a = A(x=1)
print(inspect.signature(a))  # (x: int) -> None
```

There is downstream code that does `inspect.signature(possibly_an_eqx_module_instance)`, and I think somehow that needs to continue working in the same way as it currently does.

danielward27:
Fwiw, if you define ``__signature__`` as a property of the metaclass instead
```python
@property
def __signature__(cls):
    sig = inspect.signature(cls.__init__)
    params = list(sig.parameters.values())[1:]  # Remove self parameter
    return sig.replace(parameters=params)
```
then this seems to yield the correct signature for callable modules (and avoids breaking the code in filter_jit), and the documentation still builds as expected. Again not really sure of why that occurs and if it breaks anything though.

danielward27:
By defining it as a property, when we instantiate a module like above,  ``__signature__`` is no longer found, which means ``__call__`` is used by ``inspect.signature`` instead.

patrick-kidger:
Excellent! Using a property like that is an elegant solution.

This LGTM, so merged!

PR #579, opened 2023-10-31
Fixing broken link to Enumueration objects

LouisDesdoigts:
The link to the `Enumeration` objects was stale, linking to the old docs location. This was propagating up through `Lineax` and `Optimisitx`.


patrick-kidger:
LGTM! Thank you for the fix.

PR #581, opened 2023-11-07
Fix re-initialisation method in `tricks.md`

homerjed:
Hi Patrick,

I was looking to re-initialise an MLP. I noticed the nice example you provided has (I think) a small bug. 

The weights are returned, regardless of their shape, as scalars. I fixed this in my commit.

(sorry not sure why my last commit is still there... git noob here)


patrick-kidger:
Thank you for the fix!
I've re-implemented this in #582 to fix the extra commit you mention. :)

patrick-kidger:
Closing in favour of #582. Thanks for spotting this!

PR #589, opened 2023-11-15
Better error message when no kwargs passed to ```filter_grad```

knyazer:
Hi Patrick,

I encountered incorrect error message during using ```eqx.filter_grad```:
```python
import equinox as eqx
from jax import numpy as jnp, tree_util as jtu
@eqx.filter_grad
def f(a):
    return jnp.sum(a)
eqx.filter_grad(jtu.Partial(f, a=jnp.ones((3,2))))()
```
Throws 
```
Functions wrapped with `equinox.filter_{grad, value and grad}` must have their 
first argument passed by position, not keyword. (This is the argument that is differentiated).
```
which does not make sense in this scenario: there are no keyword arguments!

I simply added a check for no keyword arguments, that now states that ```... there must be at least one positional argument ... ``` when no kwargs are available.

### DIFF
--- a/equinox/_ad.py
+++ b/equinox/_ad.py
@@ -62,11 +62,18 @@ class _ValueAndGradWrapper(Module):
             return self._fun(_x, *_args, **_kwargs)
 
         if len(args) == 0:
-            raise TypeError(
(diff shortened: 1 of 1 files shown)


patrick-kidger:
LGTM! Thank you for this improvement. :)

PR #597, opened 2023-11-23
Mixed precision for attention

### LINKED ISSUE #553: Mixed precision with existing Equinox modules
Hi, I am trying to use [jmp](https://github.com/google-deepmind/jmp) to implement mixed precision in a transformer model in Equinox. These [pytorch docs](https://pytorch.org/docs/stable/amp.html#cuda-op-specific-behavior) have a list of fp16-safe ops and fp16-unsafe ops.

Since softmax is not safe in fp16, I'd like to make sure that all softmaxes are in fp32. If I'm using [eqx.nn.MultiheadAttention](https://docs.kidger.site/equinox/api/nn/attention/#equinox.nn.MultiheadAttention), how can I make sure the softmax runs in fp32? 

```py
# x is a sequence of token vectors in fp32
# self.mhsa is an instance of eqx.nn.MultiheadAttention

x_ln = jax.vmap(self.layernorm1)(x)

# cast to fp16
x_ln, mhsa = self.policy.cast_to_compute((x_ln, self.mhsa))

# how do I make sure the softmax is run in fp32?
post_attn = self.mhsa(query=x_ln, key_=x_ln, value=x_ln, key=attn_key)  

# this will recast to fp32 because x is still fp32
x = x + post_attn  
```

mk-0:
resolves #553
I had the same problem as described in the issue in my own implementation. I added `softmax_dtype` as an argument.
Its type should be `jax.typing.DTypeLike`, but this type is not runtime checkable.
I made it a string for now, but I hope there is a better solution.

### DIFF
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -18,6 +18,7 @@ def dot_product_attention_weights(
     query: Float[Array, "q_seq qk_size"],
     key: Float[Array, "kv_seq qk_size"],
     mask: Optional[Bool[Array, "q_seq kv_seq"]] = None,
+    softmax_dtype: Optional[str] = None,
 ) -> Float[Array, "q_seq kv_seq"]:
     query = query / math.sqrt(query.shape[-1])
     logits = jnp.einsum("sd,Sd->sS", query, key)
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Hey there! Thanks for the PR.

I think I've since figured out a neater way to handle this, without needing an extra argument: cast the softmax dtype to at least `float32`, then downcast again:
```
dtype = jnp.result_type(logits.dtype, jnp.float32)
weights = jax.nn.softmax(logits.astype(dtype)).astype(logits.dtype)
```
WDYT?

mk-0:
Hey!
Yes, this is a better way, I updated the code. It won't work with `fp8_*` types though.

mk-0:
I can also do `logits = typing.cast(Array, logits)` to avoid all this `# pyright: ignore`, if you think it's preferable.

patrick-kidger:
I think this LGTM! (And the `cast` is appropriate, it's reflecting an oddity with the type annotations of `jnp.where` .)

Thanks for contributing -- merged! :)

PR #608, opened 2023-11-30
Bert Example: Add needed vmap to embeddings

haydn-jones:
Running the Bert example fails with
```
ValueError: `eqx.nn.Embedding()(x)` should be called with a scalar index `x`. Use `jax.vmap` if you would like to index with multiple values.
```
just added the vmap in.


patrick-kidger:
LGTM! Thank you for the fix. (This was changed to be more strict a few releases ago.)

PR #609, opened 2023-12-03
fix warning for unexpected param

Artur-Galstyan:
Removes the "Unexpected argument" warning.

### DIFF
--- a/equinox/nn/_sequential.py
+++ b/equinox/nn/_sequential.py
@@ -152,6 +152,9 @@ class Lambda(Module):
 
     fn: Callable[[Any], Any]
 
+    def __init__(self, fn: Callable[[Any], Any]):
+        self.fn = fn
+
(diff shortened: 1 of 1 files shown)


Artur-Galstyan:
This one:
image

(I'm using PyCharm)

patrick-kidger:
It looks like this is a PyCharm bug: https://youtrack.jetbrains.com/issue/PY-54560
(Perhaps consider switching to an alternative like VSCode / Vim / Helix / etc.)

On balance I'm inclined not to include this here -- adding `__init__` methods to all classes which currently use the default `__init__` is a lot of extra noise.

patrick-kidger:
Haha, sorry!
FWIW allow me to recommend [Helix](https://github.com/helix-editor/helix), which is vim-like whilst being "batteries included": LSP, tree-sitter, file explorer etc. I'm a big fan!

PR #624, opened 2023-12-18
Add weight normalization

boris-kuz:
Hello!

Trying to make up for abandoning that last PR :)

Came across the lack of weight normalization in Equinox when trying to implement a few papers I liked in Jax. I took inspiration from the [SpectralNorm](https://github.com/patrick-kidger/equinox/blob/main/equinox/nn/_spectral_norm.py) implementation in equinox, and the [weight_norm](https://pytorch.org/docs/stable/generated/torch.nn.utils.weight_norm.html) implementation in PyTorch.

I could also add an inverse transformation which returns the layer contained in the WeightNorm object with the trained weights matrix set, if that would be useful?

Thanks for all of your great work!


equinox/nn/_weight_norm.py
```diff
@@ -0,0 +1,100 @@
...
+_Layer = TypeVar("_Layer")
+
+
+def _norm_except_axis(
+    v: Array, pow: Optional[Union[int, str]] = None, axis: Optional[int] = 0
+) -> Array:
+    for ax in range(len(v.shape)):
+        if ax != axis:
+            v = jnp.linalg.norm(v, ord=pow, axis=ax, keepdims=True)
+    return v if axis is not None else v.reshape([])
```
patrick-kidger, line 20: I think it should be possible to express this function as just a single `norm`, without needing to make repeated calls. (And if `axis is not None` then `jax.vmap` over that axis.)

equinox/nn/_weight_norm.py
```diff
@@ -0,0 +1,100 @@
...
+
+class WeightNorm(Module, Generic[_Layer]):
+    r"""
+    Applies weight normalisation to a given parameter.
+
+    Given a weight matrix $\mathbf{W}$, computes the follow reparametrization:
+
+    $\mathbf{W} = g \frac{\mathbf{v}}{\lVert \mathbf{v} \rVert}$
+
+    where $g$ is initially chosen to equal $\lVert \mathbf{v} \rVert$.
```
patrick-kidger, line 31: and `v` is initially chosen as `W`.

equinox/nn/_weight_norm.py
```diff
@@ -0,0 +1,100 @@
...
+        axis: Optional[int] = 0,
+    ):
+        """**Arguments:**
+
+        - `layer`: The layer to wrap. Usually a [`equinox.nn.Linear`][] or
+            a convolutional layer (e.g. [`equinox.nn.Conv2d`][]).
+        - `weight_name`: The name of the layer's parameter (a JAX array) to apply
+            weight normalisation to.
+        - `axis`: The norm is computed across every axis except this one.
+                  If `None`, compute across every axis.
```
patrick-kidger, line 77: Nit: the convention is to indent by four spaces regardless of the length of the name of the argument.

patrick-kidger (commented):
Thanks for the contribution!

Nits aside this LGTM. :) If you can make the small tweaks then I'd be happy to merge this.

equinox/nn/_weight_norm.py
```diff
@@ -0,0 +1,107 @@
...
+    norm_func = lambda x: jnp.linalg.norm(x, ord=pow, keepdims=True)
+    vmapped_norm_func = lambda axis: jax.vmap(
+        norm_func,
+        in_axes=axis,
+        out_axes=axis,
+    )(v)
+    if axis is not None:
+        return vmapped_norm_func(axis)
+    else:
+        return norm_func(vmapped_norm_func(0)).reshape([])
```
boris-kuz, line 26: Did you have something like this in mind @patrick-kidger? I couldn't figure out how to vmap over all axes, so I opted for this solution. I guess this could also be achieved as a scan? Though I'm not sure what would be gained by this.

boris-kuz:
Thanks for your quick response and happy holidays! I pushed a change addressing your comments :)

equinox/nn/_weight_norm.py
patrick-kidger: So I think something like this might be what you're after:
```python
from jaxtyping import Array, PyTree, Scalar

def _norm_except_axis(v: Array, norm: Callable[[PyTree[Array]], Scalar], axis: Optional[int]):
    if axis is None:
        return norm(x)
    else:
        return jax.vmap(norm, in_axes=axis, out_axes=axis)(x)
```
where for bonus points I've also (a) removed default arguments, and (b) made the norm selectable (so that one can pass in any norm function -- we have a number of these defined elsewhere in the Equinox ecosystem, we could imagine consolidiating them here). In practice right now we can call this with `norm=lambda x: jnp.linalg.norm(x, keepdims=True)`.

WDYT?

boris-kuz: This looks great! Somehow I thought that passing `None` to `jnp.linalg.norm` doesn't work - not sure why.

One minor addition, we'll have to reshape the `if axis is None` branch if we want the same behaviour as the torch equivalent. We've already strayed from it a little by having `axis=None` denote computing the norm over all dimensions as opposed to `-1` (though I think that is the better approach here, as the torch API is confusing and not consistent, as `-2` for example works like indexing, i.e. computes the norm over the second to last axis), so I guess we could further abandon a strict adherence here.

Additionally, if we're now making this function a bit more general purpose, it might make sense to move it elsewhere and add dedicated tests for it, what do you think? Do you have a good place for this in mind? Nothing immediately jumps out at me scrolling through the codebase.

Do we want to add defaulted arguments also?

So taken together it would look like this:

```python
from functools import partial

def _norm_except_axis(v: Array
, norm: Callable[[PyTree[Array]], Scalar] = partial(jnp.linalg.norm, keepdims=True)
, axis: Optional[int] = 0):
    if axis is None:
        return norm(v).reshape([])
    else:
        return jax.vmap(norm, in_axes=axis, out_axes=axis)(v)
```

patrick-kidger: Yup, I think departing from the PyTorch API makes sense here.

I think I'm happy to leave the `_norm_except_axis` function here. It's pretty small, we can worry about that later if we ever need it. :)

I definitely wouldn't add default arguments. It's a general good practice for internal APIs not to include default arguments, as these can be a source of silently surprising behaviour. (Especially here, where we are always passing every argument explicitly anyway!)

boris-kuz: Good point regarding the defaulted arguments, I agree.

Leaving the function where it is then for now then :) 

I've made all the changes discussed - additionally, I've imported from `beartype.typing` as opposed to `typing` directly to silence a PEP 585 related warning (see [here](https://beartype.readthedocs.io/en/latest/api_roar/#pep-585-deprecations)). I hope this is OK? It looks like `equinox` has a  dependency on`beartype` already (correct me if I'm wrong!)

patrick-kidger: Equinox should only use `beartype` for its tests -- we don't have it as a dependency outside of that.

So in our case the appropriate thing to do is:
```python
from typing import Generic, Optional, TypeVar
```

(A lot of things in the `typing` module have been deprecated: some in favour of a builtin, e.g. `List -> list`, and some in favour of an item of `collections.abc`, e.g. `Callable`. Fixing up these warnings is usually a matter of making one of those two changes.)

equinox/nn/_weight_norm.py
```diff
@@ -0,0 +1,104 @@
...
+        - `weight_name`: The name of the layer's parameter (a JAX array) to apply
+        weight normalisation to.
+        - `axis`: The norm is computed across every axis except this one.
+        If `None`, compute across every axis.
+        """
+        self.layer = layer
+        self.weight_name = weight_name
+        self.axis = axis
+
+        self.v = getattr(layer, weight_name)
```
patrick-kidger, line 84: Oh, I've just spotted a minor bug in this implementation. We now have both `self.v` and `self.layer.weight`, both of which are the "weight" of this layer.

[This means each version of the weight is treated as a separate copy.](https://docs.kidger.site/equinox/faq/#how-to-share-a-layer-across-two-parts-of-a-model)

I think the fix is to remove `self.v` from `__init__` entirely. And then in the forward pass, do:
```python
v = getattr(self.layer, self.weight_name)
weight = self.g * v / self._norm(v)
layer = tree_at(lambda l: getattr(l, self.weight_name), self.layer, weight)
return layer(x)
```

equinox/nn/_weight_norm.py
boris-kuz: import `Callable` from `collections.abc` fixed the warning, thank you! `beartype` should suggest this as a fix in their documentation. Might make a PR for it.

equinox/nn/_weight_norm.py
boris-kuz: Yeah, I initially chose to implement it this way to cleanly separate what is actually being trained - `v` and `g` and not the `weight` of the layer directly. But that distinction is definitely not worth wasting memory.
This makes me think though, maybe it'd be worth it to add a "commit" function or something like that, which returns the contained layer with the weight matrix set to its final value by evaluating `g * v/norm(v)`. This can be done manually by the user also of course, but that's not very ergonomic. WDYT?

boris-kuz: So something like this:
```python
    def commit(self) -> PyTree:
        v = getattr(self.layer, self.weight_name)
        weight = v * self.g / self._norm(v)
        layer = tree_at(lambda l: getattr(l, self.weight_name), self.layer, weight)
        return layer
```

Which would simplify forward like so:
```python
        return self.commit()(x)
```.

Or alternatively we'd change the way this behaves during inference, where it's not necessary to recalculate `g * v/norm(v)` every forward pass - where we could overwrite `self.layer.weight` with the final weight matrix and simply reverse the operation for training mode.

patrick-kidger: No strong feelings on whether to pull this functionality out into a `commit` method. Do it or don't as you prefer I think! :)

As for inference-time optimisations, this is an interesting point. I think I'd say for now that we shouldn't worry about this: there's a lot of these optimisations that we don't do at the moment, and they often tend to also involve fusing layers together, this kind of thing. So perhaps best to find another way to tackle this problem.

boris-kuz: I'll leave it be for now then - if someone creates an issue for this, I'll be happy to add it. Feel free to tag me if this should happen!

Pushed all changes now :)

patrick-kidger:
Alright, marvellous!

Thank you for your efforts, and I've just merged this into the dev branch. It'll appear in the next release of Equinox :)

PR #629, opened 2023-12-27
Add RMSNorm

jondeaton:
Branched from #545 addressing reviewers' comments.

### DIFF
--- a/equinox/nn/_normalisation.py
+++ b/equinox/nn/_normalisation.py
@@ -372,18 +372,11 @@ class RMSNorm(Module):
                 "`RMSNorm(shape)(x)` must satisfy the invariant `shape == x.shape`"
                 f"Received `shape={self.shape} and `x.shape={x.shape}`. You might need "
                 "to replace `rms_norm(x)` with `jax.vmap(rms_norm)(x)`.\n"
-                "\n"
-                "If this is a new error for you, it might be because this became "
-                "stricter in Equinox v0.11.0. Previously all that was required is that "
-                "`x.shape` ended with `shape`. However, this turned out to be a "
-                "frequent source of bugs, so we made the check stricter!"
             )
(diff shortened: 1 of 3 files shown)


equinox/nn/_normalisation.py
```diff
@@ -267,3 +267,117 @@ def __call__(
...
+
+    $$\frac{x}{\sqrt{\varepsilon + \frac{1}{n}\Vert x \Vert^2_2}} \gamma + \beta$$
+
+    where $\Vert x \Vert^2_2 = \sum_{i=1}^n x_i^2$, $n = \dim(x)$, and $\gamma$ is a
+    learned array with the same shape as $x$ if `use_weight=True`, or
+    $\gamma = 1$ if `use_weight=False`, as proposed in
+    [this paper](https://browse.arxiv.org/abs/2307.14995). `\beta` is an optional bias
+    term.
+
+    ??? cite
```
patrick-kidger, line 287: Nit: this will need a new line here in order to render correctly in the docs.

i.e.
```python
??? cite

    foo
```

patrick-kidger: Also, whilst we're here -- maybe worth emphasising that `use_{weight,bias}=False` is the default.

(Should we keep it as the default? I appreciate that's maybe more useful, but it's inconsistent with the other normalisation layers. WDYT?)

patrick-kidger (commented):
Doc nit aside, this LGTM! Thank you for tidying this up.
I'll merge this into the new `dev` branch shortly.

equinox/nn/_normalisation.py
jondeaton: Thanks for raising the point about the defaults. I feel its good to maintain consistency, especially with `LayerNorm`. Changed defaults to on.

patrick-kidger:
Alright, LGTM! Thank you for tidying this one up.
I've merged it into the dev branch, so this will appear in the next release of Equinox :)

PR #637, opened 2024-01-10
Fix documentation bug

lockwo:
In the documentation with the new RMS the citation is not rendering properly
Screenshot 2024-01-10 at 4 48 44 PM

I believe this is just because the } was offset, so I changed that to match the other citations


patrick-kidger:
LGTM!

Thank you for the quick fix :)

PR #656, opened 2024-02-13
Add Mamba Block

Artur-Galstyan:
This PR adds the Mamba Block. 

It's still in early draft modus. Along with the Mamba Block, the full Mamba model will be shown in the docs as an example.


Artur-Galstyan:
@patrick-kidger This isn't 100% done yet of course, but the example `mamba.iypnb` is almost done. The actual implementation of the `SelectiveStateSpaceModel` module is in there and will later be transferred to its own file. Besides the `TODO` in the example, what do you think of it? Anything extra you would like to see there?

equinox/nn/_selective_state_space_models.py
```diff
@@ -0,0 +1,198 @@
...
+
+    x_res = jnp.zeros(shape=(d_inner, state_space_dims))
+
+    def step(x, i):
+        x = delta_A[i] * x + delta_B_u[i]
+
+        y = jnp.einsum("d n,n -> d", x, C[i, :])
+        return x, y
+
+    _, ys = jax.lax.scan(step, x_res, jnp.arange(seq_len))
```
jondeaton, line 35: This should really be done with `jax.lax.associative_scan` for computing the scan in parallel. This is more faithful to the original and is much faster for long sequences on parallel processing accelerators (e.g. GPU).  Its unclear to me based on the [Griffin paper](https://arxiv.org/abs/2402.19427) if JAX's associative scan will be fater than linear scan on TPUs also, so perhaps there should be an option to choose which implementation to use.

See the S5 JAX implementation for a very similar implementation with associative scan https://github.com/lindermanlab/S5/blob/main/s5/ssm.py

patrick-kidger: I'm not sure about this, actually. JAX's associative scan appears to completely unroll the whole operation:

https://github.com/google/jax/blob/63ceb5f539c45fe00766634ce7b01ea5176e0cc4/jax/_src/lax/control_flow/loops.py#L2171-L2186

I think there's probably a better way to implement an associative scan. Probably doing something like `equinox/internal/loop/bounded.py`, or possibly a `lax.scan` over the levels of the tree.

examples/mamba.ipynb
patrick-kidger, line 1: Do please credit yourself here! See the way it's done in some of the other advanced examples.

Also, let's add this file to `mkdocs.yml`.

equinox/nn/_selective_state_space_models.py
```diff
@@ -0,0 +1,198 @@
...
+            key=x_proj_key,
+        )
+
+        self.dt_proj = Linear(
+            self.dt_rank, self.d_inner, use_bias=True, key=dt_proj_key
+        )
+
+        A = jnp.repeat(jnp.arange(1, self.state_space_dims + 1), self.d_inner).reshape(
+            self.d_inner, self.state_space_dims
+        )
```
aj111000, line 153: I'm assuming this is supposed to initialize A to have identical rows representing a sequence of identical diagonal matrices (with A_ii = i as in the paper) compactly, e.g. 
A = [1, 2, 3, 4; 
        1, 2, 3, 4; 
        1, 2, 3, 4]
if state_space_dims = 4 and d_inner = 3. However, this line will actually make
A = [1, 1, 1, 2; 
        2, 2, 3, 3; 
        3, 4, 4, 4]

To fix this, you can just replace the reshape command with:
.reshape(self.state_space_dims, self.d_inner).transpose()

Artur-Galstyan:
Alright, I've added those changes in, added those to the docs, added a little training script for the MAMBA example and some model outputs too. But I think there's still a faulty test here. I might have to `git rebase dev` if they still fail.

Issue #638, opened 2024-01-11
[Feature Request] Allow to specify `Conv` padding by strings


ChenAo-Phys:
Hello! I hope to specify the padding in `Conv` by strings like `'SAME'` or `'VALID'`. This is compatible with jax API since it's already implemented in `jax.lax.conv_general_dilated`.

Furthermore, I also would like to have a `'CIRCULAR'` padding which is not in `jax.lax.conv_general_dilated` but implemented in flax and pytorch. In the 1D case, for instance, this padding behaves as `x[L + i] = x[i]` so that the chain becomes effectively a ring. This is particularly helpful in some physical systems.

I can propose a PR if you think these changes are suitable. Thanks!

patrick-kidger:
I think that sounds reasonable to me. :)
I'd be happy to take a pull request adding these options.

Do you know how the circular option might be implemented efficiently? Done naively I suspect we might end up copying the input.

ChenAo-Phys:
> I think that sounds reasonable to me. :) I'd be happy to take a pull request adding these options.
> 
> Do you know how the circular option might be implemented efficiently? Done naively I suspect we might end up copying the input.

As far as I know, there is no "clever" way to do circular padding other than copying the input, probably because the circular padding is not internally supported in cudnn. That's also how other packages do it.

As an aside, would you like to also accept string padding in `ConvTranspose` and `Pool`? This will make the API more self-consistent.

patrick-kidger:
> As far as I know, there is no "clever" way to do circular padding other than copying the input, probably because the circular padding is not internally supported in cudnn. That's also how other packages do it.

Ah well. Makes sense.

> As an aside, would you like to also accept string padding in ConvTranspose and Pool? This will make the API more self-consistent.

Absolutely! That'd be great for consistency.

ChenAo-Phys:
I have been testing the new codes these days and it works quite well. There are some details I want to clarify before submitting the PR.

- `Conv`
`"CIRCULAR"` is actually a "padding mode" specifying how to fill padding values instead of how many additional units should be appended to the edge. In [pytorch](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html) they have two different arguments, `padding` and `padding mode`, but in our case this will complicate the API. So I define `"CIRCULAR"` to be the same amount of padding as "`SAME`" filled with circular values instead of zeros, which is a suitable choice in most physical scenarios.

- `ConvTranspose`
[...]

PR #658, opened 2024-02-15
New padding options for `Conv` and `ConvTranspose`

ChenAo-Phys:
According to #638, some changes are made to the padding in `Conv` and `ConvTranspose`.

The `padding` argument now supports three string inputs, "VALID", "SAME", or "SAME_LOWER", similar to [`jax.lax.conv_general_dilated`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.conv_general_dilated.html).

Furthermore, there is a new argument `padding_mode` which accepts string inputs including "ZEROS" (default), "REFLECT", "REPLICATE", and "CIRCULAR", similar to [PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html).

To make the "SAME" or "SAME_LOWER" padding work for `ConvTranspose`, the definition of `output_padding` is slightly changed to make `output_size = (input_size + output_padding) / stride` in these cases.

Only "ZEROS" and "CIRCULAR" are implemented in `ConvTranspose`. "CIRCULAR" further requires `output_padding == 0` 
 and `padding == "SAME"` or `"SAME_LOWER"` to simplify it, which is the typical case when people use "CIRCULAR" padding.

### DIFF
--- a/equinox/nn/_conv.py
+++ b/equinox/nn/_conv.py
@@ -32,6 +32,41 @@ def _ntuple(n: int) -> Callable[[Union[_T, Sequence[_T]]], tuple[_T, ...]]:
     return parse
 
 
+def _padding_init(
+    padding: Union[str, int, Sequence[int], Sequence[tuple[int, int]]],
+    num_spatial_dims: int,
+) -> Union[str, tuple[tuple[int, int], ...]]:
+    if isinstance(padding, str):
+        padding = padding.upper()
+        if padding not in ("SAME", "SAME_LOWER", "VALID"):
+            raise ValueError(
+                "`padding` string must be `'SAME'`, `'SAME_LOWER'`, or `'VALID'`."
+            )
+    elif isinstance(padding, int):
+        padding = tuple((padding, padding) for _ in range(num_spatial_dims))
+    elif isinstance(padding, Sequence) and len(padding) == num_spatial_dims:
+        if all_sequences(padding):
+            padding = tuple(padding)
+        else:
+            padding = tuple((p, p) for p in padding)
+    else:
+        raise ValueError(
+            "`padding` must either be a string, an int, or tuple of length "
+            f"{num_spatial_dims} containing ints or tuples of length 2."
+        )
+    return padding
+
+
+def _padding_mode_init(padding_mode: str) -> str:
+    padding_mode = padding_mode.upper()
+    if padding_mode not in ("ZEROS", "REFLECT", "REPLICATE", "CIRCULAR"):
+        raise ValueError(
+            "`padding_mode` must be `'ZEROS'`, `'REFLECT'`, `'REPLICATE'`, or "
+            "`'CIRCULAR'`."
+        )
+    return padding_mode
+
+
 class Conv(Module, strict=True):
     """General N-dimensional convolution."""
 
@@ -42,10 +77,11 @@ class Conv(Module, strict=True):
     out_channels: int = field(static=True)
     kernel_size: tuple[int, ...] = field(static=True)
     stride: tuple[int, ...] = field(static=True)
-    padding: tuple[tuple[int, int], ...] = field(static=True)
+    padding: Union[str, tuple[tuple[int, int], ...]] = field(static=True)
     dilation: tuple[int, ...] = field(static=True)
     groups: int = field(static=True)
     use_bias: bool = field(static=True)
+    padding_mode: str = field(static=True)
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


equinox/nn/_conv.py
```diff
@@ -80,6 +116,8 @@ def __init__(
             concatenating the results along the output channel dimension.
             `in_channels` must be divisible by `groups`.
         - `use_bias`: Whether to add on a bias after the convolution.
+        - `padding_mode`: One of the following strings specifying the padding values.
+            `'ZEROS'` (default), `'REFLECT'`, `'REPLICATE'`, or `'CIRCULAR'`.
```
patrick-kidger, line 120: Can you explain what each of these options does?

equinox/nn/_conv.py
```diff
@@ -134,21 +174,34 @@ def __init__(
...
         self.groups = groups
         self.use_bias = use_bias
+        self.padding_mode = _padding_mode_init(padding_mode)
+
+    def _nonzero_pad(self, x: Array) -> Array:
+        if isinstance(self.padding, str):
+            rhs_shape = tuple(
+                d * (k - 1) + 1 for k, d in zip(self.kernel_size, self.dilation)
+            )
+            padding = lax.padtype_to_pads(
```
patrick-kidger, line 188: I don't think this is public API, unfortunately. ([It's not listed here, anyway.](https://jax.readthedocs.io/en/latest/jax.lax.html)) Maybe we can ask to add it to the public API for JAX?

equinox/nn/_conv.py
```diff
@@ -134,21 +174,34 @@ def __init__(
...
+                x.shape[2:], rhs_shape, self.stride, self.padding
```
patrick-kidger, line 189: Why `[2:]`?
`x` has shape `(in_channels, dim_1, ..., dim_N)`, so this is trimming off `dim_1` as well.

equinox/nn/_conv.py
```diff
@@ -393,27 +466,72 @@ def __init__(
...
             self.bias = None
 
+        padding = _padding_init(padding, num_spatial_dims)
+        padding_mode = _padding_mode_init(padding_mode)
+        if padding_mode in ("REFLECT", "REPLICATE"):
+            raise NotImplementedError(
+                "'REFLECT' or 'REPLICATE' padding mode is not implemented"
+            )
+        elif padding_mode == "CIRCULAR":
+            if np.any(output_padding):
```
patrick-kidger, line 476: Nit: I prefer avoiding implicit-cast-to-bool. (Here for the int->bool conversion on each element.) Can this be `if any(x !=0 for x in output_padding)`?

patrick-kidger (commented):
Thank you for the contribution! This is excellent; thank you for implementing so much.

My main two comments are:
1. Can you add some more documentation to the docstrings? (E.g. what does it mean to use `padding_mode == "CIRCULAR"` in `ConvTranspose`?)
2. Can you add some tests? I think I spot a few cases where the wrong thing is going to happen at the moment (I've commented on a couple of them below.)

equinox/nn/_conv.py
ChenAo-Phys: A batch dimension is expanded in line 226, so the shape is `(batch, in_channels, dim_1 , ...)`.

patrick-kidger: Ah, makes sense! Thanks.

equinox/nn/_conv.py
ChenAo-Phys: Could you raise a new issue in JAX? I think your proposal is much more likely to be accepted

ChenAo-Phys:
> Thank you for the contribution! This is excellent; thank you for implementing so much.
> 
> My main two comments are:
> 
> 1. Can you add some more documentation to the docstrings? (E.g. what does it mean to use `padding_mode == "CIRCULAR"` in `ConvTranspose`?)
> 2. Can you add some tests? I think I spot a few cases where the wrong thing is going to happen at the moment (I've commented on a couple of them below.)

Some more documentation and tests have been added according to your suggestion. Thanks a lot!

equinox/nn/_conv.py
patrick-kidger: Done! We'll see what they say.

patrick-kidger:
Looks like the pre-commit hooks are failing. (See CONTRIBUTING.md for setting these up locally.)

Modulo that detail, this PR LGTM! Thank you for putting it together, it'll be a great addition to have. Let me know once you've got the tests passing and I'll merge this PR.

ChenAo-Phys:
Some type hints have been fixed.

There seems to be something wrong with the test. It always reports 
`FAILED tests/test_errors.py::test_tracetime - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.`

patrick-kidger:
Rebase on top of the latest `dev`, I fixed this one a few days ago :)

patrick-kidger:
And merged! Thank you for the contribution. this is great to have. This will appear in the next release of Equinox :)

ChenAo-Phys:
> And merged! Thank you for the contribution. this is great to have. This will appear in the next release of Equinox :)

Nice! Also thanks for your time!

PR #661, opened 2024-02-20
Tweaked new `Conv` and `ConvTranspose` padding options

patrick-kidger:
Tagging @ChenAo-Phys as I think I've bumped into a bug.

I've added an assert statement that when `padding='SAME'/'SAME_LOWER'`, the input and output spatial dimensions should be the same. This causes a couple of tests to fail. It looks like the combination of `padding='SAME'/'SAME_LOWER'` with `padding_mode != 'ZEROS'` and `stride > 1` does not produce the right amount of padding.

This PR:
- Adds these now-failing assert statements.
- Refactors some docstrings (primarily so that they render correctly in the online docs).
- Switches the order of `jnp.expand_dims(x, axis=0)` and `self._nonzero_pad`, so that the unsqueezing/squeezing happens directly either side of the convolution. This is just for cleaner readability, really. I've checked that this is unrelated to the failing asserts.

Do you think you can take a look at what it would take to fix this up? (By submitting a PR against this branch.) Else I'm afraid we may have to revert #658 for the time being.

### DIFF
--- a/equinox/nn/_conv.py
+++ b/equinox/nn/_conv.py
@@ -117,29 +117,36 @@ class Conv(Module, strict=True):
             `in_channels` must be divisible by `groups`.
         - `use_bias`: Whether to add on a bias after the convolution.
         - `padding_mode`: One of the following strings specifying the padding values.
-            `'ZEROS'` (default): pads with zeros, 1234 -> 00123400.
-            `'REFLECT'`: pads with the reflection on boundary, 1234 -> 32123432.
-            `'REPLICATE'`: pads with the replication of edge values, 1234 -> 11123444.
-            `'CIRCULAR'`: pads with circular values, 1234 -> 34123412.
+            - `'ZEROS'` (default): pads with zeros, `1234 -> 00123400`.
+            - `'REFLECT'`: pads with the reflection on boundary, `1234 -> 32123432`.
+            - `'REPLICATE'`: pads with the replication of edge values,
+                `1234 -> 11123444`.
+            - `'CIRCULAR'`: pads with circular values, `1234 -> 34123412`.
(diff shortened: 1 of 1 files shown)


ChenAo-Phys:
OK, I will check it

ChenAo-Phys:
I think my code is correct. There are some problems with the assertion.

- For `Conv` the assertion is `assert tuple(x.shape[1:]) == tuple(input_shape)`. However, the input and output sizes are not always the same for "SAME" padding (sounds wierd). It's actually defined as 
`output_shape = ceil(input_shape / stride)`. 
So this assertion is correct only if `stride = 1`. The correct assertion should be 
`assert all(-(i // -s) == o for i, o, s in zip(input_shape, x.shape[1:], self.stride))`
(here `-(i // -s)` is ceil division)

- For `ConvTranspose` the assertion is `assert x.shape[1:] == input_shape`. One mistake here is that `x.shape` is a tuple while `input_shape` is a list, so they are always unequal. The previous problem in `Conv` still exists. In `ConvTranspose` it should be
`output_shape = (input_shape + output_padding) / stride`.
So the correct assertion should be
`assert all((i + a) // s == o for i, o, s, a in zip(x.shape[1:], input_shape, self.stride, self.output_padding))`

I have tested the new assertions in pytest and it works well. You could also check our previous discussion #638 for details, although I have to admit I didn't make it fully clear there.

ChenAo-Phys:
No problem. Thanks for polishing my codes :D

ChenAo-Phys:
`_, *input_shape = x.shape` was not deleted

patrick-kidger:
Good spot -- I've just pushed a commit to the `dev` branch to remove these.
Thank you for being a second pair of eyes!

PR #670, opened 2024-03-01
Make code that parses callback errors more tolerant of message format.

hawkinsp:
JAX is switching from pybind11 to nanobind, and nanobind formats this error differently, with the traceback preceding the EqxRuntimeError message. If we change the "At:" split to accept zero or more "At:" pieces, things work as intended with both versions.


patrick-kidger:
LGTM, thank you!

(Ignoring the CI error which is a known issue already fixed on `dev`.)

Issue #659, opened 2024-02-17
BatchNorm training instability


andrewdipper:
Hi, I think the use of running_mean and running_var during training time in BatchNorm causes training instability and increased learning_rate sensitivity.  With momentum low (say 0.5) the layer works fine but this is poor for test time due to the short duration of the running average.  When momentum is at the default 0.99 training performance is degraded (higher train losses).

I originally observed this in a larger ResNet model but used the simplified introductory equinox CNN model to verify.  Plotting the l2 norms of the gradients they spike 2-3 iterations in.  I haven't fully traced the behavior but I think it comes from the gradient flow effectively being clipped by (1 - momentum) when the running_mean / running_var is used as opposed to batch_mean / batch_var.  

I haven't seen the use of running_mean, running_var for the training side.  It's nice that the forward behavior is the same for train and test time but what is the expected behavior for the gradients?  I'm not sure if I'm missing something

I included my generated plots and quick test with the introductory CNN example.  Ema batch norm is eqx.nn.BatchNorm and the batch batch norm plot is using batch_mean and batch_var instead of running_mean and running_var during train time.

At least it might be worth noting the batch norm behavior - it took some time for me to chase my particular performance issue to BatchNorm.

Thanks!

![image](https://github.com/patrick-kidger/equinox/assets/5267190/a2a423a8-8c0e-4ce8-a6f9-992260f9a6dd)
![image](https://github.com/patrick-kidger/equinox/assets/5267190/924e52dd-24c0-46a1-8243-f6e8fc070335)
![image](https://github.com/patrick-kidger/equinox/assets/5267190/2d3e4e3b-e0b3-4c92-a165-b0d66aede3c0)
![image](https://github.com/patrick-kidger/equinox/assets/5267190/f8137319-01c5-4a5a-b9f8-65e33b41644b)

```

import equinox as eqx
import jax
import jax.numpy as jnp
import optax  # https://github.com/deepmind/optax
import torch  # https://pytorch.org
import torchvision  # https://pytorch.org
from jaxtyping import Array, Float, Int, PyTree  # https://github.com/google/jaxtyping
import localbatchnorm

import matplotlib.pyplot as plt

jax.config.update('jax_platform_name', 'cpu')

BATCH_SIZE = 64
LEARNING_RATE = 3e-3
STEPS = 100
PRINT_EVERY = 10
SEED = 5678

key = jax.random.PRNGKey(SEED)

normalise_data = torchvision.transforms.Compose(
    [
[...]

PR #675, opened 2024-03-07
BatchNorm training instability fix

andrewdipper:
This is in reference to issue [659](https://github.com/patrick-kidger/equinox/issues/659).

I modified BatchNorm to have two approaches `"batch"` and `"ema"`.  `"batch"` just uses the batch statistics during training time.  If approach is not specified it defaults to `"batch"` with a warning.  It's robust and seems to be the standard choice - it's far less likely to kill a model just by adding it.

`"ema"` is based of the smooth start method in the above issue.  So keep a running mean and variance but instead of renormalizing Adam style the parts of the running averages that are zeroed are filled with the batch statistics.   The problem is it's still not robust - the momentum parameter is simultaneously specifying a warmup period (when we're expecting the input distribution to change significantly) and how long we want the running average to be.  So I added a linear warmup period.

Now for any choice of momentum there seems to be a `warmup_period` choice that will give good results.  And validation performance was at least as good as with batch mode for my tests.  However, I don't see a good default for `warmup_period`.

Some considerations:
- having `approach="batch"` and the common `axis_name="batch"` is a little awkward
- There's an example using BatchNorm - that will start raising a warning and should probably get changed
- The current BatchNorm behavior can't be exactly replicated (ema / momentum=0.99 / warmup_period=1) is close but different at the start
- There's one more piece of state hence the test_stateful.py change.  Though this could be conditionally removed for `approach="batch"` if desired

Let me know what you think or if any changes or tests need to be added

image

### DIFF
--- a/equinox/nn/_batch_norm.py
+++ b/equinox/nn/_batch_norm.py
@@ -53,7 +53,7 @@ class BatchNorm(StatefulLayer, strict=True):
     axis_name: Union[Hashable, Sequence[Hashable]]
     inference: bool
     input_size: int = field(static=True)
-    approach: Union[None, str] = field(static=True)
+    approach: Literal["batch", "ema"] = field(static=True)
     eps: float = field(static=True)
     channelwise_affine: bool = field(static=True)
     momentum: float = field(static=True)
--- a/equinox/nn/_batch_norm.py
+++ b/equinox/nn/_batch_norm.py
@@ -1,10 +1,11 @@
+import warnings
 from collections.abc import Hashable, Sequence
-from typing import Optional, Union
+from typing import Literal, Optional, Union
 
 import jax
 import jax.lax as lax
 import jax.numpy as jnp
-from jaxtyping import Array, Bool, Float, PRNGKeyArray
+from jaxtyping import Array, Float, Int, PRNGKeyArray
 
 from .._misc import default_floating_dtype
 from .._module import field
@@ -44,24 +45,29 @@ class BatchNorm(StatefulLayer, strict=True):
 
     weight: Optional[Float[Array, "input_size"]]
     bias: Optional[Float[Array, "input_size"]]
-    first_time_index: StateIndex[Bool[Array, ""]]
+    count_index: StateIndex[Int[Array, ""]]
     state_index: StateIndex[
         tuple[Float[Array, "input_size"], Float[Array, "input_size"]]
     ]
+    zero_frac_index: StateIndex[Float[Array, ""]]
     axis_name: Union[Hashable, Sequence[Hashable]]
     inference: bool
     input_size: int = field(static=True)
+    approach: Union[None, str] = field(static=True)
     eps: float = field(static=True)
     channelwise_affine: bool = field(static=True)
     momentum: float = field(static=True)
+    warmup_period: int = field(static=True)
 
     def __init__(
         self,
         input_size: int,
         axis_name: Union[Hashable, Sequence[Hashable]],
+        approach: Optional[Literal["batch", "ema"]] = None,
         eps: float = 1e-5,
         channelwise_affine: bool = True,
         momentum: float = 0.99,
+        warmup_period: int = 1000,
         inference: bool = False,
         dtype=None,
     ):
@@ -71,11 +77,17 @@ class BatchNorm(StatefulLayer, strict=True):
         - `axis_name`: The name of the batch axis to compute statistics over, as passed
             to `axis_name` in `jax.vmap` or `jax.pmap`. Can also be a sequence (e.g. a
             tuple or a list) of names, to compute statistics over multiple named axes.
+        - `approach`: The approach to use for the running statistics. If `approach=None`
+            a warning will be raised and approach will default to `"batch"`. During
+            training `"batch"` only uses batch statisics while`"ema"` uses the running
+            statistics.
         - `eps`: Value added to the denominator for numerical stability.
         - `channelwise_affine`: Whether the module has learnable channel-wise affine
             parameters.
         - `momentum`: The rate at which to update the running statistics. Should be a
             value between 0 and 1 exclusive.
+        - `warmup_period`: The period to warm up the running statistics. Only used when
+            `approach=\"ema\"`.
         - `inference`: If `False` then the batch means and variances will be calculated
             and used to update the running statistics. If `True` then the running
             statistics are directly used for normalisation. This may be toggled with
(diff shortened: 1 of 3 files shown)


equinox/nn/_batch_norm.py
```diff
@@ -86,26 +98,37 @@ def __init__(
...
         )
         self.state_index = StateIndex(init_buffers)
+        self.zero_frac_index = StateIndex(jnp.array(1.0, dtype=dtype))
         self.inference = inference
         self.axis_name = axis_name
         self.input_size = input_size
         self.eps = eps
         self.channelwise_affine = channelwise_affine
         self.momentum = momentum
+        self.warmup_period = max(1, warmup_period)
```
patrick-kidger, line 131: Why the `max`? Perhaps it would be better to just error out on values that are too small?

equinox/nn/_batch_norm.py
```diff
@@ -86,26 +98,37 @@ def __init__(
...
 
     @jax.named_scope("eqx.nn.BatchNorm")
     def __call__(
```
patrick-kidger, line 134: It's not completely obvious to me that the `ema` implementation, with default arguments, reproduces the previous behaviour. (For example, we have `warmup_period=1000` by default?)

Can you add some comments explaining what each `approach` corresponds to?

equinox/nn/_batch_norm.py
```diff
@@ -154,23 +180,35 @@ def _stats(y):
                 var = jnp.maximum(0.0, var)
                 return mean, var
 
-            first_time = state.get(self.first_time_index)
-            state = state.set(self.first_time_index, jnp.array(False))
+            momentum = self.momentum
+            zero_frac = state.get(self.zero_frac_index)
+            zero_frac *= momentum
```
patrick-kidger, line 185: Stylistic nit: I tend not to use the inplace operations in JAX code. This (a) fits with the functional style a bit better, and (b) emphasises that we're definitely falling back to the `zero_frac = zero_frac * momentum` interpretation of the syntax. (Gosh, why does Python has two different meanings for the same syntax?)

equinox/nn/_batch_norm.py
```diff
@@ -154,23 +180,35 @@ def _stats(y):
...
+            state = state.set(self.zero_frac_index, zero_frac)
 
             batch_mean, batch_var = jax.vmap(_stats)(x)
             running_mean, running_var = state.get(self.state_index)
-            momentum = self.momentum
             running_mean = (1 - momentum) * batch_mean + momentum * running_mean
             running_var = (1 - momentum) * batch_var + momentum * running_var
```
patrick-kidger, line 191: These don't appear to be used on the `batch` branch. I think the lines here can be reorganised to keep each approach only using the things it needs.

equinox/nn/_batch_norm.py
```diff
@@ -154,23 +180,35 @@ def _stats(y):
...
+            if self.approach == "ema":
+                warmup_count = state.get(self.count_index)
+                warmup_count = jnp.minimum(warmup_count + 1, self.warmup_period)
+                state = state.set(self.count_index, warmup_count)
+
+                warmup_frac = warmup_count / self.warmup_period
+                norm_mean = zero_frac * batch_mean + running_mean
+                norm_mean = (1.0 - warmup_frac) * batch_mean + warmup_frac * norm_mean
+                norm_var = zero_frac * batch_var + running_var
+                norm_var = (1.0 - warmup_frac) * batch_var + warmup_frac * norm_var
```
patrick-kidger, line 203: I'm definitely going to have to sit down and grok what's going on here more carefully! As above it would be good to have some comments / docstrings / references / etc. describing what each approach is meant to do.

(C.f. something like the `MultiheadAttention` docstring for an example on how to use LaTeX if it'd be helpful.)

equinox/nn/_batch_norm.py
andrewdipper: warmup_period=0 seemed natural for off - Changed to just check and error out

equinox/nn/_batch_norm.py
andrewdipper: these are used by the batch branch when we're in inference mode so they still need to be computed and stored

equinox/nn/_batch_norm.py
andrewdipper: makes sense, done

equinox/nn/_batch_norm.py
andrewdipper: ema with warmup_period=1 approximately reproduces previous behavior.  As I noted the start is different because of how the running statistics are initially populated.  With warmup_period=1 there's no interpolation between the batch and running stats - the running stats are always used as with the previous behavior.  I can give an exact replication with an extra approach if necessary.  

Added some to the documentation

equinox/nn/_batch_norm.py
andrewdipper: Added some commentary and tried making it a bit cleaner.

But overall batch mode should follow the cited paper. Ema follows the prior behavior but changes the initialization of the running stats and adds interpolation so it can be stable while training.

equinox/nn/_batch_norm.py
```diff
@@ -202,8 +259,15 @@ def _stats(y):
                 norm_var = zero_frac * batch_var + running_var
                 norm_var = (1.0 - warmup_frac) * batch_var + warmup_frac * norm_var
             else:
+                axis_size = jax.lax.psum(jnp.array(1.0), self.axis_name)
+                debias_coef = (axis_size) / jnp.maximum(axis_size - 1, self.eps)
+                running_var = (
+                    1 - momentum
+                ) * debias_coef * batch_var + momentum * running_var
```
andrewdipper, line 271: I neglected to use unbiased variance so corrected that here

equinox/nn/_batch_norm.py
```diff
@@ -202,8 +259,15 @@ def _stats(y):
                 norm_var = zero_frac * batch_var + running_var
                 norm_var = (1.0 - warmup_frac) * batch_var + warmup_frac * norm_var
             else:
+                axis_size = jax.lax.psum(jnp.array(1.0), self.axis_name)
```
andrewdipper, line 267: I'm using this to get the length of the "batch" axis - but not sure it's the best / correct way

equinox/nn/_batch_norm.py
patrick-kidger: I think an exact replication is probably important for the default behaviour, just because I'd like to be sure that we're bit-for-bit backward compatible.

equinox/nn/_batch_norm.py
patrick-kidger: I think this is the correct way! IIRC `psum(1)` is actually special-cased for this purpose.

equinox/nn/_batch_norm.py
```diff
@@ -71,11 +132,17 @@ def __init__(
         - `axis_name`: The name of the batch axis to compute statistics over, as passed
             to `axis_name` in `jax.vmap` or `jax.pmap`. Can also be a sequence (e.g. a
             tuple or a list) of names, to compute statistics over multiple named axes.
+        - `approach`: The approach to use for the running statistics. If `approach=None`
+            a warning will be raised and approach will default to `"batch"`. During
+            training `"batch"` only uses batch statisics while`"ema"` uses the running
+            statistics.
```
patrick-kidger, line 138: So continuing from my previous comment -- probably the default should be `ema` if `approach=None`.

patrick-kidger (commented):
Okay! Sorry for taking so long to getting back around to reviewing this.

Lmk once you're happy that the previous behaviour is replicated by default, and I'll sit down with a pen and paper and satisfy myself that the calculations all look reasonable!

equinox/nn/_batch_norm.py
andrewdipper: Makes sense, it was different enough that I added it as `"ema_compatibility"`.  I changed the warning to rather strongly recommend against using `"ema_compatibility"`.  I haven't found a use case where I wouldn't expect to see the instability (at least with a larger learning rate) but that could very much be due to a lack of imagination on my part..  That part can definitely change if needed

andrewdipper:
All good - I got caught up in other things myself!

From my tests the replication is exact now.  It added another approach that is very similar to `"ema"` but it seemed like the most reasonable way to organize it.  Let me know if anything isn't clear

lockwo:
What's the status of this PR? I'm encountering some equinox batch norm related trouble as I switch over some code from haiku, and I think it's related to some of the zero debiased issues also tackled in this PR

patrick-kidger:
Ah, the refactor here got large enough that I never found time to check that it was (a) still doing the same thing for backward compatibility; (b) was doing things that I thought were correct going forward.

If you're interested -- and using such functionality in practice (I pretty much never use BatchNorm, which is part of why it's hard for me to really confirm this is doing the right thing) -- then I'd welcome your input here!

lockwo:
Makes sense, I'll see about taking a stab at a small-ish change to allow for an option to align it with other packages. Tbh I haven't really followed the lore of batch norm ever since the whole internal covariate shift illusion debacle. I did follow the whole drama in 2022 around the post-batch norm models and ViT vs CNN with stuff like https://arxiv.org/abs/2102.06171 and https://arxiv.org/abs/2201.03545 (although idk what is SotA these days), but for this specific area I'm just working on some AlphaZero based stuff now and it has batch norm. 

Mostly I just want test the operations are the same between packages (or at least, there is an option to be the same) rather than the leaning too heavily on the theory (since even if flax and haiku are doing things wrong in a different way most probably just use it, and it basically becomes the expected way).

PR #680, opened 2024-03-11
Simple `dtype` argument addition

AakashKumarNain:
Simple addition of the dtype argument along with a very simple test (will increase the code coverage later).

A few thoughts:

1. If we don't make dtype an attribute, then we won't be able to do certain "nice to have" things in the `__call__`. For example, what if the user passed a `float32` array to a layer initialized with `float16`? Ideally we should warn the user to avoid silent bugs
2. I am not sure what should be the right annotation for the dtype argument

### DIFF
--- a/equinox/nn/_linear.py
+++ b/equinox/nn/_linear.py
@@ -23,6 +23,7 @@ class Linear(Module, strict=True):
         in_features: Union[int, Literal["scalar"]],
         out_features: Union[int, Literal["scalar"]],
         use_bias: bool = True,
+        dtype=jnp.float32,
         *,
         key: PRNGKeyArray,
     ):
@@ -47,10 +48,12 @@ class Linear(Module, strict=True):
         out_features_ = 1 if out_features == "scalar" else out_features
         lim = 1 / math.sqrt(in_features_)
         self.weight = jrandom.uniform(
-            wkey, (out_features_, in_features_), minval=-lim, maxval=lim
+            wkey, (out_features_, in_features_), minval=-lim, maxval=lim, dtype=dtype
         )
         if use_bias:
-            self.bias = jrandom.uniform(bkey, (out_features_,), minval=-lim, maxval=lim)
+            self.bias = jrandom.uniform(
+                bkey, (out_features_,), minval=-lim, maxval=lim, dtype=dtype
+            )
         else:
             self.bias = None
 
--- a/equinox/nn/_linear.py
+++ b/equinox/nn/_linear.py
@@ -6,6 +6,7 @@ import jax.numpy as jnp
 import jax.random as jrandom
 from jaxtyping import Array, PRNGKeyArray
 
+from .._misc import default_floating_dtype
 from .._module import field, Module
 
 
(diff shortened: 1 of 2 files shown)


equinox/nn/_linear.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         in_features: Union[int, Literal["scalar"]],
         out_features: Union[int, Literal["scalar"]],
         use_bias: bool = True,
+        dtype=jnp.float32,
```
ArturGalstyan1, line 26: Shouldn't the default dtype be dependent on whatever this is set to:

`jax.config.x64_enabled`

?

Artur-Galstyan, line 26: Shouldn't the default dtype be dependent on whatever this is set to:

`jax.config.x64_enabled`

?

AakashKumarNain: We can do that but I wonder if that will work in every case, especially with caveats like this one `XLA doesn’t support 64-bit convolutions on all backends!`

equinox/nn/_linear.py
```diff
@@ -23,6 +24,7 @@ def __init__(
         in_features: Union[int, Literal["scalar"]],
         out_features: Union[int, Literal["scalar"]],
         use_bias: bool = True,
+        dtype=None,
```
Artur-Galstyan, line 27: The dtype of dtype is missing here. I don't have my laptop for a couple of days but I think the dtype of dtype is jax.numpy.dtype. And it should also be something like "dtype: Optional[jax.numpy.dtype] = None"

equinox/nn/_linear.py
```diff
@@ -33,6 +35,8 @@ def __init__(
         - `out_features`: The output size. The output from the layer will be a vector
             of shape `(out_features,)`.
         - `use_bias`: Whether to add on a bias as well.
+        - `dtype`: The dtype to use. Defaults to either `jax.numpy.float32` or
```
Artur-Galstyan, line 38: This should also tell the user that it applies to all trainable parameters (not just the weights but also the bias) IMO

Artur-Galstyan (changes requested):

patrick-kidger:
I think I like this! I wouldn't worry about the type annotation, so the only thing I'd tweak is the docstring a little bit.

I don't think we need to make dtype an attribute / I wouldn't try to raise a warning in `__call__`. That opens up too great a can of worms, I think, as in general there are all kinds of cases where folks might want mixed dtypes.

equinox/nn/_linear.py
AakashKumarNain: I have updated this

AakashKumarNain:
> I think I like this! I wouldn't worry about the type annotation, so the only thing I'd tweak is the docstring a little bit.

I have updated the docstring to reflect the effect of this argument in a better way.

>I don't think we need to make dtype an attribute / I wouldn't try to raise a warning in __call__. That opens up too great a can of worms, I think, as in general there are all kinds of cases where folks might want mixed dtypes.

SGTM

patrick-kidger:
Alright, I think I'm happy with this pattern! Since we're settled on that, do I understand correctly that you're now aiming to repeat this across other parts of `eqx.nn`?

AakashKumarNain:
Yes, I will send a separate PR for that. You can merge this one so that I am also unblocked for my Mistral based experiments.

patrick-kidger:
Okay, great!
Just merged into `dev`.

AakashKumarNain:
Thank you for the review and the help

Issue #681, opened 2024-03-11
Not using `filter_jit` produces `None` during backward pass


neel04:
So I've had this problem a couple times and finally decided to make an issue of it 😝 

Effectively, its just:
```diff
- @eqx.filter_jit
+ @partial(jax.jit, static_argnums=(6, 8))
```
```py
def make_step(model: eqx.Module, # replicated across devices
              x: Array, y: Array, pad_mask: Array, n: Array, k: Array, # Arrays
              optim: Callable, # static
              opt_state: Tuple[PyTree], # cannot be static marked static as tuple
              num_classes: int, # static
              keys: List[PRNGKeyArray]):
    
    loss, grads = compute_loss(model, x, y, pad_mask, n, k, num_classes, keys)
    updates, opt_state = optim.update(grads, opt_state, model)
    
    model = eqx.apply_updates(model, updates)
    
    return loss, model, opt_state
```
Which in theory should work as well as `filter_jit`. But actually yields error:

```py
  File "/Users/neel/miniconda3/envs/react_jax/lib/python3.10/site-packages/optax/_src/transform.py", line 83, in <lambda>
    lambda g, t: (1 - decay) * (g ** order) + decay * t, updates, moments)
TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
```
In the `optax.optim.update` step. 

More digging, its apparently the `Dropout` layer in the `PyTree`. More specifically, `p` takes `None` presumably after the `eqx.filter` step to obtain the `opt_state`. During the [`tree_map`](https://github.com/google-deepmind/optax/blob/95af789cb7f1d0279f44e143e3207e2e7b4d70d1/optax/_src/transform.py#L80) for the update, eventually moments (`t` becomes `None`).

This problem is always in `Dropout` for some reason, which is weird because its possible the simplest layer.

From the `moments` PyTree:
```py
dropout=Dropout(p=None, inference=None),
```

Whereas the same place in the `updates` PyTree:
```py
dropout=Dropout(p=f32[], inference=None),
```

**The solution** here turned out to use `is_array_like` instead of `is_array` during the `init`.

```py
opt_state = optim.init(eqx.filter(model, eqx.is_array_like))
```

---

I'm curious to understand - why's that `filter_jit` does make it work wo/ `is_array_like`? What's the interplay between `Dropout` and `is_array_like`?

[...]

PR #684, opened 2024-03-13
Wrap dropout probability into an array

knyazer:
Hey, Patrick!

I was working a bit with attention recently, and I discovered that the obvious implementation of a training loop with ```scan``` and ```optax``` fails. The culprit is the probability of the dropout: the value is ```float```, so it is not included in the model filtered with ```eqx.filter(model, eqx.is_array)``` since ```float``` is not an array. So, when initializing optax optimizer using this filtered model, the value is for ```p``` is None. Later, if we try to apply gradients, everything fails, because the gradient, on the other hand, is ```0.0``` and not None. And optax cannot comprehend this. 

I thought about three main solutions:
1. Force the gradient of p to be None. After some experiments with custom jvp rules it seems to be impossible, since the output shapes of the gradient function should be different: either static None or a jnp.float; And using just jnp.nan instead of None does not work.
2. Add new ```eqx.field(differentiate=True/False)``` flag, and new ```is_differentiable``` filter, that would together allow to specify whether field is forced to be differentiable or forced to be non-differentiable. This sounds like quite a bit of work, for such a simple bug-fix. But it might be useful for people who write libraries: like, users of the 3rd party library would love to be able to figure out dynamic and static parameters using the same ```eqx.is_differentiable```, which might have sensible defaults (e.g. similar to ```is_array```).
3. Just wrap the p in Dropout into jax array if it is not a jax array. This might break some compatibility... But I am not sure how badly is it in reality.

I implemented the 3rd approach + added a test with a very simple training pipeline for a multihead attention. WDYT?

### DIFF
--- a/equinox/nn/_dropout.py
+++ b/equinox/nn/_dropout.py
@@ -6,7 +6,7 @@ import jax.lax as lax
 import jax.numpy as jnp
 import jax.random as jrandom
 import numpy as np
-from jaxtyping import Array, Float, PRNGKeyArray
+from jaxtyping import Array, ArrayLike, Float, PRNGKeyArray
 
 from .._errors import error_if
 from .._filters import is_array
@@ -23,14 +23,14 @@ class Dropout(Module, strict=True):
     """
 
     # Not static fields as it makes sense to modify them via equinox.tree_at
-    p: Float[np.ndarray, ""] = field(
-        converter=lambda x: x if is_array(x) else np.array(x)
+    p: Float[ArrayLike, ""] = field(
+        converter=lambda x: x if is_array(x) else np.array(x, dtype=np.float32)
     )
     inference: bool
 
     def __init__(
         self,
-        p: float = 0.5,
+        p: ArrayLike = 0.5,
         inference: bool = False,
         *,
         deterministic: Optional[bool] = None,
@@ -51,7 +51,7 @@ class Dropout(Module, strict=True):
                 "Dropout(deterministic=...) is deprecated "
                 "in favour of Dropout(inference=...)"
             )
-        self.p = p  # type: ignore
+        self.p = p
         self.inference = inference
 
     # Backward compatibility
--- a/equinox/nn/_dropout.py
+++ b/equinox/nn/_dropout.py
@@ -5,9 +5,10 @@ import jax
 import jax.lax as lax
 import jax.numpy as jnp
 import jax.random as jrandom
-from jaxtyping import Array, PRNGKeyArray
+from jaxtyping import Array, Float, PRNGKeyArray
 
-from .._module import Module
+from .._filters import is_array
+from .._module import field, Module
 
 
 class Dropout(Module, strict=True):
@@ -20,7 +21,7 @@ class Dropout(Module, strict=True):
     """
 
     # Not static fields as it makes sense to want to modify them via equinox.tree_at.
-    p: float
+    p: Float[Array, ""] = field(converter=lambda x: x if is_array(x) else jnp.array(x))
     inference: bool
 
     def __init__(
@@ -46,7 +47,7 @@ class Dropout(Module, strict=True):
                 "Dropout(deterministic=...) is deprecated "
                 "in favour of Dropout(inference=...)"
             )
-        self.p = p
+        self.p = p  # type: ignore
         self.inference = inference
 
     # Backward compatibility
--- a/equinox/nn/_dropout.py
+++ b/equinox/nn/_dropout.py
@@ -84,8 +84,6 @@ class Dropout(Module, strict=True):
 
         if inference is None:
             inference = self.inference
-        if isinstance(self.p, (int, float)) and self.p == 0:
-            inference = True
         if inference:
             return x
         elif key is None:
--- a/equinox/nn/_dropout.py
+++ b/equinox/nn/_dropout.py
@@ -5,8 +5,10 @@ import jax
 import jax.lax as lax
 import jax.numpy as jnp
 import jax.random as jrandom
+import numpy as np
 from jaxtyping import Array, Float, PRNGKeyArray
 
+from .._errors import error_if
 from .._filters import is_array
 from .._module import field, Module
 
@@ -20,8 +22,10 @@ class Dropout(Module, strict=True):
     [`equinox.nn.inference_mode`][].
     """
 
-    # Not static fields as it makes sense to want to modify them via equinox.tree_at.
-    p: Float[Array, ""] = field(converter=lambda x: x if is_array(x) else jnp.array(x))
+    # Not static fields as it makes sense to modify them via equinox.tree_at
+    p: Float[np.ndarray, ""] = field(
+        converter=lambda x: x if is_array(x) else np.array(x)
+    )
     inference: bool
 
     def __init__(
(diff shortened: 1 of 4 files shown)


knyazer:
Also, found a little inconsistency in the dropout api too: one could run dropout in non-inference mode without a key. This is possible because of the special-case condition, that automatically switches to inference mode when probability of dropout is zero. However, I believe this is an inconsistent behaviour, so I just made it so that we always request the key when not under inference.

knyazer:
Though now, when I think about it, I am not actually sure what is more important: to have a 'simpler' behaviour, or to keep a consistent behaviour. 

In this case keeping consistent behaviour will be a bit annoying, since we will have to use ```eqx.error_if```, but it is still doable. I just wonder WDYT? Because the old behaviour of allowing to not pass the key if the rate is zero seems like a bad decision.. So fixing it would be good?

patrick-kidger:
So I believe the issue you describe is essentially that of #681? I.e., this is something that only occurs when using a non-filtered transformation?

Unfortunately I don't think there's really anything we can do to change this now.
- Not accepting a key is important, as `Dropout` as used as sublayer within `MultiheadAttention`, where it is frequently used with `p=0` and no key; we wouldn't want to demand passing in a key to every attention layer.
- Keeping `p` as a non-array is important, so that we can entirely remove dropout at compile-time, in the common case when `p=0`.

FWIW this does speak to an idea I've had noodling around in my head for a bit: namely, what if `eqx.Module` autodetected which fields are static vs nonstatic (based on `eqx.is_array`) during `__init__` time? This would make life difficult for anyone trying to tree-map to get a non-array (e.g. `eqx.nn.tree_inference` or `eqx.tree_at`), but maybe there's a way to finesse that problem. (Fundamentally it's a backward compatibility issue.) The end result would be that Equinox modules are always going to be compatible with all JAX transforms out of the box, no filtering required.

knyazer:
Yeah, I can see your point. Sorry for not researching issues well enough before making a PR :) But.... do you think that "removing dropout at runtime" is important enough? While performance is indeed important, is it important enough that we will make (some) people struggle with debugging training of anything attention-related? Not working the same way as any other equinox layers breaks the assumption that I can just hot-swap layers and be happy about myself.

I reckon that one of the main reasons behind torch success is that it cares more about logical syntax and less about being the fastest framework out there. And also, how bad could that be? I mean, dropout is a fast operation, I don't think including it at runtime will lead to more than a few percent performance drop: especially considering that dropout is linear in complexity (wrt layer size), while feedforward is quadratic, and attention is about ^2.5, but this all also does not matter because of the tensor cores. I think this is more of the tradeoff between compute time and human time: like how much time one would spend debugging it over the time one wins by having dropout statically removed during training. Like, let's say I spent two hours debugging this issue (yeah, I am slow), then how much money one would need to spend on training an LLM with Equinox so that the compute time that you win will be worth more than my salary for two hours :)

I am fine with either decision (keeping stuff the way it is/making ```p``` dynamic), but it seems that the approach of making ```p``` fully dynamic is a bit better from the viewpoint of the average library user.


About the idea of modules being dynamic: sounds fun! I also cannot see how to implement that neatly though. Another question is where to store the static part of the module: an obvious decision is to have ```dynamic``` and ```static``` properties, that store the pytree of the dynamic part and py-object of the static part, but then we still cannot pass raw modules into the pure jax transforms. I will try to think about something meaningful, and if I come up with anything that might work, I will tell you! Though I would guess that you are much more likely to solve this first :)

patrick-kidger:
Hmmm. How about this for a middle group: allow `p` to be `ArrayLike` (and thus either dynamic or static), but have the default, both here and in `MultiheadAttention`, be an array? That way it should continue to "just work" in the simple case. (In practice btw one can't use JAX operations at the top level for any library used in Google, so to avoid breaking all of my old coworkers, the appropriate thing to do is probably to use a NumPy array rather than a JAX array for the default value.)

One thing I don't see how we handle is the `inference` field, though.

As for my design idea: yeah, I agree there are some points there that aren't super obvious to me either.

knyazer:
Yep, using numpy is a good suggestion. I guess it solves most of likely backwards compatibility issues :+1: 

I didn't really get the part about
> not using jax operations on the top level
but I anyways implemented the numpy variant, since it looks a bit better than the jax.Array one anyway.

So, WDYT?

Again, if you feel like there could be some weird issues, and just don't want to merge this, I am alright with your decision; it seems like not that many people stumbled upon issues with dropout, and after I figured out the culprit, I just needed to go from 
```eqx.filter(..., eqx.is_array)```
 to 
```eqx.filter(..., lambda p: eqx.is_array(p) or (type(p) is float))```
 for it to work, so not a big deal! 

---

P.S. Good luck at the new job! I hope you will still support the JAX ecosystem though, without you it will die quite quickly :)

Another fun fact, my official job title is "Head of Magic"! Geniuses think alike, right? :laughing:

**Comment by patrick-kidger** [comment]
So I think this PR can be even simpler! I think we just need two changes:
- for the type annotation to be `p: ArrayLike`
- for the defaults to be `Dropout(p=np.array(0.5, dtype=np.foat32))` and `MultiheadAttention(p=np.array(0.0, dtype=np.float32))`
?

As for the new job, thank you! Absolutely I hope to keep this alive. Honestly most of the work I do on this is navigating new ever-more-obscure feature requests, and actual bugs seem to happen very infrequently. So fortunately my workload isn't too high :)

**Comment by knyazer** [comment]
If I remove the conversion from arbitrary type to numpy in the ```eqx.field``` of ```p```, then some of the tests I just added, break: there are some issues with static/dynamic values again. 

I think the test I added covers all (?) functionality that was supported before, and some bonuses,e.g. dynamically setting dropout probability: that was not supported (since the ```p``` was  static), and also I am still allowed to initialize anything with scalar values, since they will be automatically converted to a numpy array.

I don't really see a way to avoid automatic conversion in the ```field```, and other than that the PR is very simple. 

WDYT? Are you aware of a way to avoid using the complicated ```field``` without sacrificing flexibility? 

---
As a side note, there does not seem to be many tests that involve dropout, so it's hard to tell if I am breaking any functionality. Maybe we should write some property-based (~fuzz) testing of all the _important_ modules? Just shove lots of random values through the feedforward/norms/attention/etc, and check that the output is valid for all of them? 

This does not sound thaaat useful, since most of Equinox is effectively a wrapper over JAX/XLA, but e.g. dropout is sufficiently condition-heavy that it might be meaningful, but I am not sure. Just would be nice to have more guarantees on the new changes not breaking some old stable API.

File: equinox/nn/_dropout.py [hunk 1]
```diff
@@ -19,13 +22,15 @@ class Dropout(Module, strict=True):
     [`equinox.nn.inference_mode`][].
     """
 
-    # Not static fields as it makes sense to want to modify them via equinox.tree_at.
-    p: float
+    # Not static fields as it makes sense to modify them via equinox.tree_at
+    p: Float[ArrayLike, ""] = field(
+        converter=lambda x: x if is_array(x) else np.array(x, dtype=np.float32)
+    )
```
patrick-kidger, line 28: I think we can move the `converter` into `__init__`?

equinox/nn/_dropout.py
```diff
@@ -83,15 +88,24 @@ def __call__(
...
         else:
-            q = 1 - lax.stop_gradient(self.p)
+            p = self.p
+            if key is None:
+                p = error_if(
+                    p,
+                    p != 0,
+                    "Dropout requires a key when running in non-"
+                    "deterministic mode with non-zero probability.",
+                )
```
patrick-kidger, line 102: Ack, I hadn't really thought about the fact that we're using using `error_if` here. This is actually pretty slow + sometimes buggy on TPUs etc. It's very much a tool of last resort: as such I try to reserve its usage for the scientific ecosystem stuff (Diffrax etc.) and up till now it has been entirely unused in `eqx.nn`.

I'm not sure what our possibilities are here -- can this be avoided? I'm tempted to say that the status quo is better than using `error_if`.

knyazer:
As per discussion in #681:

I agree with everything you said; the only thing that I thought of is maybe it is possible to add a field parameter like ```eqx.field(static=True, add_to_pytree=True)```. When I get time to work on this stuff I will open a superceding PR with the implementation of your proposal. Probably will happen on Monday-ish.

Or you can work on it yourself, but I guess exploiting my free labor is better :)



---

As per this PR, since using eqx.error_if is a problem, then I don't really see a nice way to solve the issue in this way. Though, now I think it might be nice to have a "stable" error_if, maybe via error lifting? E.g. when the error is detected, we pass the error to the higher context until the interpreter level is reached, that just raises the error normally. I might try to experiment and produce something meaningful, but that's out of the scope for now.

PR #687, opened 2024-03-20
add missing parameter to update

RaulPL:
Add missing parameter (the model) to the update function in stateful.ipynb.


patrick-kidger:
Hey there!
I've just double-checked and the example seems to run for me. (At least with my current set of library versions.)
Is this failing for you when you try to run this?

RaulPL:
My bad, I just found out that the params parameter in the update function it is not necessary. Will close this PR, thanks for answering :)

PR #689, opened 2024-03-23
Add dtypes to the rest of eqx.nn

Artur-Galstyan:
Basically the same as #680 but for the rest of the API.

Double check `SpectralNorm` - I'm not 100% confident in that one.

### DIFF
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -10,6 +10,7 @@ import jax.numpy as jnp
 import jax.random as jrandom
 from jaxtyping import Array, Bool, Float, PRNGKeyArray
 
+from .._misc import default_floating_dtype
 from .._module import field, Module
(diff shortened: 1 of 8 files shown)


patrick-kidger (commented):
On `SpectralNorm`, I think the appropriate thing to do is probably to use `weight.dtype` instead.

By the way, I think `BatchNorm` could also do with updating: it has a `dtype` argument but so far only uses these for the moving statistics, but not its learnt parameters. We should probably change that. Would you be able to do that as part of this PR?

Other than that, thank you for doing this! This is all great to see. :D

Artur-Galstyan:
Yup, done.

patrick-kidger:
Neat! LGTM, and thank you for the contribution. We'll have this appear in the next version of Equinox.

PR #690, opened 2024-03-23
Add autoregressiveness to MHA (again)

Artur-Galstyan:
Hi, I noticed that the autoregressive-attention branch has become stale and a few things have changed in the dev branch since then. So I thought I'd revive the auto regressive branch.

Note that this does not yet have the MQA <-> MHA interpolation. In our last discussion, I told you about my "double VMAP" strategy and that LLMs such as LLaMA don't use that but instead simply "repeat" the keys and values until they match the queries. You were not sure which approach to follow.

I actually have a working version of the MHA implementation in which I allow the user to choose which one they want (see [this](https://github.com/Artur-Galstyan/kira/blob/main/kira/modules/mha.py#L136). Maybe in the future we could have something similar, but for now, I think it's ok to add autoregressive decoding - even without the MHA<->MQA interpolation. 

WDYT?

### DIFF
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -401,22 +401,9 @@ class MultiheadAttention(Module, strict=True):
                     "`MultiheadAttention(..., state_length=...)`."
                 )
             key_state, value_state, index = state.get(self.autoregressive_index)
-            key_seq_length = key_heads.shape[0]
-            if index + key_seq_length > self.state_length:
-                key_shift = index + key_seq_length - self.state_length
-                key_shift = min(key_shift, key_seq_length)
-                key_state = jnp.roll(key_state, -key_shift, axis=0)
-
             key_state = lax.dynamic_update_slice_in_dim(
                 key_state, key_heads, index, axis=0
             )
-
-            value_seq_length = value_heads.shape[0]
-            if index + value_seq_length > self.state_length:
-                value_shift = index + value_seq_length - self.state_length
-                value_shift = min(value_shift, value_seq_length)
-                value_state = jnp.roll(value_state, -value_shift, axis=0)
-
             value_state = lax.dynamic_update_slice_in_dim(
                 value_state, value_heads, index, axis=0
             )
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -6,16 +6,15 @@ from functools import partial
 from typing import cast, Literal, Optional, overload, Union
 
 import jax
-import jax.lax as lax
 import jax.numpy as jnp
 import jax.random as jrandom
 from jaxtyping import Array, Bool, Float, PRNGKeyArray
 
-from .._misc import default_int_dtype
 from .._module import field, Module
 from ._dropout import Dropout
+from ._kv_cache import AbstractKVCache
 from ._linear import Linear
-from ._stateful import State, StateIndex
+from ._stateful import State
 
 
 def dot_product_attention_weights(
@@ -125,7 +124,7 @@ class MultiheadAttention(Module, strict=True):
     value_proj: Linear
     output_proj: Linear
     dropout: Dropout
-    autoregressive_index: Optional[StateIndex]
+    kv_cache: Optional[AbstractKVCache]
 
     num_heads: int = field(static=True)
     query_size: int = field(static=True)
@@ -156,6 +155,7 @@ class MultiheadAttention(Module, strict=True):
         use_output_bias: bool = False,
         dropout_p: float = 0.0,
         inference: bool = False,
+        kv_cache: Optional[AbstractKVCache] = None,
         *,
         key: PRNGKeyArray,
     ):
@@ -224,18 +224,7 @@ class MultiheadAttention(Module, strict=True):
         self.use_value_bias = use_value_bias
         self.use_output_bias = use_output_bias
         self.state_length = state_length
-
-        def _make_autoregressive_cache(**_):
-            key_shape = state_length, num_heads, qk_size
-            value_shape = state_length, num_heads, vo_size
-            _int = default_int_dtype()
-            return jnp.empty(key_shape), jnp.empty(value_shape), jnp.zeros((), _int)
-
-        self.autoregressive_index = (
-            StateIndex(_make_autoregressive_cache())
-            if state_length is not None
-            else None
-        )
+        self.kv_cache = kv_cache
 
     @overload
     def __call__(
@@ -394,24 +383,18 @@ class MultiheadAttention(Module, strict=True):
         if state is None:
             causal_mask_offset = 0
         else:
-            if self.autoregressive_index is None:
+            if self.kv_cache is None:
                 raise ValueError(
                     "State was provided, but cannot use autoregressive decoding "
                     "without specifying "
-                    "`MultiheadAttention(..., state_length=...)`."
+                    "`MultiheadAttention(..., kv_cache=...)`. "
+                    "See `equinox.nn.StandardKVCache` for an example."
                 )
-            key_state, value_state, index = state.get(self.autoregressive_index)
-            key_state = lax.dynamic_update_slice_in_dim(
-                key_state, key_heads, index, axis=0
+
+            key_state, value_state, index, state = self.kv_cache(
+                state, key_heads, value_heads
             )
-            value_state = lax.dynamic_update_slice_in_dim(
-                value_state, value_heads, index, axis=0
-            )
-            causal_mask_offset = index
-            index = index + kv_seq_length
-            state = state.set(
-                self.autoregressive_index, (key_state, value_state, index)
-            )
+            causal_mask_offset = index - kv_seq_length
             key_heads = key_state
             value_heads = value_state
             kv_seq_length = self.state_length
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -3,16 +3,19 @@ import math
 import warnings
 from collections.abc import Callable
 from functools import partial
-from typing import cast, Optional, Union
+from typing import cast, Literal, Optional, overload, Union
 
 import jax
+import jax.lax as lax
 import jax.numpy as jnp
 import jax.random as jrandom
 from jaxtyping import Array, Bool, Float, PRNGKeyArray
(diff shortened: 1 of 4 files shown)


equinox/nn/_attention.py
```diff
@@ -215,6 +237,60 @@ def __init__(
         self.use_key_bias = use_key_bias
         self.use_value_bias = use_value_bias
         self.use_output_bias = use_output_bias
+        self.autoregressive_index = StateIndex(_make_autoregressive_cache())
```
patrick-kidger, line 240: If `state_length is None` then I think we'll hit the error message here -- so I think this will unconditionally raise?

equinox/nn/_attention.py
```diff
@@ -323,10 +438,53 @@ def __call__(
...
+        else:
+            return out, state
 
     def _project(self, proj, x):
         seq_length, _ = x.shape
         projection = jax.vmap(proj)(x)
         return projection.reshape(seq_length, self.num_heads, -1)
+
+
+def self_attention(
```
patrick-kidger, line 454: I think this function was most useful back when we had MQA, and for which the `MultiheadAttention` API was a bit too much.

I don't think we're *quite* there yet (although it's definitely still pretty complicated), so I think I'd probably suggest removing this function for now.

patrick-kidger (commented):
Thank you for this! This would be awesome to get in.

Would you be interested in writing some tests? Frankly the implementation of this is fairly complicated. I think I mostly trust past-me when I wrote this... but that not *that* much!

equinox/nn/_attention.py
```diff
@@ -308,6 +390,39 @@ def __call__(
...
+        if state is sentinel:
+            causal_mask_offset = 0
+        else:
+            key_state, value_state, index = state.get(self.autoregressive_index)
+            key_state = lax.dynamic_update_slice_in_dim(
+                key_state, key_heads, index, axis=0
+            )
+            value_state = lax.dynamic_update_slice_in_dim(
+                value_state, value_heads, index, axis=0
+            )
```
patrick-kidger, line 402: By the way, I think there is actually a subtle bug in these two lines: when `index` is close to the end, then it gets pushed back so that all of the inserted array is inserted (rather than simply truncating the inserted array to only insert the bits that fit).

I don't know if that's something that can be resolved efficiently, though.

Artur-Galstyan: Ah I see. Well, the solution is probably to truncate `key|value_state` to account for that before passing them into the `dynamic_update_slice_in_dim` function. I'll see what I can do here.

Artur-Galstyan: Alternatively, we might want to employ the FiFo strategy and roll the key and value state arrays to make room for the new entries.

Artur-Galstyan: I've added rolling; but I'll need to train something to see if it still works fine this way. I'll keep you posted.

equinox/nn/_attention.py
```diff
@@ -308,6 +391,58 @@ def __call__(
...
+        else:
+            if self.autoregressive_index is None:
+                raise ValueError(
+                    "State was provided, but cannot use autoregressive decoding "
+                    "without specifying "
+                    "`MultiheadAttention(..., state_length=...)`."
+                )
+            key_state, value_state, index = state.get(self.autoregressive_index)
+            key_seq_length = key_heads.shape[0]
+            if index + key_seq_length > self.state_length:
```
patrick-kidger, line 405: These `if` statements will fail under JIT.

FWIW I suspect it won't be possible to make things work efficiently at the end anyway, and it's not obvious to me that rolling is really the better choice. (E.g. some techniques make sure to use the start of an attention window as a place to put globally-attended-to information.) I think I'm probably happy with the behaviour without this.

Artur-Galstyan: I see. I reverted it back to the previous implementation.

AFAICT in the current implementation of the cache, if the cache is:

`1 2 3 4 5 6 7 0 0`

and the `seq_len` is 3, then after the update it becomes:

`1 2 3 4 5 6 A B C` 

But if I understood you correctly (from your previous comment), you want it to be 

`1 2 3 4 5 6 7 A B` 

right? In the rolling strategy, it would have become:

`2 3 4 5 6 7 A B C`.

 Perhaps David could weigh in on this, because I don't know what the best approach is. @dlwh WDYT?

patrick-kidger: A very neat explanation! Yes, that's it exactly.

dlwh: I've stayed away from inference so far, so I mostly haven't had to think about caching. That seems sensible and clean.

It does seem like a thing this could be pluggable? I know it's built on State right now, but if the KVCache were a module with an abstract method, you could choose the behavior.

The MaxText code is pretty hard to follow (for me) but I think there are two caches there, one that does the prefilling/prefix that Patrick talks about, and another that's for the ar decoding. [this](https://github.com/google/maxtext/blob/1ae752103af6f2939fbb3f6bbbfa998eee8e4d76/MaxText/layers/attentions.py#L653) means it's doing a roll for AR tokens and leaving the prefill tokens in place. I think?

Artur-Galstyan: > It does seem like a thing this could be pluggable? I know it's built on State right now, but if the KVCache were a module with an abstract method, you could choose the behavior.

I really like this idea! I can cook something up. It would also make the already pretty large MHA file slightly more svelte.

Artur-Galstyan:
I've added a first draft for an `AbstractKVCache`. The idea is that people can define their own KVCache (if they wish to) and pass that to the MHA constructor. The abstract class calls the abstract method in `__call__` and does some checks around it to make sure that it was implemented correctly. From there, the MHA implementation just takes the new KV matrices from the KV cache and works with those. 

I also provided a "StandardKVCache" which is essentially what we had before (without truncation). I was thinking of adding another one which does rolling and then let the user choose which one they want -- or let them create their own.

LMK what you think of this approach.

patrick-kidger:
I didn't close this intentionally -- deleting the target branch (in this case, `dev`), causes all PRs against it to be automatically closed. Feel free to reopen this against `main`!

equinox/nn/_attention.py
```diff
@@ -144,12 +148,14 @@ def __init__(
...
         qk_size: Optional[int] = None,
         vo_size: Optional[int] = None,
+        state_length: Optional[int] = None,
         use_query_bias: bool = False,
         use_key_bias: bool = False,
         use_value_bias: bool = False,
         use_output_bias: bool = False,
         dropout_p: float = 0.0,
         inference: bool = False,
+        kv_cache: Optional[AbstractKVCache] = None,
```
patrick-kidger, line 158: This argument doesn't seem to be documented in the docstring.

equinox/nn/_attention.py
```diff
@@ -308,6 +380,39 @@ def __call__(
...
+                    "See `equinox.nn.StandardKVCache` for an example."
+                )
+
+            key_state, value_state, index, state = self.kv_cache(
+                state, key_heads, value_heads
+            )
+            causal_mask_offset = index - kv_seq_length
+            key_heads = key_state
+            value_heads = value_state
+            kv_seq_length = self.state_length
```
patrick-kidger, line 400: I like this! It helps to make the `MultiheadAttention.__call__` code a little less of a gigantic monolith.

equinox/nn/_attention.py
```diff
@@ -256,6 +325,9 @@ def __call__(
             be a JAX array of shape `(query_seq_length, kv_seq_length)`, or (for custom
```
patrick-kidger, line 325: The `mask="causal"` option needs documenting.

I suspect this might end up being quite long. Maybe we should pull it out into an admonition (the `!!! foo` things you see in the docs). As a reference point, se how we handle the discussion on padding for `eqx.nn.Conv`.

equinox/nn/_attention.py
patrick-kidger: Also, I think it would be really useful to have a short example somewhere -- maybe at the bottom of the `__init__` docstring? Not a strong feeling on location -- on how to use `MultiheadAttention` with state and kv caching.

patrick-kidger: Also I think `state_length` is unused here now? It appears solely in the `kv_cache`. We should remove it as an argument.

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,103 @@
...
+import jax.lax as lax
+import jax.numpy as jnp
+from jaxtyping import Array, Float, Int
+
+from .._misc import default_int_dtype
+from .._module import field, Module
+from ._stateful import State, StateIndex
+
+
+class AbstractKVCache(Module, strict=True):
```
patrick-kidger, line 12: So I suspect this doesn't need to actually be a full-blown abstract class. It looks like all the caller needs is a `__call__` with the appropriate API.

I think this means that we could have
```python
class MultiheadAttention(eqx.Module:
    def __init__(..., kv_cache: Optional[Callable[...signature here...]] = None):

class StandardKVCache(eqx.Module): ...
```
with the latter just collapsing the existing implementation of `AbstractKVCache` and `StandardKVCache` into just one class.

For bonus points this would mean we achieve consistency with the API for `process_heads`, which also just uses a `Callable` rather than an ABC. (In fact I don't think we really have ABCs anywhere else in `eqx.nn` -- neural networks seem to be simply enough that they're not needed.)

patrick-kidger: As a side note, some comment on the design as you currently have it (just for your own interest):
- `AbstractKVCache` is defining an `autoregressive_index` attribute, but this isn't actually used anywhere that `AbstractKVCache` is accepted. Rather, it is only used inside the `StandardKVCache` subclass. So this should be moved into the subclass.
- You might [this doc](https://docs.kidger.site/equinox/pattern/) interesting! It's the pattern I'm trying to use in Equinox and its ecosystem going forward. In particualr it advocates against `super`.

patrick-kidger (commented):
Okay, I really like this as an idea! This had not occured to me at all, but I think it makes the code much simpler to read.

I've left some comments on how we might neated the structure some more, but let's definitely do this.

Issue #698, opened 2024-04-09
Broken link in documentation


lockwo:
Just noticed this link https://docs.kidger.site/examples/parallelism/ is broken in the documentation, I can put a fix in for it later 
<img width="809" alt="Screenshot 2024-04-09 at 3 58 34 PM" src="https://github.com/patrick-kidger/equinox/assets/42878312/243c7b67-a4ce-4f5e-b204-638f322de2c5">

PR #701, opened 2024-04-10
fix link

lockwo:
addresses: https://github.com/patrick-kidger/equinox/issues/698


patrick-kidger:
LGTM! Thank you for the fix.

PR #704, opened 2024-04-15
MHA KV Cache

Artur-Galstyan:
Hi Patrick, I've incorporated the changes. 

It's not 100% done yet, but I'd especially like you to double check the error handling regarding the shapes in the MHA file.

### DIFF
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -380,6 +380,7 @@ class MultiheadAttention(Module, strict=True):
                 )
         if state is None:
             state_length = None
+            index = None
             causal_mask_offset = 0
         else:
             if self.kv_cache is None:
@@ -400,20 +401,9 @@ class MultiheadAttention(Module, strict=True):
             value_heads = value_state
             kv_seq_length = state_length
 
-        if mask == "causal":
-            query_indices = jnp.arange(query_seq_length)[:, None]
-            kv_indices = jnp.arange(kv_seq_length)[None, :]
-            mask = kv_indices <= query_indices + causal_mask_offset
+        mask = _generate_mask(mask, query_seq_length, kv_seq_length, causal_mask_offset)
         if state_length is not None:
-            # Also mask out the latter parts of the state we haven't written into yet.
-            unwritten_mask = jnp.arange(state_length) < index  # pyright: ignore
-            if mask is None:
-                mask = jnp.broadcast_to(
-                    unwritten_mask, (query_seq_length, state_length)
-                )
-            else:
-                mask = mask & unwritten_mask
-
+            mask = _mask_unwritten_parts(state_length, query_seq_length, mask, index)
         attn_fn = partial(
             dot_product_attention, dropout=self.dropout, inference=inference
         )
@@ -471,3 +461,40 @@ def _check_kv_shapes(
             f"Expected {(value_heads_num_heads, value_heads_vo_size)}"
             f"got {(value_num_heads, value_vo_size)},"
         )
+
+
+def _generate_mask(
+    mask: Union[
+        None,
+        Bool[Array, "q_seq kv_seq"],
+        Bool[Array, "num_heads q_seq kv_seq"],
+        Literal["causal"],
+    ],
+    query_seq_length: int,
+    kv_seq_length: int,
+    causal_mask_offset: Array | Literal[0],
+) -> Optional[Array]:
+    if mask == "causal":
+        query_indices = jnp.arange(query_seq_length)[:, None]
+        kv_indices = jnp.arange(kv_seq_length)[None, :]
+        return kv_indices <= query_indices + causal_mask_offset
+    else:
+        return mask
+
+
+def _mask_unwritten_parts(
+    state_length: int,
+    query_seq_length: int,
+    mask: Union[
+        Optional[Bool[Array, "q_seq kv_seq"]],
+        Optional[Bool[Array, "num_heads q_seq kv_seq"]],
+    ],
+    index: Optional[Array],
+):
+    # Also mask out the latter parts of the state we haven't written into yet.
+    unwritten_mask = jnp.arange(state_length) < index  # pyright: ignore
+    if mask is None:
+        mask = jnp.broadcast_to(unwritten_mask, (query_seq_length, state_length))
+    else:
+        mask = mask & unwritten_mask.reshape(*mask.shape)
+    return mask
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -389,10 +389,11 @@ class MultiheadAttention(Module, strict=True):
                     "`MultiheadAttention(..., kv_cache=...)`. "
                     "See `equinox.nn.StandardKVCache` for an example."
                 )
-
             key_state, value_state, index, state = self.kv_cache(
                 key_heads, value_heads, state
             )
+            print("key_state", key_state, key_state.shape)
+            print("value_state", value_state, value_state.shape)
             _check_kv_shapes(key_state, value_state, key_heads, value_heads)
             state_length, _, _ = key_state.shape
 
@@ -462,7 +463,7 @@ def _check_kv_shapes(
     if (key_num_heads, key_qk_size) != (key_heads_num_heads, key_heads_qk_size):
         raise ValueError(
             "key_state has different `num_heads` or `qk_size` than key_heads\n"
-            f"Expected {(key_heads_num_heads, key_heads_qk_size)}"
+            f"Expected {(key_heads_num_heads, key_heads_qk_size)} "
             f"got {(key_num_heads, key_qk_size)}, "
         )
 
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -193,15 +193,18 @@ class MultiheadAttention(Module, strict=True):
         !!! info
 
         The following shows an example of how to use `MultiheadAttention` for
-        autoregressive decoding with a `kv_cache` and `state`:
+        autoregressive decoding with a `kv_cache` and `state`
+        (see [`equinox.nn.StandardKVCache`][]):
         ```python
         import equinox as eqx
         import jax
 
-        query_size = 6
-        num_heads = 1
-        state_length = 8
+
         seq_len = 3
+
+        state_length = 8
+        num_heads = 1
+        query_size = 6
 
         standard_kv_cache = eqx.nn.StandardKVCache(
             key_shape=(state_length, num_heads, query_size),
@@ -212,12 +215,12 @@ class MultiheadAttention(Module, strict=True):
             query_size=query_size,
             num_heads=num_heads,
             kv_cache=standard_kv_cache,
-            key=jax.random.PRNGKey(0)
+            key=jax.random.key(0)
 
         )
 
         for states in range(state_length):
-            x = jax.numpy.ones(shape=(seq_len, query_size)) * (states + 1)
+            x = jax.numpy.ones(shape=(seq_len, query_size))
             y, state = mha(x, x, x, mask="causal", state=state)
 
         ```
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -8,14 +8,14 @@ from typing import cast, Literal, Optional, Union
 import jax
 import jax.numpy as jnp
 import jax.random as jrandom
-from jaxtyping import Array, Bool, Float, PRNGKeyArray
+from jaxtyping import Array, Bool, Float, Int, PRNGKeyArray
 
-from .._misc import default_floating_dtype
+from .._misc import default_floating_dtype, default_int_dtype
 from .._module import field, Module
 from ._dropout import Dropout
 from ._kv_cache import KVCacheCallable
 from ._linear import Linear
-from ._stateful import State
+from ._stateful import State, StateIndex
 
 
 def dot_product_attention_weights(
@@ -127,6 +127,7 @@ class MultiheadAttention(Module, strict=True):
     dropout: Dropout
 
     kv_cache: Optional[KVCacheCallable]
+    index: Optional[StateIndex]
 
     num_heads: int = field(static=True)
     query_size: int = field(static=True)
@@ -277,6 +278,12 @@ class MultiheadAttention(Module, strict=True):
         self.use_value_bias = use_value_bias
         self.use_output_bias = use_output_bias
         self.kv_cache = kv_cache
+        if self.kv_cache is not None:
+            self.index = StateIndex(
+                jnp.zeros((), default_int_dtype()),
+            )
+        else:
+            self.index = None
 
     @jax.named_scope("eqx.nn.MultiheadAttention")
     def __call__(
@@ -301,6 +308,7 @@ class MultiheadAttention(Module, strict=True):
                     Float[Array, "seq_length num_heads qk_size"],
                     Float[Array, "seq_length num_heads qk_size"],
                     Float[Array, "seq_length num_heads vo_size"],
+                    Int[Array, ""],
                 ],
                 tuple[
                     Float[Array, "seq_length num_heads qk_size"],
@@ -332,16 +340,16 @@ class MultiheadAttention(Module, strict=True):
         - `inference`: As [`equinox.nn.Dropout.__call__`][]. (Keyword only
             argument.)
         - `deterministic`: (Deprecated in favour of `inference`.)
-        - `process_heads`: A function that takes in the query, key, and value heads and
-            returns new query, key, and value heads. For example, this can be
-            used to implement relative positional embeddings -
-            see e.g. `RotaryPositionalEmbedding`for an example. (Keyword only argument.)
+        - `process_heads`: A function that takes in the query, key, value heads as well
+            as the current autoregressive index and returns new query, key, and value
+            heads. For example, this can be used to implement relative positional
+            embeddings - see e.g. `RotaryPositionalEmbedding`for an example.
+            (Keyword only argument.)
 
         **Returns:**
 
         A JAX array of shape `(query_seq_length, output_size)`.
         """
-
         if deterministic is not None:
             inference = deterministic
             warnings.warn(
@@ -349,67 +357,27 @@ class MultiheadAttention(Module, strict=True):
                 "in favour of MultiheadAttention()(inference=...)"
             )
 
-        query_seq_length, _ = query.shape
-        kv_seq_length, _ = key_.shape
-        kv_seq_length2, _ = value.shape
+        query_seq_length, kv_seq_length = self._get_query_and_kv_seq_lengths(
+            query, key_, value
+        )
 
-        if kv_seq_length != kv_seq_length2:
-            raise ValueError("key and value must both be sequences of equal length.")
-
-        query_heads = self._project(self.query_proj, query)
-        key_heads = self._project(self.key_proj, key_)
-        value_heads = self._project(self.value_proj, value)
-
-        # TODO: Apply RoPE somehow, somewhere here
+        query_heads, key_heads, value_heads = self._project_heads(query, key_, value)
+        index = self._get_start_index(state)
 
         if process_heads is not None:
-            q_shape, k_shape, v_shape = (
-                query_heads.shape,
-                key_heads.shape,
-                value_heads.shape,
-            )
-            query_heads, key_heads, value_heads = process_heads(
-                query_heads,
-                key_heads,
-                value_heads,
+            query_heads, key_heads, value_heads = _process_heads(
+                process_heads, query_heads, key_heads, value_heads, index
             )
 
-            if (
-                query_heads.shape != q_shape
-                or key_heads.shape != k_shape
-                or value_heads.shape != v_shape
-            ):
-                raise ValueError(
-                    "process_heads must not change the shape of the heads."
-                )
-
-        if state is None:
-            state_length = None
-            index = None
-            causal_mask_offset = 0
-        else:
-            if self.kv_cache is None:
-                raise ValueError(
-                    "State was provided, but cannot use autoregressive decoding "
-                    "without specifying "
-                    "`MultiheadAttention(..., kv_cache=...)`. "
-                    "See `equinox.nn.StandardKVCache` for an example."
-                )
-
-            key_state, value_state, index, state = self.kv_cache(
-                key_heads, value_heads, state
+        if state:
+            key_heads, value_heads, kv_seq_length, state = self._handle_kv_cache(
+                key_heads, value_heads, index, query_seq_length, state
             )
-            _check_kv_shapes(key_state, value_state, key_heads, value_heads)
-            state_length, _, _ = key_state.shape
 
-            causal_mask_offset = index
-            key_heads = key_state
-            value_heads = value_state
-            kv_seq_length = state_length
+        mask = _generate_mask(mask, query_seq_length, kv_seq_length, index)
+        if self.kv_cache is not None:
+            mask = _mask_unwritten_parts(kv_seq_length, query_seq_length, mask, index)
 
-        mask = _generate_mask(mask, query_seq_length, kv_seq_length, causal_mask_offset)
-        if state_length is not None:
-            mask = _mask_unwritten_parts(state_length, query_seq_length, mask, index)
         attn_fn = partial(
             dot_product_attention, dropout=self.dropout, inference=inference
         )
@@ -431,6 +399,49 @@ class MultiheadAttention(Module, strict=True):
             return out
         else:
             return out, state
+
+    def _get_query_and_kv_seq_lengths(self, query, key_, value):
+        query_seq_length, _ = query.shape
+        kv_seq_length, _ = key_.shape
+        kv_seq_length2, _ = value.shape
+
+        if kv_seq_length != kv_seq_length2:
+            raise ValueError("key and value must both be sequences of equal length.")
+
+        return query_seq_length, kv_seq_length
+
+    def _handle_kv_cache(self, key_heads, value_heads, index, query_seq_length, state):
+        if self.kv_cache is None:
+            raise ValueError(
+                "State was provided, but cannot use autoregressive decoding "
+                "without specifying "
+                "`MultiheadAttention(..., kv_cache=...)`. "
+                "See `equinox.nn.StandardKVCache` for an example."
+            )
+
+        key_state, value_state, state = self.kv_cache(
+            key_heads, value_heads, index, state
+        )
+        _check_kv_shapes(key_state, value_state, key_heads, value_heads)
+        kv_seq_length, _, _ = key_state.shape
+
+        assert self.index is not None
+        state = state.set(self.index, index + query_seq_length)
+        return key_state, value_state, kv_seq_length, state
+
+    def _get_start_index(self, state):
+        if state is not None and self.index is not None:
+            index = state.get(self.index)
+        else:
+            index = jnp.array(0, dtype=default_int_dtype())
+
+        return index
+
+    def _project_heads(self, query, key_, value):
+        query_heads = self._project(self.query_proj, query)
+        key_heads = self._project(self.key_proj, key_)
+        value_heads = self._project(self.value_proj, value)
+        return query_heads, key_heads, value_heads
 
     def _project(self, proj, x):
         seq_length, _ = x.shape
@@ -489,7 +500,7 @@ def _generate_mask(
 
 
 def _mask_unwritten_parts(
-    state_length: int,
+    kv_seq_length: int,
     query_seq_length: int,
     mask: Union[
         Optional[Bool[Array, "q_seq kv_seq"]],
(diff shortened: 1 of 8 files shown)


Artur-Galstyan:
I also updated the ruff version from 0.1.x to the latest 0.4.4 version. I also ran the pre-commit hooks over the repo with the new ruff version, which lead to the changes. 

On second thought, this might not have been the best idea as it's now 2 distinct things:
1) the MHA KV Cache
2) updating ruff

LMK if this is a problem, then I'll revert the other changes coming from updating ruff.

patrick-kidger:
Ah, wonderful! Yeah, separate PRs would be appreciated I think.
(And gosh, I'm aware that reviewing this one seems to have sat on my backlog for a while, I promise to try and get to this... )

Artur-Galstyan:
No worries, I created a separate PR #727 and reverted the changes here.

equinox/nn/_attention.py
```diff
@@ -152,6 +169,21 @@ def __init__(
...
+                    Float[Array, "seq_length num_heads vo_size"],
+                ],
+                tuple[
+                    Float[Array, "state_length num_heads qk_size"],
+                    Float[Array, "state_length num_heads vo_size"],
+                    Int[Array, ""],
+                    State,
+                ],
+            ]
+        ] = None,
```
patrick-kidger, line 186: Oh gosh, this is probably going to make our documentation unreadable.

I don't love it, but maybe we should switch this out with just `Optional[Callable]` and then explain that it should match the same API as `StandardKVCache`?

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
         - `dtype`: The dtype to use for all trainable parameters in this layer.
             Defaults to either `jax.numpy.float32` or `jax.numpy.float64` depending
             on whether JAX is in 64-bit mode.
+        - `kv_cache`: The callable module to handle KV caching.
         - `key`: A `jax.random.PRNGKey` used to provide randomness for parameter
             initialisation. (Keyword only argument.)
+
+
```
patrick-kidger, line 218: Nit: spurious newlines.

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+        The following shows an example of how to use `MultiheadAttention` for
+        autoregressive decoding with a `kv_cache` and `state`:
+        ```python
+        import equinox as eqx
+        import jax
+
+        query_size = 6
+        num_heads = 1
+        state_length = 8
+        seq_len = 3
```
patrick-kidger, line 231: Nit: can we order these in the same length that we use them?

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+        - `kv_cache`: The callable module to handle KV caching.
         - `key`: A `jax.random.PRNGKey` used to provide randomness for parameter
             initialisation. (Keyword only argument.)
+
+
+
+        !!! info
+
+        The following shows an example of how to use `MultiheadAttention` for
+        autoregressive decoding with a `kv_cache` and `state`:
```
patrick-kidger, line 223: Can you add a link to `StandardKVCache` here? You can do this with specifically the syntax ``[`equinox.nn.StandardKVCache`][]``.

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+        standard_kv_cache = eqx.nn.StandardKVCache(
+            key_shape=(state_length, num_heads, query_size),
+            value_shape=(state_length, num_heads, query_size),
+        )
+
+        mha, state = eqx.nn.make_with_state(MultiheadAttention)(
+            query_size=query_size,
+            num_heads=num_heads,
+            kv_cache=standard_kv_cache,
+            key=jax.random.PRNGKey(0)
```
patrick-kidger, line 242: JAX now has `jax.random.key` that we should probably be using instead. I still need to update the rest of the documentation, but we might as well do the correct thing where we can.

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+
+        )
+
+        for states in range(state_length):
+            x = jax.numpy.ones(shape=(seq_len, query_size)) * (states + 1)
```
patrick-kidger, line 247: I think the `* (states + 1)` can be removed? Just to help make clear that it's not necessary.

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+            y, state = mha(x, x, x, mask="causal", state=state)
```
patrick-kidger, line 248: Ah, this is a beautiful API. Very happy to see this in!

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+
```
patrick-kidger, line 249: Nit: spurious newline.

equinox/nn/_attention.py
```diff
@@ -179,8 +211,44 @@ def __init__(
...
+        ```
+
```
patrick-kidger, line 251: (Another another spurious newline here too.)

Think of all those precious bytes wasted on GitHub... !

equinox/nn/_attention.py
```diff
@@ -297,10 +375,8 @@ def __call__(
             )
 
         query_seq_length, _ = query.shape
-        kv_seq_length, _ = key_.shape
-        kv_seq_length2, _ = value.shape
-        if kv_seq_length != kv_seq_length2:
-            # query length can be different
+
+        if (kv_seq_length := key_.shape[0]) != value.shape[0]:
```
patrick-kidger, line 379: Actually, one benefit of the previous approach was that it implicitly checked the number of dimensions in the input as well. I think that's important!

equinox/nn/_attention.py
```diff
@@ -325,6 +401,42 @@ def __call__(
...
+                raise ValueError(
+                    "State was provided, but cannot use autoregressive decoding "
+                    "without specifying "
+                    "`MultiheadAttention(..., kv_cache=...)`. "
+                    "See `equinox.nn.StandardKVCache` for an example."
+                )
+
+            key_state, value_state, index, state = self.kv_cache(
+                state, key_heads, value_heads
+            )
```
patrick-kidger, line 418: Nit: we have `state` as a first argument but as a last output.

The general convention has been to have it at the end -- could this be `self.kv_cache(key_heads, value_heads, state)` instead?

equinox/nn/_attention.py
```diff
@@ -325,6 +401,42 @@ def __call__(
...
+            _check_kv_shapes(key_state, value_state, key_heads, value_heads)
+            state_length = key_state.shape[0]
```
patrick-kidger, line 420: Likewise here I'd comment on prefering unpacking over indexing:
```python
state_length, _, _ = key_state.shape
```
as another implicit assertion that everything is as we think it is.

(This is a general thing I try to do -- always use unpacking and never use indexing. It's a style that has caught my mistakes many times!)

equinox/nn/_attention.py
```diff
@@ -341,10 +453,38 @@ def __call__(
...
         projection = jax.vmap(proj)(x)
         return projection.reshape(seq_length, self.num_heads, -1)
+
+
+def _check_kv_shapes(
+    key_state: Float[Array, "state_length num_heads qk_size"],
+    value_state: Float[Array, "state_length num_heads vo_size"],
+    key_heads: Float[Array, "seq_length num_heads qk_size"],
+    value_heads: Float[Array, "seq_length num_heads vo_size"],
+) -> None:
```
patrick-kidger, line 474: Also here, I'd either use unpacking or add an `assert key_state.ndim == 3` etc. as well.

equinox/nn/_attention.py
```diff
@@ -325,6 +401,42 @@ def __call__(
...
+                    "See `equinox.nn.StandardKVCache` for an example."
+                )
+
+            key_state, value_state, index, state = self.kv_cache(
+                state, key_heads, value_heads
+            )
+            _check_kv_shapes(key_state, value_state, key_heads, value_heads)
+            state_length = key_state.shape[0]
+
+            causal_mask_offset = index - kv_seq_length
```
patrick-kidger, line 422: I think we can skip this subtraction by removing the corresponding addition inside of `StandardKVCache`:

```python
        state = state.set(self.autoregressive_index, (key_state, value_state, index + kv_seq_length))
        return key_state, value_state, index, state
```
(Note that I've still done the addition when updating `state` though!)

equinox/nn/_attention.py
```diff
@@ -325,6 +401,42 @@ def __call__(
...
+            key_heads = key_state
+            value_heads = value_state
+            kv_seq_length = state_length
+
+        if mask == "causal":
+            query_indices = jnp.arange(query_seq_length)[:, None]
+            kv_indices = jnp.arange(kv_seq_length)[None, :]
+            mask = kv_indices <= query_indices + causal_mask_offset
```
patrick-kidger, line 430: I think it'd be really good to have a test for this. I remember writing this originally and *gosh*, I don't know about you but I do not trust the past version of Patrick!

equinox/nn/_attention.py
```diff
@@ -325,6 +401,42 @@ def __call__(
...
+        if state is not None and state_length is not None:
```
patrick-kidger, line 431: I think `state is not None` should imply `state_length is not None`, right?

If you just need to communicate this fact to a static type checker then this can be done using an assert:
```python
if state is not None:
    assert state_length is not None
    ...
```

equinox/nn/_attention.py
```diff
@@ -325,6 +401,42 @@ def __call__(
                 raise ValueError(
                     "process_heads must not change the shape of the heads."
                 )
+        state_length = None
```
patrick-kidger, line 404: By the way, prefer single-assignment-form here:
```python
if state is None:
    state_length = None
    causal_mask_offset = 0
else:
    ...
    state_length = ...
    ...
```
which is generally a more reliable coding style than overwriting a local variable.

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,50 @@
...
+class StandardKVCache(Module):
+    key_shape: tuple[int, int, int] = field(static=True)
+    value_shape: tuple[int, int, int] = field(static=True)
+
+    autoregressive_index: StateIndex
+
+    def __init__(
+        self,
+        key_shape: tuple[int, int, int],
+        value_shape: tuple[int, int, int],
```
patrick-kidger, line 19: Can we make the names passed in here match the names we also pass into `eqx.nn.MultiheadAttention`?

I also note that there's an implicit assumption that the first value of `key_shape` and `value_shape` are the same. Better to pull those out as a separate shared `state_length` argument I think.

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,50 @@
...
+    ):
+        self.key_shape = key_shape
+        self.value_shape = value_shape
+
+        def _make_cache(**_):
+            _int = default_int_dtype()
+            return jnp.empty(key_shape), jnp.empty(value_shape), jnp.zeros((), _int)
+
+        self.autoregressive_index = StateIndex(_make_cache())
```
patrick-kidger, line 28: This function is defined and then immediately called. (I recognise this as the old pattern we had for `StateIndex`!) Can we inline it instead?

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,50 @@
...
+        state: State,
+        key_heads: Float[Array, "seq_length num_heads qk_size"],
+        value_heads: Float[Array, "seq_length num_heads vo_size"],
+    ) -> tuple[
+        Float[Array, "state_length num_heads qk_size"],
+        Float[Array, "state_length num_heads vo_size"],
+        Int[Array, ""],
+        State,
+    ]:
+        kv_seq_length = key_heads.shape[0]
```
patrick-kidger, line 41: (likewise prefer unpacking over indexing)

tests/test_nn.py
```diff
@@ -770,7 +770,9 @@ def test_multihead_attention(getkey):
     q = jrandom.uniform(getkey(), (19, 3))
     k = jrandom.uniform(getkey(), (23, 5))
     v = jrandom.uniform(getkey(), (23, 7))
-    assert attn(q, k, v).shape == (19, 11)
+    out = attn(q, k, v)
+    assert isinstance(out, jnp.ndarray)
+    assert out.shape == (19, 11)
```
patrick-kidger, line 775: As above, it'd be great to have some tests for all this new functionality!

patrick-kidger (commented):
Gosh, once again it has taken me *forever* to give you a proper review. I promise I still haven't forgotten about this. It sits on my GitHub notifications reminding me...

Anyway, my only real comment here is "let's add tests". Everything else is a huge pile of nits -- I think I'm really happy with your design here!

Artur-Galstyan:
Aside from the tests (which I'll tackle this weekend) I think the rest should be solved. LMK what you think

equinox/nn/_attention.py
```diff
@@ -325,6 +377,42 @@ def __call__(
...
+                raise ValueError(
+                    "State was provided, but cannot use autoregressive decoding "
+                    "without specifying "
+                    "`MultiheadAttention(..., kv_cache=...)`. "
+                    "See `equinox.nn.StandardKVCache` for an example."
+                )
+
+            key_state, value_state, index, state = self.kv_cache(
+                key_heads, value_heads, state
+            )
```
patrick-kidger, line 394: I've just realised -- is this actually doing the correct thing if we also have a RoPE? I don't think it is.

Hmm, do we need to also pass a start index to the RoPE to make this work?

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,73 @@
...
+    value_shape: tuple[int, int, int] = field(static=True)
+
+    autoregressive_index: StateIndex
+
+    def __init__(
+        self,
+        state_length: int,
+        num_heads: int,
+        key_size: int,
+        value_size: int,
```
patrick-kidger, line 38: Oh, we should probably have a `dtype` argument here, since that seems to be a thing we're doing going forwards.

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,73 @@
...
+    ):
+        self.key_shape = state_length, num_heads, key_size
+        self.value_shape = state_length, num_heads, value_size
+
+        self.autoregressive_index = StateIndex(
+            (
+                lambda _: jnp.empty(self.key_shape),
```
patrick-kidger, line 45: I think this `lambda` is spurious?

equinox/nn/_kv_cache.py
```diff
@@ -0,0 +1,73 @@
...
+        Int[Array, ""],
+        State,
+    ]:
+        kv_seq_length, _, _ = key_heads.shape
+        key_state, value_state, index = state.get(self.autoregressive_index)
+        key_state = lax.dynamic_update_slice_in_dim(key_state, key_heads, index, axis=0)
+        value_state = lax.dynamic_update_slice_in_dim(
+            value_state, value_heads, index, axis=0
+        )
+        index = index + kv_seq_length
```
patrick-kidger, line 68: I think this line needs to be removed.

patrick-kidger (commented):
Great stuff!

equinox/nn/_attention.py
```diff
@@ -179,8 +184,45 @@ def __init__(
...
+            query_size=query_size,
+            num_heads=num_heads,
+            kv_cache=standard_kv_cache,
+            key=jax.random.key(0)
+
+        )
+
+        for states in range(state_length):
+            x = jax.numpy.ones(shape=(seq_len, query_size))
+            y, state = mha(x, x, x, mask="causal", state=state)
```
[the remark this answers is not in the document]
Artur-Galstyan: image
It looks fine, no?

equinox/nn/_attention.py
```diff
@@ -179,8 +184,45 @@ def __init__(
...
+
+        seq_len = 3
+
+        state_length = 8
+        num_heads = 1
+        query_size = 6
+
+        standard_kv_cache = eqx.nn.StandardKVCache(
+            key_shape=(state_length, num_heads, query_size),
+            value_shape=(state_length, num_heads, query_size),
```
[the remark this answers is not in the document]
Artur-Galstyan: Whoops. Fixed it.

equinox/nn/_attention.py
Artur-Galstyan: I don't quite follow. 

RoPE is applied _after_ the linear projections:

```python
        query_heads = self._project(self.query_proj, query)
        key_heads = self._project(self.key_proj, key_)
        value_heads = self._project(self.value_proj, value)

        if process_heads is not None:
            q_shape, k_shape, v_shape = (
                query_heads.shape,
                key_heads.shape,
                value_heads.shape,
            )
            query_heads, key_heads, value_heads = process_heads(
                query_heads, key_heads, value_heads
            )
```

And after RoPE is applied, the keys and values are put into the KV cache. This means that the "RoPE-applied" keys and values are stored/updated. 

Or, did you mean that we should:

1. Store/update the KV cache
2. Apply `process_heads` *after* fetching the KV from the cache:

```python
            if self.kv_cache is None:
                raise ValueError(
                    "State was provided, but cannot use autoregressive decoding "
                    "without specifying "
                    "`MultiheadAttention(..., kv_cache=...)`. "
                    "See `equinox.nn.StandardKVCache` for an example."
                )

            key_state, value_state, index, state = self.kv_cache(
                key_heads, value_heads, state
            )
            _check_kv_shapes(key_state, value_state, key_heads, value_heads)
            state_length, _, _ = key_state.shape

            causal_mask_offset = index
            key_heads = key_state
            value_heads = value_state
            kv_seq_length = state_length

            # PROCESS HEADS HERE INSTEAD
```

equinox/nn/_kv_cache.py
Artur-Galstyan: Ah, right.

equinox/nn/_kv_cache.py
Artur-Galstyan: Ah, yes.

tests/test_nn.py
Artur-Galstyan: I've just written a little test using the LLaMA implementation and our implementation and got the same results (for the KV cache) -- just need to finalise it

equinox/nn/_kv_cache.py
Artur-Galstyan: Yep, it was.

equinox/nn/_attention.py
Artur-Galstyan: I've checked out the code from LLaMA3 as a reference, and they apply RoPE before inserting the arrays into the KV cache.

Artur-Galstyan:
Hey Patrick, I've added a couple of tests regarding the functionalities. WDYT?

equinox/nn/_attention.py
patrick-kidger: Right, but RoPe uses a bunch of precomputed sines and cosines whose values oscillate according to how far along the sequence they are: https://github.com/patrick-kidger/equinox/blob/d3a01c8d1b5531f193f11514c0189104b8f62136/equinox/nn/_embedding.py#L186-L189

However when using KV caching, then we won't be adjusting that `arange` appropriately to account for how far along the sequence we are.

patrick-kidger:
On tests! I think both of these look worth including.

I think the main thing that still needs testing is self-consistency with respect to whether or not we do a single model evaluation, or a few model evaluations, or many multiple evaluations -- that the caching does the right thing all the way through. In particular also when using any combination of {RoPE / no RoPE} x {no/arbitrary/causal masking}. I think this would have caught the error I describve above, for example.

Other than that -- if you rebase on the latest main then the failing tests should now be fixed :)

Artur-Galstyan:
I think I finally understood the issue you were referring to! I'll do my best to elaborate it, but correct me if I misunderstood you.

Let's assume we have these `freqs_cis` computed, like so

```python
max_seq_len = 4
freqs_cis = precompute_freqs(start=0, end=max_seq_len, ...)
```

Let's say the frequencies are these (just for simplification)

```python
[0, 1, 2, 3]
```

When we apply RoPE like this, we get the issue:


```python
embedding_size = 4
x = jnp.ones(shape=(1, embedding_size))

# first token
out1 = rope(x)

# second token
out2 = rope(x)

# out1 is the same as out2, not what we want!
```

This is because we apply the same `freqs_cis` to both of them.

Given our `freqs_cis`

```python
[0, 1, 2, 3]
```

We use `0` in both cases, but in the second case, we should be using the `1`. 


This means we need some kind of index or something similar in the RoPE embedding. 

Is this what you meant?

patrick-kidger:
Yup -- that's exactly the issue I have in mind!

Artur-Galstyan:
Then there are 2 ways to solve this issue. Either:

1) We force the user to decide *upfront* the `max_seq_len` of the RoPE embeddings.
In that case, the API would look like this (from the user's perspective):

```python
rope = RotaryPositionalEmbedding(embedding_size, max_seq_len, theta)

x = jnp.ones(shape=(seq_len, embedding_size))

offset = jnp.array(0)
out = rope(x, offset)

offset = jnp.array(1)
out = rope(x, offset)
```

Behind the scenes, we would simply create one `freqcs_cis` array and use `jax.lax.dynamic_slice_[in_dim]` to get the correct slice given the offset. But I'm unsure if this will work well under JIT.

2) We don't force the user to decide on the `max_seq_len` but use the `offset` to continuously increase the length of the `freqs_cis` and then use slicing. Basically, we would not just use the input's `seq_len` to determine if we need a new `freqs_cis` but also add the `offset` to it . For example:

```
first input: seq_len: 1, offset: 0 

needed freqs_cis length = 1 + 0 = 1
freqs_cis = [0]
---
second input: seq_len: 1, offset: 1 

needed freqs_cis length = 1 + 1 = 2
freqs_cis = [0, 1]
actually used - sliced by offset - freqs_cis = [1]
---
third input: seq_len: 1, offset: 1 

needed freqs_cis length = 1 + 1 = 2
freqs_cis = [0, 1] (use the cached one)
actually used - sliced by offset - freqs_cis = [1]
---
fourth input: seq_len: 1, offset: 2 

needed freqs_cis length = 1 + 2 = 3
freqs_cis = [0, 1, 2]
actually used - sliced by offset - freqs_cis = [2]
```

The downside of the latter approach is that we might have to compute the `freqs_cis` many times. In the example above, we generate the entire `freqs_cis` each time, but we could also change the `arange` function to start at `offset` and then concatenate the `freqs_cis` arrays. In any case, the latter approach calls `precompute_freqs_cis` at most `max_seq_len` times. 

Personally, I like the first approach more, because we just have to compute the `freqs_cis` once and then just slice it down to what we need. 

Which one should I go for?

patrick-kidger:
Agreed on 1, I think. This is (a) more nicely aligned with the current KV strategy (which expects a `state_length`), and (b) I think the latter one might actually be impossible: we'd end up with the frequencies having a dependency on a traced value (the offset), so we wouldn't be able to put the result in a cache.

Artur-Galstyan:
One problem I'm facing is that it keeps - as I suspected - reJITting the function once the offset changes in the function arguments. Concretely:

```python
def rope(..., offset: int):
    ...
    
rope(x, offset=0) # JIT
rope(x, offset=1) # Another JIT
```

I can't set the offset to be an array, because then `jax.lax.[dynamic]_slice` complains that the start/end need to be `int`s and I can't use states, because the `int`s in the states become `jax._src.interpreters.partial_eval.DynamicJaxprTracer`, which also leads to an error :( 

You don't happen to have a good idea to tackle this kind of JIT problems? Ideally, I'd like to not reJIT the function each time, there must be a way not to do that, because we're actually dealing with static sized arrays (as long as the input `seq_len` doesn't change). This is because:

If `seq_len` is e.g. 2 and `max_seq_len` is 4 and the offset is `0`, then the `freqs_cis` shape is `(4,1)`. The desired `freqs_cis` shape is `(2, 1)` (sliced to match the `seq_len`). 

But if the offset is `1`, then the neither the original `freqs_cis` shape `(4, 1)` nor the desired `freqs_cis` shape `(2, 1)` changes. Since the shapes remain constants, there should be a way to avoid reJITting all the time. 

WDYT?

patrick-kidger:
So dynamic slicing should work using a tracer for the start index! https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.dynamic_slice.html#jax.lax.dynamic_slice

patrick-kidger:
Great stuff!
Let me know once this is in a place you're happy with and I'll take a look. Once this is in, then I'm contemplating doing a new release to celebrate this :D

Artur-Galstyan:
I added in some tests, which include testing MHA with the KV cache and RoPE. I also did a mini refactor of the MHA class, where I put chunks of code into a separate function to improve readability. 

I also removed the index from the KV cache and store it now in MHA directly instead. The reason for this is that in the previous iteration, we got the index **after** storing the KV vectors into the cache, but we needed the index before (to apply RoPE before inserting into the cache). It also doesn't make much sense semantically: what does the KV cache even have to do with some index -- all they should care about is storing the keys and values. 

RoPE now gets an offset parameter and I updated the docstring. 

LMK what you think

patrick-kidger:
Great! Can you rebase on top of the latest main, and I'd love to do a review.

(Do you know of a way to do a review based on just the changes from a branch? When I click on the 'files changed' tab I see a lot of changes that are completely unrelated to yours...)

Artur-Galstyan:
It's the unfortunate "side-effect" of using rebase (which I believe is the better way to merge branches) because it changes the commit hashes, making it look like I did those changes. 

Although I like rebasing more, I have to admit that I'm not the greatest Git guru out there and don't know of any **good** ways to resolve this. I could of course simply filter my changes into a separate PR (option 1), which makes it easier for you to review the changes (this is of course not the **right** solution, but it'll bring us one step forward). 

Personally, I usually just mark unrelated files as viewed and keep going (option 2). The problem in this case is that for some files you might not even know which of the changes came from me. The files that I changed are these ones:
- attention.md
- _misc.py
- _attention.py
- _embedding.py
- _kv_cache.py
- test_nn.py

patrick-kidger:
Ultimately this feels like a thing that should be solveable at the GitHub level. At least in principle one could cherry-pick all the commits from a branch and replay them on the latest `HEAD`, after all.

I think if you rebase on top of main and then force-push to this branch, that should suffice?

Artur-Galstyan:
I think, it should be correct now. I rebased on main and force pushed. I often get tripped up by this, that when rebasing one needs to force push when working with *origin*. LMK if it's alright now.

equinox/nn/_attention.py
```diff
@@ -179,8 +185,46 @@ def __init__(
...
+            query_size=query_size,
+            num_heads=num_heads,
+            kv_cache=standard_kv_cache,
+            key=jax.random.key(0)
+
+        )
+
+        for states in range(state_length):
+            x = jax.numpy.ones(shape=(seq_len, query_size))
+            y, state = mha(x, x, x, mask="causal", state=state)
```
patrick-kidger, line 226: This example does not run.

equinox/nn/_attention.py
```diff
@@ -272,59 +330,53 @@ def __call__(
...
 
-        query_seq_length, _ = query.shape
-        kv_seq_length, _ = key_.shape
-        kv_seq_length2, _ = value.shape
-        if kv_seq_length != kv_seq_length2:
-            # query length can be different
-            raise ValueError("key and value must both be sequences of equal length.")
+        query_seq_length, kv_seq_length = self._get_query_and_kv_seq_lengths(
+            query, key_, value
+        )
```
patrick-kidger, line 362: Can we revert this change? I find (a) find it easier to read without all the little functions, but (b) more importantly, it's substantially easier for me to review a smaller diff.

equinox/nn/_attention.py
```diff
@@ -272,59 +330,53 @@ def __call__(
...
 
-        query_heads = self._project(self.query_proj, query)
-        key_heads = self._project(self.key_proj, key_)
-        value_heads = self._project(self.value_proj, value)
+        query_heads, key_heads, value_heads = self._project_heads(query, key_, value)
```
patrick-kidger, line 364: Likewise here.

equinox/nn/_attention.py
```diff
@@ -272,59 +330,53 @@ def __call__(
...
 
-            if (
-                query_heads.shape != q_shape
-                or key_heads.shape != k_shape
-                or value_heads.shape != v_shape
-            ):
-                raise ValueError(
-                    "process_heads must not change the shape of the heads."
-                )
+        mask = _generate_mask(mask, query_seq_length, kv_seq_length, index)
```
patrick-kidger, line 377: and here!

patrick-kidger (commented):
I've just done a quick review of a few things that jumped out at me. I've also open a PR fixing a few other things over at https://github.com/Artur-Galstyan/equinox/pull/14.

That said, I'm starting to have my doubts about this PR.

First of all, I'm clearly not finding time to review it. Second, as the failing example code demonstrates, it seems pretty hard to verify correctness. (I think I'd want to be using this myself day-to-day to be sure that we're getting every detail right.) And third, this is rapidly becoming a very complicated layer to understand. I worry that we might be pushing the boundaries of usability here -- especially given all the new flavour-of-the-month kinds of attention that would necessitate writing from scratch anyway,

WDYT?

Artur-Galstyan:
I had a similar thought the other day, esp. given that [JAX has now a built-in attention function](https://jax.readthedocs.io/en/latest/_autosummary/jax.nn.dot_product_attention.html#jax.nn.dot_product_attention) and if you're doing research in LLMs, you will end up with your own MHA module anyway. This leaves us with the question of what we should do with this PR. I've given it some thought and this would be my suggestion:

First, close this PR -- entropy has increased too much in the MHA module and usefulness has declined (note that I'm not emotionally bound to the code I write or the effort it took - everything is a learning process, so I don't mind at all 😄). *But* there is still a bug in the RoPE embeddings class, which I would extract from this PR into a separate PR and merge it into main (that change should not be lost/forgotten, it's an actual bug). Then, we could create a separate PR and move this MHA + KV cache module to an example to showcase how you could extend the MHA implementation yourself to utilise a KV cache. So, to summarise:

1) Close this PR
2) PR for the RoPE bug
3) PR for MHA + KV cache as an example

patrick-kidger:
I think that sounds like a wonderful plan! I like that a lot.

Let's get the RoPE fix in, and then I'll do a new release of Equinox straight afterwards. I've been itching to release some other unrelated fixes for a while... 😁

Thank you for your understanding! I really appreciate that.

PR #708, opened 2024-04-18
Fix typo

nasyxx:
layernorm should be LayerNorm but Embedding


patrick-kidger:

Issue #715, opened 2024-05-07
Strange error with multiple tuples as attributes and `equinox.tree_at`?


leonard-gleyzer:
Hello!

I've come across an error I'm not quite sure what to do about.

MWE:

```python
import equinox as eqx


class M(eqx.Module):
    a: tuple
    b: tuple

    def __init__(self):
        self.a = ()
        self.b = ()


eqx.tree_at(lambda m: m.a, M(), ())
```

Running the above gives me the following error:

```
ValueError: `where` does not uniquely identify a single element of `pytree`. 
This usually occurs when trying to replace a `None` value:

  >>> eqx.tree_at(lambda t: t[0], (None, None, 1), True)


for which the fix is to specify that `None`s should be treated as leaves:

  >>> eqx.tree_at(lambda t: t[0], (None, None, 1), True,
  ...             is_leaf=lambda x: x is None) 
```

However, if I use lists, or a tuple and a list

```python
import equinox as eqx


class M(eqx.Module):
    a: tuple
    b: list

    def __init__(self):
        self.a = ()
        self.b = []


eqx.tree_at(lambda m: m.a, M(), ())
```

or just a single tuple

```python
import equinox as eqx


class M(eqx.Module):
    a: tuple

    def __init__(self):
        self.a = ()


eqx.tree_at(lambda m: m.a, M(), ())
```

there is no error.

patrick-kidger:
Interesting! Looks like Python actually uses the exact same object for tuple literals:
```python
() is ()  # True
(1,) is (1,)  # True
```

So the logic inside `eqx.tree_at` works by requiring that for any two nodes `x, y` in the PyTree, that `x is not y`. This is how it can identify what to replace. (Here I use "node" to refer to both composite nodes and leaves.)

Now specifically leaves frequently fail this (e.g. `x = [4, 4]; x[0] is x[1]  # True`), so we make this work by first doing a `pytree = jtu.tree_map(Wrapper, pytree)` to wrap up the leaves. No big deal.

And until this point, there haven't been any examples of composite nodes which fail the `x is not y` check. Now it seems we have one! This is why two tuples are specifically what is required to bump into this issue.

If anyone feels like tackling this in the near future then I'd be happy to take a pull request.

lockwo:
[...]

Issue #716, opened 2024-05-07
Jax 0.4.27: ValueError: safe_map() argument 2 is shorter than argument 1


GaetanLepage:
Since [jax 0.4.27](https://jax.readthedocs.io/en/latest/changelog.html#jax-0-4-27-may-7-2024), several tests fail with:
```
args = (_ClosureConvert(
  jaxpr={ lambda ; a:f32[47] b:f32[] c:i32[] d:bool[] e:bool[] f:i32[] g:f32[] h:f32[47]
    i:bool[...t 0x7fff4c435a90>,
  _makes_false_steps=False
), Traced<ShapedArray(float32[47])>with<DynamicJaxprTrace(level=5/1)>))))
kwds = {}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           ValueError: safe_map() argument 2 is shorter than argument 1
E           --------------------
E           For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

patrick-kidger:
Thanks for the report! I've reproduced this as an upstream bug in JAX: https://github.com/google/jax/issues/21116

GaetanLepage:
Thank you for the quick response !
The following tests also fail with a similar error message:
```
=========================== short test summary info ============================
FAILED tests/test_ad.py::test_closure_convert_basic - assert [] == [140730519276576]
FAILED tests/test_debug.py::test_backward_nan - AssertionError: assert 'foo:\n   pri...pe=float32)\n' == 'foo:\n   pri...pe...
FAILED tests/test_noinline.py::test_abstract - TypeError: emit_python_callback() takes 6 positional arguments but 7 positi...
FAILED tests/test_noinline.py::test_num_traces - TypeError: emit_python_callback() takes 6 positional arguments but 7 positi...
FAILED tests/test_noinline.py::test_pytree_in - TypeError: emit_python_callback() takes 6 positional arguments but 7 positi...
FAILED tests/test_noinline.py::test_simple - TypeError: emit_python_callback() takes 6 positional arguments but 7 positi...
FAILED tests/test_noinline.py::test_mlp - TypeError: emit_python_callback() takes 6 positional arguments but 7 positi...
FAILED tests/test_noinline.py::test_vmap - TypeError: emit_python_callback() takes 6 positional arguments but 7 positi...
[...]

PR #717, opened 2024-05-08
tuple fix

lockwo:
Fixes https://github.com/patrick-kidger/equinox/issues/715

### DIFF
--- a/equinox/_tree.py
+++ b/equinox/_tree.py
@@ -52,19 +52,7 @@ class _CountedIdDict:
 
 
 class _DistinctTuple:
-    def __init__(self, contents=()):
-        self.contents = tuple(contents)
-
-    def __eq__(self, other):
-        if isinstance(other, _DistinctTuple):
-            return self.contents == other.contents
-        return False
-
-    def __hash__(self):
-        return hash(self.contents)
-
-    def __repr__(self):
-        return "Distinct Tuple: " + repr(self.contents)
+    pass
 
 
 def tree_at(
@@ -153,9 +141,9 @@ def tree_at(
     # Whilst we're here: we also double-check that `where` is well-formed and doesn't
     # use leaf information. (As else `node_or_nodes` will be wrong.)
     pytree = jtu.tree_map(
-        lambda x: x if x != () else _DistinctTuple(),
+        lambda x: _DistinctTuple() if isinstance(x, tuple) and x == () else x,
         pytree,
-        is_leaf=lambda x: x == (),
+        is_leaf=lambda x: isinstance(x, tuple) and x == (),
     )
     node_or_nodes_nowrapper = where(pytree)
     pytree = jtu.tree_map(_LeafWrapper, pytree, is_leaf=is_leaf)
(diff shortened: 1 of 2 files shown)


lockwo:
The tests will fail, but there's some issues that are unrelated: https://github.com/patrick-kidger/equinox/issues/716

equinox/_tree.py
```diff
@@ -51,6 +51,22 @@ def count(self, item):
...
+    def __eq__(self, other):
+        if isinstance(other, _DistinctTuple):
+            return self.contents == other.contents
+        return False
+
+    def __hash__(self):
+        return hash(self.contents)
+
+    def __repr__(self):
+        return "Distinct Tuple: " + repr(self.contents)
```
patrick-kidger, line 67: I think a lot of this is unused -- just
```python
class _DistinctTuple:
    pass
```
should suffice, I think?

(In particular I don't think reprs are ever user-visible, and the `contents` are always empty.)

equinox/_tree.py
```diff
@@ -229,7 +250,11 @@ def _make_replacement(x: _Node) -> Any:
                 "  >>> eqx.tree_at(lambda t: t[0], (None, None, 1), True,\n"
                 "  ...             is_leaf=lambda x: x is None)"
             )
-
+    out = jtu.tree_map(
+        lambda x: x if not isinstance(x, _DistinctTuple) else (),
```
patrick-kidger, line 254: Nit: both here and above, can you remove the `not` and flip the branches?

tests/test_tree.py
```diff
@@ -31,6 +31,22 @@ def test_tree_at_replace(getkey):
...
+
+    assert pytree1[:-2] == true_pytree1[:-2]
+    assert pytree2[:-2] == true_pytree2[:-2]
+    assert jnp.all(pytree1[-2]["a"] == true_pytree1[-2]["a"])
+    assert jnp.all(pytree2[-2]["a"] == true_pytree2[-2]["a"])
+    assert jnp.all(pytree1[-1].weight == true_pytree1[-1].weight)
+    assert jnp.all(pytree1[-1].bias == true_pytree1[-1].bias)
+    assert jnp.all(pytree2[-1].weight == true_pytree2[-1].weight)
+    assert jnp.all(pytree2[-1].bias == true_pytree2[-1].bias)
+
```
patrick-kidger, line 49: I think let's pull this out as a separate test, it's edge case enough to want to highlight things. Something like this is probably what we want I think:
```python
a = ()
b= (1,)
x1 = [a]
x2 = [a, a]
x3 = [(), ()]
x4 = [b]
x5 = [b, b]
x6 = [(1,), (1,)]

for x in (x1, x2, x3, x4, x5, x6):
    new_x = eqx.tree_at(lambda xi: xi[0], x, "hello")
    assert new_x[0] == "hello"
    if len(new_x) != 1:
        assert new_x[1] != "hello"
```

WDYT?

patrick-kidger (commented):
Ah, awesome! Thank you for taking this on. I'm glad to see this doesn't look too tricky!

equinox/_tree.py
lockwo: That's true, it was a hold over from my print debugging

equinox/_tree.py
lockwo: done

tests/test_tree.py
lockwo: Sure, makes sense, pulled it out

patrick-kidger:
LGTM! Thank you for the fix here.

Indeed the latest JAX release broke some stuff, so in the meantime I'm going to merge this anyway.

PR #720, opened 2024-05-12
Added Mistral 7B example

patrick-kidger:
CC @AakashKumarNain !

(Very happy to include more details on this page if there's anything you think is better stated here rather than in your own README.)


AakashKumarNain:
Thanks @patrick-kidger I think this is good enough. Anyone interested in the details will click the link anyway, and will get the details from the README itself. Thanks agin

patrick-kidger:
Awesome! Merged.

PR #722, opened 2024-05-12
Fixed `eqx.tree_at` when used alongside an empty namedtuple.

patrick-kidger:
See #715 and #717. CC @lockwo.


lockwo:
Nice, good thought with the named tuple

PR #723, opened 2024-05-13
Fix rmsnorm

AakashKumarNain:
The norm calculations should always be done with at least `fp32` dtype, and then the result should be casted back to the original dtype

### DIFF
--- a/equinox/nn/_normalisation.py
+++ b/equinox/nn/_normalisation.py
@@ -382,10 +382,18 @@ class RMSNorm(Module, strict=True):
                 f"Received `shape={self.shape} and `x.shape={x.shape}`. You might need "
                 "to replace `rms_norm(x)` with `jax.vmap(rms_norm)(x)`.\n"
             )
-        x_dtype = x.dtype
-        upcasted_x = x.astype(jnp.float32)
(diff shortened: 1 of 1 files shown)


lockwo:
Won't this downcast if we are using float64?

AakashKumarNain:
I can fix that. Thanks for pointing it out

patrick-kidger:
I think this should be done using `jnp.result_type`, not with an explicit blocklist.

See how attention handles the same thing: https://github.com/patrick-kidger/equinox/blob/62304f5c5d1aaf0e45fcc9c27ba6a54f0e358b4e/equinox/nn/_attention.py#L36

AakashKumarNain:
> not with an explicit blocklist.

@patrick-kidger can you please elaborate a bit more on this?

patrick-kidger:
Right now you're explicitly checking against a list of dtypes. This is (a) not necessarily forward-compatible with any changes to the dtype system (consider a hypothetical `float128` being added), and (b) will already do the wrong thing for e.g. complex dtypes.

AakashKumarNain:
Makes sense. Thanks for the suggestion. Let me change that accordingly

patrick-kidger:
LGTM -- merged! Thank you for the fix.

PR #724, opened 2024-05-15
Avoid use of deprecated jax.tree_map

jakevdp:
`jax.tree_map` was deprecated in JAX v0.4.26; preferred replacements are `jax.tree.map` (JAX v0.4.25 or later) or `jax.tree_util.tree_map` (any JAX version).


patrick-kidger:
Cheers Jake -- LGTM! Seems like these snuck through.

PR #727, opened 2024-05-17
Updated ruff

Artur-Galstyan:
I've updated the pre-commit hook's ruff version to v.0.4.4.

This also meant that the top level "lint" section in the pyproject.toml had to change as well (I think this change came after v.0.2.0).


patrick-kidger:
LGTM! Thank you for the update (and for splitting this as a separate PR)!

PR #733, opened 2024-05-21
Update _rnn.py docs

Artur-Galstyan:
1. A typo: it should have been "self.cell"
2. A small clarification (in addition to the docstring) that the LSTM cell outputs a tuple - as opposed to the GRUCell, which outputs a single value


patrick-kidger:

PR #734, opened 2024-05-21
Fix jacfwd

lockwo:
I just found that jacfwd breaks if there are no static outputs from the Jacobian. E.g. it fails on the simple neural network on the home page. This fixes that (without the change, the new tests would fail, they would encounter an error because eqx.combine is call on a pytree and a None object)

### DIFF
--- a/equinox/_ad.py
+++ b/equinox/_ad.py
@@ -439,7 +439,10 @@ class _Jac(Module):
         else:
             jacobian = jax.jacfwd
         dynamic_out, (static_out, aux) = jacobian(_fun, has_aux=True)(diff_x)
-        out = combine(dynamic_out, static_out)
+        if static_out is None:
+            out = dynamic_out
+        else:
+            out = combine(dynamic_out, static_out)
         if self.has_aux:
             return out, aux
         else:
--- a/equinox/_ad.py
+++ b/equinox/_ad.py
@@ -429,20 +429,13 @@ class _Jac(Module):
                 _out, _aux = _out
             else:
                 _aux = None
-            _dynamic_out, _static_out = partition(
-                _out, lambda j: isinstance(j, ad.JVPTracer)
-            )
-            return _dynamic_out, (_static_out, _aux)
+            return _out, _aux
 
         if self.rev:
             jacobian = jax.jacrev
         else:
             jacobian = jax.jacfwd
-        dynamic_out, (static_out, aux) = jacobian(_fun, has_aux=True)(diff_x)
-        if static_out is None:
-            out = dynamic_out
-        else:
-            out = combine(dynamic_out, static_out)
+        out, aux = jacobian(_fun, has_aux=True)(diff_x)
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Actually, I think it's more complicated than this. (Possibly even undefined?) The jacobian should be something that is shaped as `output x input`, but here `static_out` is something that is shaped according to (the static part of) just `output`.

I think we need to define these operations in a different way altogether.

lockwo:
Hmmm, maybe a silly question but what's the actual need for static output? The input is already parsed to be dynamic and the output of the function has to be array types (already if you output a string or something it will fail, since jax.jacfwd doesn't know how to interact with it). Like suppose a simple example where one would probably expect jax and equinox to be the same (i.e. all jax types, no filtering actually needed)

```python
def cond_fun(val):
    x = val
    return jnp.mean(x) <= 5

def body_fun(val):
    x = val
    return x**2

def f(x):
    return jax.lax.while_loop(cond_fun, body_fun, x), 1.0

x = jnp.array([1.0, 2.0])
print("eqx", eqx.filter_jacfwd(f)(x))
print("jax", jax.jacfwd(f)(x))
```

will yield

```
eqx (Array([[ 4.,  0.],
       [ 0., 32.]], dtype=float64), 1.0)
jax (Array([[ 4.,  0.],
       [ 0., 32.]], dtype=float64), Array([0., 0.], dtype=float64, weak_type=True))
```

(this is independent of my PR, since this PR only effects if _static_out is None, but _static_out is not None here). 

If we just get ride of static out and return out directly (e.g. `return _dynamic_out, (_static_out, _aux)` -> `return _out, (None, _aux)`) then we see 

```
eqx (Array([[ 4.,  0.],
       [ 0., 32.]], dtype=float64), Array([0., 0.], dtype=float64, weak_type=True))
jax (Array([[ 4.,  0.],
       [ 0., 32.]], dtype=float64), Array([0., 0.], dtype=float64, weak_type=True))
```

patrick-kidger:
Yup. This is what I mean about probably needing to change the definition of `filter_jac{fwd,rev}`.

The resulting Jacobian is a "nested PyTree", whose structure/array-shapes are formed by composing the PyTrees of the outputs and inputs. But indeed if we have a string as both input and output, then there is no obvious way to compose these.

I suspect we can define this *either* for the case when the inputs may have static elements, *or* for the case where the outputs have static elements, but not both.

lockwo:
Yes, that makes sense. Personally, I lean towards just not allowing/supporting outputs to have static elements (seems like if you needed static outputs then there is a way for the user to combine them at the end). Static inputs are ubiquitous and seem essential to support.

patrick-kidger:
SGTM!

lockwo:
I removed the static filtering and added a warning about it in the docs

patrick-kidger:
LGTM! Thank you for spotting this, and for the fix! :)

PR #735, opened 2024-05-22
Update _embedding.py RopeEmbeddings

Artur-Galstyan:
It seems we forgot to make the `theta` variable public. This way, if you wanted to implement LLaMa3, you'd have a problem because they use a `theta` value of 50000.0.

### DIFF
--- a/equinox/nn/_embedding.py
+++ b/equinox/nn/_embedding.py
@@ -161,11 +161,7 @@ class RotaryPositionalEmbedding(Module, strict=True):
     """
 
     embedding_size: int = field(static=True)
-    theta: float = field(static=True)
-
-    def __init__(self, embedding_size: int, theta: float = 10000.0):
-        self.embedding_size = embedding_size
-        self.theta = theta
+    theta: float = field(static=True, default=10_000.0)
 
     def __check_init__(self):
         if self.embedding_size < 0:
@@ -180,7 +176,7 @@ class RotaryPositionalEmbedding(Module, strict=True):
 
     @staticmethod
     def precompute_freqs_cis(
-        embedding_size: int, end: int, theta: float = 10000.0
+        embedding_size: int, end: int, theta: float
     ) -> Complex[Array, "end half_of_embedding_size"]:
         freqs = 1.0 / (
             theta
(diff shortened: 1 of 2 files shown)


equinox/nn/_embedding.py
```diff
@@ -161,6 +161,11 @@ def process_heads(
     """
 
     embedding_size: int = field(static=True)
+    theta: float = field(static=True)
+
+    def __init__(self, embedding_size: int, theta: float = 10000.0):
+        self.embedding_size = embedding_size
+        self.theta = theta
```
patrick-kidger, line 168: This could be done without the explicit `__init__`:
```python
theta: float = field(static=True, default=10_000.0)
```

equinox/nn/_embedding.py
```diff
@@ -220,12 +225,16 @@ def __call__(
                 freqs_cis = internal_rope_embedding_cache[embedding_size]
                 freqs_cis_seq_len, _ = freqs_cis.shape
                 if seq_len > freqs_cis_seq_len:
-                    freqs_cis = self.precompute_freqs_cis(embedding_size, seq_len)
+                    freqs_cis = self.precompute_freqs_cis(
+                        embedding_size, seq_len, self.theta
```
patrick-kidger, line 229: Nit: let's remove the default argument for `precompute_freqs_cis`, I don't think it's ever used now?

patrick-kidger (commented):
Nits aside, LGTM!

equinox/nn/_embedding.py
Artur-Galstyan: Ah! I knew `field` must have an option for a default value, I must have overlooked it in the docs!

equinox/nn/_embedding.py
Artur-Galstyan: Yep!

patrick-kidger:
LGTM! Merged :)

And actually I've just noticed post-merge: could you update the docstring as well please, with the new argument?

Artur-Galstyan:
Sure, you can expect a PR tomorrow 👍

PR #739, opened 2024-05-28
Fix a broken link in the docs

matteoguarrera:
:smile:  
PS: pattern.md is commented in the mkdocs.yml and it raises an error


patrick-kidger:
Thanks for the fix!

As for `pattern.md`: I intend to uncomment that at some point fairly shortly, as the features it describes seems to have held up fairly stably in my own testing!

Issue #748, opened 2024-06-06
Diffusion example error


lockwo:
With a fresh environment and install diffrax, equinox, and einops (verified in a new colab instance, with jax 0.4.26) you get the following error when you try to run the diffusion example (specifically the ODE solver portion):

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
[<ipython-input-17-263240bbee7e>](https://localhost:8080/#) in <cell line: 1>()
----> 1 main()

10 frames
[<ipython-input-16-ee4e4fb06a9d>](https://localhost:8080/#) in main(patch_size, hidden_size, mix_patch_size, mix_hidden_size, num_blocks, t1, num_steps, lr, batch_size, print_every, dt0, sample_size, seed)
     55     sample_key = jr.split(sample_key, sample_size**2)
     56     sample_fn = ft.partial(single_sample_fn, model, int_beta, data_shape, dt0, t1)
---> 57     sample = jax.vmap(sample_fn)(sample_key)
     58     sample = data_mean + data_std * sample
     59     sample = jnp.clip(sample, data_min, data_max)

    [... skipping hidden 18 frame]

[<ipython-input-14-a0d1bc72bc29>](https://localhost:8080/#) in single_sample_fn(model, int_beta, data_shape, dt0, t1, key)
     32     y1 = jr.normal(key, data_shape)
     33     # reverse time, solve from t1 to t0
---> 34     sol = dfx.diffeqsolve(term, solver, t1, t0, -dt0, y1)
     35     return sol.ys[0]

    [... skipping hidden 15 frame]

[/usr/local/lib/python3.10/dist-packages/diffrax/_integrate.py](https://localhost:8080/#) in diffeqsolve(terms, solver, t0, t1, dt0, y0, args, saveat, stepsize_controller, adjoint, discrete_terminating_event, max_steps, throw, progress_meter, solver_state, controller_state, made_jump)
    776 
    777     # Error checking
--> 778     if not _term_compatible(
    779         y0, args, terms, solver.term_structure, solver.term_compatible_contr_kwargs
    780     ):

[/usr/local/lib/python3.10/dist-packages/diffrax/_integrate.py](https://localhost:8080/#) in _term_compatible(y, args, terms, term_structure, contr_kwargs)
    157     try:
    158         with jax.numpy_dtype_promotion("standard"):
--> 159             jtu.tree_map(_check, term_structure, terms, contr_kwargs, y)
    160     except ValueError:
[...]

PR #750, opened 2024-06-07
Fix einops

lockwo:
fixes https://github.com/patrick-kidger/equinox/issues/748


patrick-kidger:
Great stuff, thank you!

PR #753, opened 2024-06-10
Update users of jax.tree.map() to be more careful about how they hand…

hawkinsp:
…le Nones.

Due to a bug in JAX, JAX previously permitted jax.tree.map(f, None, x) where x is not None, effectively treating None as if it were pytree-prefix of any value. But None is a pytree container, and it is only a prefix of None itself.

Fix user code that was relying on this bug. Most commonly, the fix is to write jax.tree.map(lambda a, b: (None if a is None else f(a, b)), x, y, is_leaf=lambda t: t is None).

### DIFF
--- a/equinox/internal/_primitive.py
+++ b/equinox/internal/_primitive.py
@@ -77,7 +77,7 @@ def _replace_none(x):
 
 
 def _get_second(x, y):
-    return y
+    return None if x is None else y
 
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Wonderful stuff! Thank you for the fix Peter :)

PR #762, opened 2024-06-17
`warn_if`

### LINKED ISSUE #761: Add a `warn_if`
`error_if` is super useful! It would be great if there were also a `warn_if` function that surfaced a warning instead of raising an error.

nstarman:
Fixes #761.

I haven't yet cracked why the warnings aren't raised when a function is jitted.

### DIFF
--- a/equinox/_errors.py
+++ b/equinox/_errors.py
@@ -38,6 +38,13 @@ def _nan_like(x: Array | np.ndarray) -> Array | np.ndarray:
         return x
 
 
+def _warn_and_return_token(
+    msg: str, category: type[Warning] | None, stacklevel: int
+) -> bool:
+    warnings.warn(msg, category=category, stacklevel=stacklevel)
+    return False
+
+
 _tpu_msg = """
 
 Computation halted with the above error message. No Python exception is being raised as
@@ -437,10 +444,11 @@ def _warn(
 
     def handle_warning():  # pyright: ignore
         msg = msgs[index.item() if isinstance(index, Array) else index]  # pyright: ignore[reportCallIssue, reportArgumentType]
-        jax.debug.callback(
-            warnings.warn, msg, category=category, stacklevel=stacklevel + 2
+        shape_dtypes = jax.ShapeDtypeStruct((), jnp.bool_)
+        token = jax.pure_callback(
+            _warn_and_return_token, shape_dtypes, msg, category, stacklevel + 2
         )
-        return x
+        return lax.cond(token, lambda: x, lambda: x)
 
     return lax.cond(pred, handle_warning, lambda: x)
 
--- a/equinox/_errors.py
+++ b/equinox/_errors.py
@@ -3,7 +3,7 @@ import traceback
 import types
 import warnings
 from collections.abc import Sequence
-from typing import Any, cast, Literal
+from typing import Any, cast, Final, Literal
 
 import jax
 import jax._src.traceback_util as traceback_util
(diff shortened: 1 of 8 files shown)


patrick-kidger:
Let me know if you figure it out!
Other than that -- this has a lot of duplicate code with `error_if`. Probably this is most easily done with a Boolean flag instead?

nstarman:
> Other than that -- this has a lot of duplicate code with error_if. Probably this is most easily done with a Boolean flag instead?

I was considering that as well, but `warn` has a number of keyword arguments, like `category` and `stacklevel` that aren't shared with `raise`. 
Perhaps the combining could be done at the `branched_error_if_impl` level, which isn't exposed to the user and thus it's less important if the kwargs are a bit of a mess?

patrick-kidger:
Yeah, I think that's the right place to do it!

(Although FWIW I think some stuff like stacklevel may be incompatible with JIT? Or maybe you need to just preemptively grab the stack at trace time just in case you want to warn about it later?)

f0uriest:
Any update on this? It would be super useful for me!

nstarman:
@patrick-kidger I think I finally cracked it!

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+    stacklevel: int,
+) -> PyTree:
+    """Internal implementation for warn_if that handles Tracers."""
+
+    def handle_warning():  # pyright: ignore
+        msg = msgs[index.item() if isinstance(index, Array) else index]  # pyright: ignore[reportCallIssue, reportArgumentType]
+
+        jax.debug.callback(
+            warnings.warn, msg, category=category, stacklevel=stacklevel + 2
+        )
```
patrick-kidger: I'm inclined to suggest this should be a `jax.pure_callback`, much like the error implementation. Though I'm not sure how to do this without passing the data back and forth to the CPU; we only get away with it in the error case because it's actually terminal. Do you think that's doable>

This is because JAX is a bit funky with its handling of side effects and I prefer not having to reason about this, plus as an edge case I'm not convinced we'll get correct behaviour inside the Lineax linear solve (which has only an abstract eval rule, not an effectful abstract eval rule... I've never gone down the rabbit hole of seeing whether I could get that to result in an issue).

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+                        index = index.item()
+                    assert type(index) is int
+                    # Adjust stacklevel for direct call in eager mode
+                    # +1 accounts for this function, +1 for warn_if
+                    warnings.warn(
+                        msgs[index],
+                        category=category,
+                        stacklevel=stacklevel + 2,
+                    )
+                # Return early if condition was resolved at trace time
```
patrick-kidger: if `pred is True` but `isinstance(index, Tracer)` then I think we'll end up skipping the warning?

(I realise in practice that right now `index` is always set to an integer `0`...)

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+                return x
+            else:
+                return x
+
+    # Handle runtime case when pred is a Tracer (under JIT)
+    # Always use the callback path for consistency
+    dynamic_x, static_x = partition(x, is_array)
+    flat = jtu.tree_leaves(dynamic_x)
+    if len(flat) == 0:
```
patrick-kidger: I'm wondering if it might be better to make this case an explicit error, instead of allowing the warning to be raised during trace-time.

If it happens during compiletime and is missed at that time, then it won't appear at all during runtime. I'm mostly thinking of cases in which those might be two different users, either due to graph export or due to 'here is my big Jupyter notebook, just run the last cell'-type handovers.

Note that the same case doesn't occur for errors, as raising an explicit error during tracing prevents the graph from ever being compiled at all.

patrick-kidger (commented):
It's awesome to see this! (And I apologise for copilot's presence, I have no idea how to turn that off.)

nstarman:
> And I apologise for copilot's presence, I have no idea how to turn that off.

Most of the comments are correct! I personally find it useful. I plan to address the comments.

equinox/_errors.py
nstarman: > Do you think that's doable

No clue 😬.

> plus as an edge case I'm not convinced we'll get correct behaviour inside the Lineax linear solve

Do you have a specific test in mind? Something we could build Lineax with against this branch?

equinox/_errors.py
nstarman: I think this is now fixed.

equinox/_errors.py
nstarman: This should now raise an error.

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+        be raised if any batch element has `True`.
+    - `msg`: the string to display as an warning message.
+    - `category`: the warning category to use (defaults to `UserWarning` if `None`).
+    - `stacklevel`: the stack level for the warning (passed to `warnings.warn`).
+
+    **Returns:**
+
+    The original argument `x` unchanged. **If this return value is unused then the
+    warning check will not be performed.** (It will be removed as part of dead code
+    elimination.)
```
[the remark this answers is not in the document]
nstarman: Fixed.

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+    stacklevel: int = 1,
+) -> PyTree:
+    """Surfaces a warning based on runtime values. Works even under JIT.
+
+    **Arguments:**
+
+    - `x`: will be returned unchanged. This is used to determine where the warning check
+        happens in the overall computation: it will happen after `x` is computed and
+        before the return value is used. `x` can be any PyTree, and it must contain at
+        least one array.
```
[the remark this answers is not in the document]
nstarman: Done.

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+    - `pred`: a boolean for whether to raise an warning. Can be an array of bools; a
+        warning will be raised if any of them are `True`. If vmap'd then a warning will
+        be raised if any batch element has `True`.
+    - `msg`: the string to display as an warning message.
```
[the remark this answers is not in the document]
nstarman: Fixed.

equinox/_errors.py
```diff
@@ -364,3 +364,154 @@ def assert_dce(
...
+            else:
+                return x
+
+    # Handle runtime case when pred is a Tracer (under JIT)
+    # Always use the callback path for consistency
+    dynamic_x, static_x = partition(x, is_array)
+    flat = jtu.tree_leaves(dynamic_x)
+    if len(flat) == 0:
+        # Early return if no arrays - callback was already called in eager path
+        return x
```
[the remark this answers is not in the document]
nstarman: I think this is now fixed.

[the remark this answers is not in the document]
nstarman: Some refactoring happened, so it's not 100% clear the comment applies.

equinox/_errors.py
patrick-kidger: 1. Hmm, the callback is going to be a severe performance penalty anyway. Maybe we just pass e.g. a scalar zero back-and-forth through this, and then do `x + zero_returned_from_callback` afterwards.
2. The idea I have in mind is an operation with a side effect, using it within the implementation of a Lineax linear solver, then calling `lx.linear_solve` on this, and then wrapping that in something else that needs to know the effects of the whole computation (most minimally I suppose `jax.make_jaxpr(...)(...).effects`). I think the effect will be erased.

patrick-kidger:
Oh, one other thought: I wonder if we could implement this as a new mode for `error_if` (like the existing raise/breakpoint/nan/off)? This would help unify the fairly-similar code between the two.

nstarman:
@patrick-kidger I refactored it to be a mode. Actually, I took the old version of the code in which it was a separate function and asked Claude to unify the implementations and hand-edited from there. 🤷 look pretty good.
In a follow-up PR I think the `category` argument can be used for for the Exception type — which would address #1185.

equinox/_errors.py
```diff
@@ -108,11 +108,13 @@ def tpu_msg(_out, _index):
             return jtu.tree_map(_nan_like, _out)
 
         def handle_error():  # pyright: ignore
-            out = jax.pure_callback(raises, struct, index)
+            out = jax.pure_callback(raises, struct, index, vmap_method="sequential")
```
patrick-kidger, line 111: I don't think we want this to be sequential, I think ideally we should trigger just a single callback.

(On which note, I think the reason this isn't already specified is because `vmap_method` didn't exist when this code was first written, and it has presumably been using the legacy fallback all this time.)

equinox/_errors.py
```diff
@@ -130,18 +132,32 @@ def to_nan(_index):
...
         struct = jax.eval_shape(lambda: x)
         return lax.cond(pred, handle_error, lambda: x)
 
     elif on_error == "nan":
         return lax.cond(pred, ft.partial(jtu.tree_map, _nan_like), lambda y: y, x)
+    elif on_error == "warn":
+
+        def handle_warning():  # pyright: ignore
+            def warn_callback(_index):
+                msg_index = _index.item() if isinstance(_index, Array) else _index
```
patrick-kidger: I think we should always trigger the `isinstance`?

equinox/_errors.py
```diff
@@ -130,18 +132,32 @@ def to_nan(_index):
...
+                msg = cast(Any, msgs)[msg_index]  # pyright: ignore
+                warnings.warn(msg, category=category, stacklevel=stacklevel + 3)
+
+            jax.debug.callback(warn_callback, index)
```
patrick-kidger: I imagine we could get a close-to-best-of-both-worlds here by doing something like:
```python
def handle_warning():
    def warn_callback(_index):
        ...
        return 0
    zero = jax.pure_callback(warn_callback, ...)
    return x + zero
```
which would involve sending the minimal amount of data back-and-forth across the callback boundary, whilst still maintaining the approach that `eqx.error_if` will be DCE'd if its output is unused.

equinox/_errors.py
```diff
@@ -192,62 +219,71 @@ def error_if(
...
-            x = error_if(x, x < 0, "x must be >= 0")
-            # ...use x in your computation...
-            return x
+        ```python @jax.jit def f(x):
+            x = error_if(x, x < 0, "x must be >= 0") # ...use x in your
+            computation...  return x
 
-        f(jax.numpy.array(-1))
-        ```
+        f(jax.numpy.array(-1)) ```
```
patrick-kidger: (I think there's a lot of spurious reformatting happening here.)

equinox/_errors.py
```diff
@@ -364,3 +436,6 @@ def assert_dce(
     else:
         # Don't run if not JIT'ing, as without the compiler nothing will be DCE'd.
         return x
+
+
+# ------------------------------------
```
patrick-kidger: ?

patrick-kidger (commented):
Nice! This is much cleaner. Nitty comments only I think.

equinox/_errors.py
nstarman: I was thinking sequential, because warnings don't stop a computation and can stack, unlike exceptions.
But I can change this.

equinox/_errors.py
nstarman: Fixed!

equinox/_errors.py
nstarman: Done!

equinox/_errors.py
nstarman: Fixed!

equinox/_errors.py
nstarman: Removed!

patrick-kidger:
@nstarman – gosh, this took me a while to get around to. Okay so first of all, I've tidied up `error_if` itself:

1. I've rebased this on the new `dev`.
2. `msg` can now be a function, making it possible to use the token `x` in the error message, and also meaning that `branched_error_if` is not needed any more (and now just exists as a backwards-compatibility shim).
3. I've fixed some issues with vmap-of-`error_if`, as this code never actually got updated for when `pure_callback(..., vmap_method=...)` became required.
4. I've kept the short-circuit to the identity if the predicate is known-falsey at compile time, but remove the known-truthy short-circuit. This was actaully subtly incorrect: it was possible for an `error_if` to be DCE'd without this being triggered. It also added undesired complexity to the implementation.

Moving on to warnings:
1. I really like the unified internals, so I've kept these integrated into the `error_if` internals.
2. I like `warn_if` as being the user API, so I've re-introduced this.

WDYT? :)

(PS – I tried also fixing the way that `jax.debug.breakpoint` tends to explode (affecting the breakpoint mode of our errors), as per https://github.com/jax-ml/jax/issues/16732, and managed to identify https://github.com/jax-ml/jax/issues/38460 as a possible fix. Sadly there has been no movement yet on this on the JAX tracker, so we may still be stuck with buggy breakpoints though.)

Issue #763, opened 2024-06-20
`Conv` doesn't support complex dtype


ChenAo-Phys:
The convolutional layer in jax supports complex data types in principle, but the current `Conv` layer in equinox doesn't. The only problem is the default weight initialization `self.weight = jrandom.uniform(...)`, which doesn't work with complex types.

I want to have a minor change to make the default initialization `jrandom.normal` when the dtype is complex. I can submit a PR if it looks good to you @patrick-kidger

ChenAo-Phys:
I forgot to mention that some other layers like `Linear` have similar problems.

patrick-kidger:
It looks like JAX has intentionally not added support for this: https://github.com/google/jax/issues/4680

I'd be happy to add support for this, but I'd suggest keeping the distribution used consistent with the existing choice, i.e. make it be a uniform distribution. (Whether that's defined componentwise or magnitude-wise.)

Issue #755, opened 2024-06-12
`filter_shard` test crashes on JAX 0.4.29


patrick-kidger:
Looks like JAX used to do some "broadcasting" here, and no longer does. Bearing in mind that a PyTree may have arrays of multiple ranks, I'm not immediately sure what the appropriate fix is.

Tagging @homerjed, who originally contributed this; also @dlwh for visibility. What do you think?

homerjed:
Ah this is interesting. I will have a look at this but not sure a priori!

nasyxx:
Is it possible to align the rank by manual reshaping, put it to sharding, and then reshape it back to the original shape as a repair?

dlwh:
We're wrestling with sharding over in Levanter land right now, especially in the context of changing meshes around and such, which isn't exactly what you're talking about.

Because of the named axis stuff in Haliax the broadcasting thing you're talking about shouldn't come up since we compute a unique sharding for every named array (not shilling named axes, really, just excusing my lack of insight here)

I do think that in the general case, a model can't really be sharded usefully with broadcasting (since the FSDP axis and the TP axis aren't always in the same position), so it's not really a permanent solution. Data is a bit clearer where you almost always want to shard all arrays along the 0th axis, except for really fancy workloads.

So maybe just have a helper/special case for "shard every array along the 0th axis" and require anything else to be fully specified?

patrick-kidger:
Thanks for the input folks! (In particular for the great write up David.)

It sounds like the general answer to this is really an "its complicated", and fundamentally this is a topic that isn't a core competency for Equinox. Properly sharding multiply-ranked tensors is the kind of thing where a library like Haliax's named axes really start to shine instead!

@ChenAo-Phys has done a great job fixing the test over as part of #765. Let's keep `eqx.filter_shard` as a thin wrapper around `lax.with_sharding_constraint`, and consider the incompatibility here a JAX-level thing that we don't try to engineer around.

PR #765, opened 2024-06-21
Support complex dtypes in networks

ChenAo-Phys:
As discussed in #763, this PR adds support of complex dtypes for `Linear`, `Conv`, `GRUCell`, and `LSTMCell`.

A default initializer is defined in `eqx._misc` and called in initializations. Is it better to define it in `eqx.nn._misc`?

Some tests are added in `tests/test_nn.py` for complex dtypes and passed.

A minor problem that makes the test fail in my `jax==0.4.25` is also fixed in `tests/test_debug.py`.

### DIFF
--- a/equinox/_misc.py
+++ b/equinox/_misc.py
@@ -1,10 +1,7 @@
-from typing import Any
-
 import jax
 import jax.core
 import jax.numpy as jnp
-import jax.random as jrandom
-from jaxtyping import Array, PRNGKeyArray
+from jaxtyping import Array
 
 
 def left_broadcast_to(arr: Array, shape: tuple[int, ...]) -> Array:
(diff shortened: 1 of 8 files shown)


ChenAo-Phys:
It seems that the `eqx.filter_shard` test case in `tests/test_sharding.py` is no longer compatible with the newest `jax==0.4.30`, which is also seen in #342.

Now `PositionalSharding` requires all arrays in the pytree to have the same number of dimensions. But `NamedSharding` doesn't need it, so I modify the test and let it pass. The change of `PositionalSharding` could be a bug in the newest jax?

equinox/_misc.py
```diff
@@ -18,3 +21,17 @@ def default_floating_dtype():
...
     else:
         return jnp.float32
+
+
+def default_init(
+    key: PRNGKeyArray, shape: tuple[int, ...], dtype: Any, lim: float
+) -> jax.Array:
+    if jnp.issubdtype(dtype, jnp.complexfloating):
+        # only two possible complex dtypes, jnp.complex64 or jnp.complex128
+        real_dtype = jnp.float32 if dtype == jnp.complex64 else jnp.float64
```
patrick-kidger, line 31: I think a neater trick is this: https://github.com/patrick-kidger/lineax/blob/1909d190c1963d5f2d991508c1b2714f2266048b/lineax/_misc.py#L92-L93

equinox/_misc.py
```diff
@@ -18,3 +21,17 @@ def default_floating_dtype():
         return jnp.float64
     else:
         return jnp.float32
+
+
+def default_init(
```
patrick-kidger, line 26: Agreed, this should go in `equinox.nn._misc`!

patrick-kidger (commented):
This looks excellent. Minor comments aside I'd be very happy to merge this as-is.

Thank you in particular for fixing the sharding test! We were discussing this over in #755. I think I like your fix!

ChenAo-Phys:
The revision is done. It looks good to merge now.

patrick-kidger:
Yup, this all looks great to me!

Thank you for the excellent contribution, I'm really happy to have this in :)

PR #769, opened 2024-06-22
Fix for a case where we're hashing tracers.

### LINKED ISSUE #768: _TrivialClosureConvert mis-specifies dynamic attributes
I believe this should say `static=False`: https://github.com/patrick-kidger/equinox/blob/5a5bf287f73032201c34ff2c73e73573f8232739/equinox/_ad.py#L560

patrick-kidger:
Fixes #768.


patrick-kidger:
Hmm, I've realised this fix is actually wrong -- the whole reason we closure-convert is to make sure the tracers cross the `custom_vjp` boundary, after all.

I think perhaps the correct thing to do is the same `lambda` trick as this fix, but immediately before any `lax.while_loop` call? I'll give it a go.

patrick-kidger:
CI says yes :)
Merged!

Issue #782, opened 2024-07-12
Better error messages for strict


lockwo:
When programming on objects with lots of abstract method (e.g. https://github.com/lockwo/distreqx/blob/main/distreqx/distributions/_distribution.py), you sometimes forget to define a method (especially if the base class doesn't actually implement and just raises a NotImplementedError). In cases like this, you get a somewhat uninformative error message along the lines of `TypeError: Abstract strict 'eqx.Module's must be named starting with 'Abstract' or '_Abstract'.`

For example:

```python
import equinox as eqx
from abc import abstractmethod


class AbstractA(eqx.Module, strict=True):

    @abstractmethod
    def foo(self):
        pass


class B(AbstractA, strict=True):

    #def foo(self):
    #    return 

b = B()
```
yields `TypeError: Abstract strict 'eqx.Module's must be named starting with 'Abstract' or '_Abstract'. Got B when defining __main__.B.`

(uncommenting fixes it). Of course, this is expected and should error. However, it would be very helpful if the error message would indicate which abstract function isn't defined (which is the cause of the error). E.g. `TypeError: Abstract strict 'eqx.Module's must be named starting with 'Abstract' or '_Abstract'. Method 'foo' is undefined in concrete class B` or something like that would be great.

Does this suggestion sound within the realm of possibility?

patrick-kidger:
Absolutely! I'd be happy to take a PR on this.

PR #783, opened 2024-07-14
Better strict error

lockwo:
Addresses https://github.com/patrick-kidger/equinox/issues/782.

```python
import equinox as eqx
from abc import abstractmethod


class AbstractA(eqx.Module, strict=True):

    @abstractmethod
    def foo(self):
        pass


class B(AbstractA, strict=True):
    pass
    #def foo(self):
    #    return 

b = B()
```
old error`TypeError: Abstract strict 'eqx.Module's must be named starting with 'Abstract' or '_Abstract'. Got B when defining __main__.B.`

Now yields: 
```
TypeError: Abstract strict `eqx.Module`s must be named starting with 'Abstract' or '_Abstract'. 
Got B when defining __main__.B. For concrete moduels, B has abstract methods: ['foo'].
```

### DIFF
--- a/equinox/_module.py
+++ b/equinox/_module.py
@@ -477,10 +477,33 @@ class _ModuleMeta(ABCMeta):  # pyright: ignore
                     if not has_abstract_name:
                         # Invariant: abstract classes have names beginning with
                         # `Abstract`.
+                        inner = " "
+                        if len(cls.__abstractmethods__) > 0:
+                            inner = (
+                                inner
+                                + f"abstract methods: {list(cls.__abstractmethods__)}"
+                            )
+                        if len(cls.__abstractvars__) > 0:
+                            if len(inner) > 2:
+                                inner = inner + ", "
+                            inner = (
+                                inner
+                                + f"abstract variables: {list(cls.__abstractvars__)}"
+                            )
+                        if len(cls.__abstractclassvars__) > 0:
+                            if len(inner) > 2:
+                                inner = inner + ", "
+                            inner = (
+                                inner
+                                + f"abstract class variables: \
+                                {list(cls.__abstractclassvars__)}"
+                            )
+                        inner = inner + "."
                         raise TypeError(
                             "Abstract strict `eqx.Module`s must be named starting "
                             f"with 'Abstract' or '_Abstract'. Got {name} when defining "
-                            f"{cls.__module__}.{cls.__qualname__}.",
+                            f"{cls.__module__}.{cls.__qualname__}. For concrete "
+                            f"moduels, {name} has" + inner
                         )
                 else:
                     if has_abstract_name:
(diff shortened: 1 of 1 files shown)


equinox/_module.py
```diff
@@ -477,10 +477,33 @@ def __init__(self, *args, **kwargs):
...
+                            )
+                        if len(cls.__abstractclassvars__) > 0:
+                            if len(inner) > 2:
+                                inner = inner + ", "
+                            inner = (
+                                inner
+                                + f"abstract class variables: \
+                                {list(cls.__abstractclassvars__)}"
+                            )
+                        inner = inner + "."
```
patrick-kidger, line 501: I think this woudl be simply done via something like
```python
pieces = []
if ...:
    pieces.append(some string)
...
message = ", ".join(pieces)
```
?

equinox/_module.py
```diff
@@ -477,10 +477,33 @@ def __init__(self, *args, **kwargs):
...
                         raise TypeError(
                             "Abstract strict `eqx.Module`s must be named starting "
                             f"with 'Abstract' or '_Abstract'. Got {name} when defining "
-                            f"{cls.__module__}.{cls.__qualname__}.",
+                            f"{cls.__module__}.{cls.__qualname__}. For concrete "
+                            f"moduels, {name} has" + inner
```
patrick-kidger, line 506: I don't think this will do the right for modules that are forced to be abstract? (`strict=StrictConfig(force_abstract=True)`)

Also typo `moduels`.

equinox/_module.py
lockwo: Yea, they are equivalent, I changed it to that way. Neither seemed particularly more readable to me.

equinox/_module.py
lockwo: Fixed typo and added a flag to ignore the inner if force is true.

equinox/_module.py
patrick-kidger: Technically the repeated concatenation is quadratic ;)

equinox/_module.py
```diff
@@ -477,11 +477,32 @@ def __init__(self, *args, **kwargs):
                     if not has_abstract_name:
                         # Invariant: abstract classes have names beginning with
                         # `Abstract`.
-                        raise TypeError(
+                        main = (
                             "Abstract strict `eqx.Module`s must be named starting "
-                            f"with 'Abstract' or '_Abstract'. Got {name} when defining "
-                            f"{cls.__module__}.{cls.__qualname__}.",
+                            f"with 'Abstract' or '_Abstract'. Got {name} when defining"
+                            f"{cls.__module__}.{cls.__qualname__}."
```
patrick-kidger, line 483: I think you're missing a space across these two lines? The class name will appear right up against the "defining".

Other than that, LGTM!

lockwo: fixed

patrick-kidger:
Awesome -- merged! Thank you for the fix :)

PR #799, opened 2024-08-12
Added the offset to RoPE embedding and fixed the pre-commit pyright

Artur-Galstyan:
As mentioned in #704, this is the fix for the RoPE embeddings. I also added `pyright: ignore` to the `test_nn.py` file. I think perhaps someone forgot to run the PCH hook. I checked GH and it seems we don't run the Pyright check anymore?


equinox/nn/_embedding.py
```diff
@@ -126,18 +126,22 @@ def __call__(...):
                 def process_heads(
                     query_heads: Float[Array, "seq_length num_heads qk_size"],
                     key_heads: Float[Array, "seq_length num_heads qk_size"],
-                    value_heads: Float[Array, "seq_length num_heads vo_size"]
+                    value_heads: Float[Array, "seq_length num_heads vo_size"],
+                    index: Int[Array, ""]
```
patrick-kidger, line 130: I think this will fail, as we don't pass this extra argument inside MHA?

patrick-kidger (commented):
Hmm, I don't think this was actually a bug before? It's really just a new feature for `RotaryEmbedding`, in that it now supports an offset.

My concern here is that in order to use it, we'd need to make a backward-incompatible change to MHA. (To pass the index.)

Since we've decided against implementing KV caching in MHA, then I don't think this really comes up.

equinox/nn/_embedding.py
```diff
@@ -194,12 +201,14 @@ def precompute_freqs_cis(
     def __call__(
         self,
         x: Float[Array, "seq_length embedding_size"],
+        offset: Int[Array, ""] = jnp.array(0),
```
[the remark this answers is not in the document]
Artur-Galstyan: Ah, I didn't know. Is this just Equinox specific or should I in general avoid JAX arrays for defaults? 

In any case, I made this mandatory then and added some more examples. Unfortunately, something like:

```python
offset: Optional[Int[Array, ""]] = None

offset = offset if offset is not None else jnp.array(0)
...

doesn't work under JIT and `jax.lax.select` doesn't work either because it evaluates both branches. 

You don't happen to have a solution for these kind of issues somewhere in your JAX wizard hat? Understanding JAX's behaviour under JIT is definitely something I need to catch up on!

File: equinox/nn/_embedding.py (hunk 1 above)
Comment (Artur-Galstyan) [reply equinox/nn/_embedding.py]: Yep, you're right. I've added a more complete example for this.

**Comment by Artur-Galstyan** [comment]
I agree that this is not a bug per-se but still a shortcoming in our current version, especially when compared to a [PyTorch counterpart](https://pytorch.org/torchtune/stable/_modules/torchtune/modules/position_embeddings.html#RotaryPositionalEmbeddings.forward) which includes the `input_pos` in its `forward` function. 

> My concern here is that in order to use it, we'd need to make a backward-incompatible change to MHA. (To pass the index.)
Since we've decided against implementing KV caching in MHA, then I don't think this really comes up.

In the example I provided, it's even possible to make it work with the offset even without changing the MHA implementation by using `functools.partial`. TBH, I didn't think it would work and that it would re-JIT on every call when using this:

```python
        x = self.mha_attention(
            query=query,
            key_=key_,
            value=value,
            process_heads=functools.partial(process_heads, index=index), # <--
        )
```

But I guess the compiler is pretty smart!

equinox/nn/_embedding.py
```diff
@@ -161,13 +206,16 @@ def process_heads(
     """
 
     embedding_size: int = field(static=True)
+    max_seq_length: int = field(static=True)
```
patrick-kidger, line 209: Adding this argument is a breaking change unfortunately.

equinox/nn/_embedding.py
```diff
@@ -208,44 +258,55 @@ def __call__(
         A JAX array of shape `(seq_length, embedding_size)`, with the rotary positional
         encoding applied to the input.
         """
-
+        print("JIT ROPE")
```
patrick-kidger, line 261: !

patrick-kidger (commented):
I see.

Okay, so on balance I think I'm happy to have this extra argument, but for simplicity let's not include any extra examples in the documentation. We're not discussing this kind of progressive updating anywhere else, and I'd like to keep this change as easy to review as possible.

equinox/nn/_embedding.py
Artur-Galstyan: Oh, this is embarrassing 😮

Artur-Galstyan:
Ok, so I've removed the `max_seq_length` argument, it's a breaking change with no backwards compatibility but more importantly, because it actually provides no benefit. 

<details>
<summary>Unfold, if you want to know</summary>
Initially, I added it, because I wanted to avoid re-JITting. But in the current implementation, we are not enforcing the input array to the RoPE module to be of length <code>max_seq_length</code> (and then applying a mask to effectively only "use" the intended <code>seq_length</code>). The input array can have any <code>seq_length</code>, which means the module <b>will always</b> be re-JITted. Thus, the <code>max_seq_length</code> argument is obsolete. In other words, if we were to ever include the <code>max_seq_length</code> argument, we would also need to include not just the <code>offset</code> but also the <code>cutoff</code> position and then apply a mask on the input array. But that would be <i>an even breakier change</i> 😄 .<sub>(Though, personally, I don't mind breaking changes as long as they are an improvement and properly communicated)</sub>

</details>

equinox/nn/_embedding.py
```diff
@@ -118,31 +118,50 @@ class RotaryPositionalEmbedding(Module, strict=True):
         ```python
         class TransformerBlock(eqx.Module):
             rope_embeddings: RotaryPositionalEmbedding
+            mha_attention: MultiheadAttention
```
patrick-kidger, line 121: I think I liked the old example! It was much more concise.

equinox/nn/_embedding.py
```diff
@@ -194,12 +213,14 @@ def precompute_freqs_cis(
     def __call__(
         self,
         x: Float[Array, "seq_length embedding_size"],
+        offset: Int[Array, ""],
```
patrick-kidger, line 216: Let's make this `offset: Int[ArrayLike, ""] = 0` so that we have backward compatibility?

equinox/nn/_embedding.py
```diff
@@ -233,19 +254,22 @@ def __call__(
                 )
                 internal_rope_embedding_cache[embedding_size] = freqs_cis
 
+        freqs_cis = jax.lax.dynamic_slice_in_dim(freqs_cis, offset, seq_len)
```
patrick-kidger, line 257: This looks wrong to me. If we hit the `else` branch of `if embedding_size in internal_rope_embedding_cache ... else:` above, then we'll compute an array of length `seq_len`, which will not all be valid when sliced into here -- we'll be indexing off the end.

patrick-kidger (commented):
Responding to your comment on backward compatibility -- I think this is very important :)

I'm willing to make a compatibility break if absolutely required, but that's generally been fairly unusual / often only on niche features / only in the early life of a library prior to achieving adoption.

It's just not providing a useful tool to those downstream if we do that kind of thing!

equinox/nn/_embedding.py
Artur-Galstyan: Yes, this was a bug. The fix is to ensure that `freqs_cis` is at least as long as `seq_len + offset`.

Artur-Galstyan:
I actually caught another bug in the implementation, which I hadn't noticed before. Our previous implementation grouped the sin/cos like this:

`[cos1, cos2, cos3, cos4, sin1, sin2, sin3, sin4]` (all grouped)

but it should have been interleaved like this:

`[cos1, sin1, cos2, sin2, cos3, sin3, cos4, sin4]`

I used the implementation from [lucidrains](https://github.com/lucidrains/rotary-embedding-torch/tree/main) as a reference for the expected values and updated the test accordingly.

2 problems are left to fix: 

1) Pyright complains at the `TODO` spots because `ArrayLike` might be a complex number and `Operator ">" not supported for types "complex" and "int"` :( 

2) It keeps re-jitting when using integers for the offset and when using arrays we get a `ConcretizationTypeError`. There is an [_almost MVP_](https://gist.github.com/Artur-Galstyan/544b4d29261a6774bd1c24e96cd5eb74) if you're curious. I'll invest more time into this - I feel like there should be a good solution.

equinox/nn/_embedding.py
```diff
@@ -176,14 +181,14 @@ def rotate_half(x: Float[Array, "seq_length embedding_size"]):
...
-        embedding_size: int, end: int, theta: float
+        embedding_size: int, end: Int[ArrayLike, ""], theta: float
     ) -> Complex[Array, "end half_of_embedding_size"]:
         freqs = 1.0 / (
             theta
             ** (jnp.arange(0.0, embedding_size, 2)[jnp.newaxis, :] / embedding_size)
         )
 
-        t = jnp.arange(float(end))
+        t = jnp.arange(end / 1.0)  # promote to float
```
Artur-Galstyan, line 191: (This was also because Pyright kept complaining)

patrick-kidger: I think I preferred `float(end)`, rather than implicitly promoting. Even if that means a pyright-ignore.

patrick-kidger:
I think both interleaved and non-interleaved are acceptable. In fact when we first wrote this, I checked this against the ESM2 implementation [here](https://github.com/facebookresearch/esm/blob/main/esm/rotary_embedding.py) to ensure correctness.

We can't switch this now for backward compatibility.

patrick-kidger:
Bother, this got autoclosed because the target branch got merged. Feel free to re-open.

Artur-Galstyan:
No worries, I'll fix it once I get back from vacation.

PR #804, opened 2024-08-20
Update parallelism.ipynb

garymm:
Minor improvements:
* `nvidia-smi` works on a single computer, not a cluster.
* The example code already does donate arguments, so the sentence about adding that is unneeded.
* Remove backslashes around double quotes. They are rendered in the equinox web site and don't appear to be needed.


patrick-kidger:
These all look good to me!
Thank you for the improvements :)

PR #805, opened 2024-08-20
Remove polyfill

ColCarroll:
This updates the documentation by removing polyfill. IANAFED (I am not a front end developer), but I guess polyfill.io allowed all browsers to use the es6 features. Turns out:

- polyfill [has been serving malware](https://www.techradar.com/pro/security/thousands-of-websites-told-to-ditch-polyfill-service-after-chinese-hackers-hijack-it-to-serve-malware)
- https://squidfunk.github.io/mkdocs-material/reference/math/#mathjax-mkdocsyml no longer suggests polyfill
- It [probably isn't needed for 97% of traffic for es6 features](https://caniuse.com/?search=es6#google_vignette)

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -42,9 +42,8 @@ edit_uri: ""  # No edit button, as some of our pages are in /docs and some in /e
 strict: true  # Don't allow warnings during the build process
 
 extra_javascript: 
-    # The below three make MathJax work, see https://squidfunk.github.io/mkdocs-material/reference/mathjax/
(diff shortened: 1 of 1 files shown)


ColCarroll:
Oh right, you can actually get a better conversation at https://github.com/squidfunk/mkdocs-material/issues/7295

patrick-kidger:
Thank you for the fix! Didn't realise we still had that floating around there...

ColCarroll:
thanks for a pleasant mkdocs setup!

Issue #818, opened 2024-08-29
_DistinctTuple is not a tuple causing errors in tree_at


danielward27:
Take for example:
```python
import equinox as eqx

class MyModule(eqx.Module):
    tup: tuple

my_mod = MyModule(())
eqx.tree_at(lambda t: t.tup, my_mod, replace_fn=len)
```
Gives:
```
TypeError: object of type '_DistinctTuple' has no len()
```
I would expect it to replace the node with 0.

patrick-kidger:
Closing as fixed in #819.

PR #819, opened 2024-08-29
Make _DinstinctTuple inherit from tuple

danielward27:
Fix for https://github.com/patrick-kidger/equinox/issues/818.


patrick-kidger:
LGTM -- merged! Thank you for both the report and for the fix :)

PR #821, opened 2024-08-29
Add editorconfig

NeilGirdhar:


.editorconfig
```diff
@@ -0,0 +1,23 @@
...
+max_line_length = 88
+indent_size = 4
+
+[*.toml]
+max_line_length = 150
+indent_size = 4
+
+[*.yml]
+max_line_length = 120
+indent_size = 2
```
patrick-kidger, line 19: I'd cut these two. Right now we rarely use either kind of file and both are unformatted.

patrick-kidger (commented):
Do you feel that this adds much over the existing autoformatting (via pre-commit + ruff)? It feels more natural to me to have just a single place that handle this.

NeilGirdhar:
editorconfig automatically configures editors like vim, emacs, etc. so that they know how to save files, how long the lines should be (for examples, vim's automatic line width formatter), where to show the line length helper.

.editorconfig
NeilGirdhar: This is for editors, not formatters.  The indent size tells the editor how many spaces to add when <tab> is pressed.    I got rid of the line length then.

NeilGirdhar:
You can see a list of editors that support this file here: https://editorconfig.org/

patrick-kidger:
Indeed, I'm familiar with `.editorconfig`.

My point is that if we care about such things then IMO it'd be better to have a single source of truth on such matters -- an autoformatter -- and allow anything at all whilst editing.

Anyway, not a strong opinion. `*.py` files are the only ones we attempt to format so trim this down to just that and I'd be happy to merge this.

NeilGirdhar:
> My point is that if we care about such things then IMO it'd be better to have a single source of truth on such matters -- an autoformatter -- and allow anything at all whilst editing.

Yes, I see your point.  I guess I forgot that you've got the autoformatter enabled.

Anyway, updated as requested.

patrick-kidger:
Alright, merged! :) Thank you for the contribution!

NeilGirdhar:
Thanks for being flexible about this.  Maybe we should propose to Ruff that editorconfig be the defaults to satisfy your "one source of truth" desire?

patrick-kidger:
Seems not planned: https://github.com/astral-sh/ruff/issues/1530

NeilGirdhar:
Ah, nice find!

Issue #820, opened 2024-08-29
MyPy doesn't understand Module


NeilGirdhar:
MyPy only looks at the first definition of Module, so it doesn't see the redefinition in `TYPE_CHECKING`:
```python
import dataclasses
from typing import TYPE_CHECKING, dataclass_transform


class _ModuleMeta(type):  # This needs to be in a if not TYPE_CHECKING block.
    pass


if TYPE_CHECKING:
    @dataclass_transform(field_specifiers=(dataclasses.field,))
    class _ModuleMeta(type):
        pass


class Module(metaclass=_ModuleMeta):
    pass


class A(Module):
    x: int


A(1)  # MyPy doesn't realize that A is a dataclass
```

patrick-kidger:
Ah -- this must be a mypy vs pyright difference.

Happy to take a PR on this :)

patrick-kidger:
Fixed in #822 :)

PR #822, opened 2024-08-29
Fix MyPy's view of _ModuleMeta

NeilGirdhar:


patrick-kidger:
The downside of this approach is that it means we lose type-checking within the definition of `_ModuleMeta` itself.

Would something like the following work?
```python
# As today, but the name has an additional underscore
class __ModuleMeta(ABCMeta):
    ...

if TYPE_CHECKING:
    class _ModuleMeta: ...  # as today
else:
    _ModuleMeta = __ModuleMeta
```
?

NeilGirdhar:
Yup, that's a good idea!

patrick-kidger:
Looks like tests are failing (due to something to do with name mangling). Other than that, LGTM!

NeilGirdhar:
Fixes #820

patrick-kidger:
Wonderful stuff. LGTM, and merged!

PR #826, opened 2024-08-31
[Feat] Add fused QKV projection

neel04:
Fused QKV projections, which is a common optimization used in LLM training.

This exposes a `fuse_qkv` option when initialising equinox's MHA, which the user can enable if they know the corresponding `seqlen`s are correct beforehand. Defaults to `True` as this is applicable for the majority of use cases.

For the failed tests, I'll explicitly disable the fusion, and add another version of them with fusion enabled.

### DIFF
--- a/equinox/nn/_attention.py
+++ b/equinox/nn/_attention.py
@@ -118,9 +118,6 @@ class MultiheadAttention(Module, strict=True):
         refer to it as "$d_\text{model}$".
     """
 
-    query_proj: Linear
-    key_proj: Linear
-    value_proj: Linear
     output_proj: Linear
     dropout: Dropout
 
@@ -135,6 +132,11 @@ class MultiheadAttention(Module, strict=True):
     use_key_bias: bool = field(static=True)
     use_value_bias: bool = field(static=True)
     use_output_bias: bool = field(static=True)
+
+    qkv_proj: Linear | None = None
+    query_proj: Linear | None = None
+    key_proj: Linear | None = None
+    value_proj: Linear | None = None
 
     def __init__(
         self,
@@ -151,6 +153,7 @@ class MultiheadAttention(Module, strict=True):
         use_output_bias: bool = False,
         dropout_p: float = 0.0,
         inference: bool = False,
+        fuse_qkv: bool = True,
         dtype=None,
         *,
         key: PRNGKeyArray,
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Thanks for the contribution!

So we recently decided over in #704 that we'd like to avoid extending `eqx.nn.MultiheadAttention` further for the time being -- there are too many kinds of attention variants, and it's getting a bit out of hand trying to support them all. The two main things going on there are that:
1. Realistically most people with an opinion on the topic are going to implement their own layer from scratch anyway.
2. There's still a changing landscape of kernels being called underneath (JAX seems to have a few new `jax.maybe_something_experimental.some_kind_of_attention` functions at this point.), so it's hard to really do a good job offering best-possible performance here.

I think for this reason that this PR is out-of-scope for now, I'm afraid.

(That aside, we definitely couldn't ever have made `fuse_qkv=True` the default, as this would be a backward-incompatible change.)

neel04:
Hmm...  I get your point. Personally, I think we should just be supporting vanilla MHA - which can be used as x-attn and MHSA already. This covers the vast majority of usecases without going too into the weeds about other attention implementations. I agree supporting other attention variants should be out of scope.

For the second point, JAX is offering a SDPA utility - not full attention, very much like torch's [SDPA](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html). The reason its stripped down is that it can be used in an attention layer's _construction_ easily, but you still have to handle things - like QKV and mask construction. 

I can actually add that to `eqx.nn.MultiHeadAttention` if you want - so you get the full performance XLA has to offer and add on fused QKV.  As for KV-cache, it's an inference-time optimization which is a lot of added complexity for not a lot of utility. Most people using `equinox`/`jax` are researchers so they aren't planning on hosting their models - and those who do would re-implement it anyways. So I think `equinox` should have a fast & efficient implementation of cross-attention and self-attention for training atleast, out-of-the-box.

neel04:
Also, for the MHA API - why are biases exposed for every attention component? what's the use-case for such granular control of biases? most LLMs either don't use biases at all or use them - I don't think really there's a need for such granular control if nobody is using it and it ends up cluttering the API 🤷

patrick-kidger:
> I can actually add that to eqx.nn.MultiHeadAttention if you want

I'm open to considering this, although last I checked it still had some footguns about working on every backend! If you think you can always dispatch to the best thing for each hardware then sure.

> why are biases exposed for every attention component?

This was back in the early days of Equinox, and we just decided to do the more-flexible thing. No strong argument either way here I think.

neel04:
>  although last I checked it still had some footguns about working on every backend

So apparently, according to the [impl.](https://github.com/google/jax/blob/b6aedbc41fff9b71c167e42e251766bbeb2ac41a/jax/_src/nn/functions.py#L1015), the `XLA` backend should work fine but `CuDNN` is "restricted to a subset of shapes and dtypes".

We can just expose the `implementation` argument that the user can override if they wish so, or leave it as it is for compatibility and the mercy of XLA:GPU.

> No strong argument either way here I think.

I think it adds unnecessary clutter in the documentation, and makes it so the user has to explicitly specify multiple arguments when there should only be 1 for the biases. Plus, that way its more consistent with other frameworks & codebases so I see this as a net positive.

LMK what stuff I can do here, or if this PR is better off closed.

patrick-kidger:
Hmm, the implementation claims that by default it will attempt to select the best backend. Maybe that's just what we should let it do.

Indeed I think this PR is out of scope for now. :)

Issue #824, opened 2024-08-30
Spliting a module in more than 2 partitions


francois-rozet:
I regularly need to split modules into more than 2 partitions. For example, linear parameters, conv parameters and others. It is not easy to do with the current `eqx.partition` function.

In Inox, [`inox.tree_partition`](https://inox.readthedocs.io/stable/api/inox.tree_util.html#inox.tree_util.tree_partition) allows any number of partitions. If a single predicate (filter) is provided, it has the same behavior as `eqx.partition`. If more than one are provided, ambiguities are resolved by the order of the predicates.

I think it would be easy to adapt the logic to `eqx.partition`. I can do it if you think this feature is worth adding.

patrick-kidger:
Sure, I'd be happy to take a PR on that!

francois-rozet:
Hi @patrick-kidger, I have created the PR!

I took the opportunity to do a few speed tests. In Equinox, the partitions are trees where some leaves are replaced by `None` (or another sentinel). In Inox, the partitions are path-leaf mappings (i.e. flat dictionaries), similar to `state_dict` in PyTorch.

There are three advantages to using dictionaries:
1. They are easy to "read" for a human.
2. They are safe to pickle/serialize.
3. For simple pytrees, the cost of flattening is proportional to the number of leaves. With tree partitions (Equinox), the number of leaves is constant, even when there is a lot of metadata leaves compared to array leaves. With dictionary partitions (Inox), the array partition is much smaller (has less leaves) than the full tree. This results in faster partition flattening and, consequently, faster evaluation of jitted functions.

Follows a small example to illustrate the difference.

```python
import equinox as eqx
import inox
import jax

tree = None

for i in range(100):
    tree = {
        "child": tree,
        "index": i,
        "array": jax.numpy.ones(i + 1),
        "meta": ("irrelevant", "information"),
    }

def f(tree):
    total = 0

    while tree:
        total = total + tree["array"][tree["index"]]
        tree = tree["child"]

    return total

## Equinox
arrays, others = eqx.partition(tree, eqx.is_array)

def g(arrays):
    tree = eqx.combine(arrays, others)
    return f(tree)

g_jit = jax.jit(g)

%timeit g_jit(arrays)  # 67 µs ± 6.64 µs per loop

## Inox
treedef, arrays, others = inox.tree_partition(tree, jax.Array)

def g(arrays):
    tree = inox.tree_combine(treedef, arrays, others)
    return f(tree)

[...]

PR #829, opened 2024-09-02
Allow splitting to any number of partitions

francois-rozet:
Targets #824 

Hi @patrick-kidger, in this PR I modify `eqx.partition` such that it accepts any ($N$) number of filter specifications and returns $N + 1$ partitions. To keep backward compatibility, I have kept the `filter_spec` argument and added a variadic positional argument `*filter_specs`. This however forces `replace` and `is_leaf` to become keyword arguments. I think that is ok.

### DIFF
--- a/equinox/_filters.py
+++ b/equinox/_filters.py
@@ -135,21 +135,45 @@ def filter(
 def partition(
     pytree: PyTree,
     filter_spec: PyTree[AxisSpec],
+    *filter_specs: PyTree[AxisSpec],
     replace: Any = None,
     is_leaf: Optional[Callable[[Any], bool]] = None,
 ) -> tuple[PyTree, PyTree]:
     """Splits a PyTree into two pieces. Equivalent to
     `filter(...), filter(..., inverse=True)`, but slightly more efficient.
 
+    More generally, provide $N$ filter specifications to split the tree into $N + 1$
+    non-overlapping partitions.
+
     !!! info
 
         See also [`equinox.combine`][] to reconstitute the PyTree again.
     """
 
-    filter_tree = jtu.tree_map(_make_filter_tree(is_leaf), filter_spec, pytree)
-    left = jtu.tree_map(lambda mask, x: x if mask else replace, filter_tree, pytree)
-    right = jtu.tree_map(lambda mask, x: replace if mask else x, filter_tree, pytree)
-    return left, right
+    filter_trees = [
+        jtu.tree_map(_make_filter_tree(is_leaf), spec, pytree)
+        for spec in (filter_spec, *filter_specs)
+    ]
+
+    partitions = []
+
+    for i in range(len(filter_trees)):
+        partition = jtu.tree_map(
+            lambda x, curr, *prev: x if curr and not any(prev) else replace,
+            pytree,
+            filter_trees[i],
+            *filter_trees[:i],
+        )
+
+        partitions.append(partition)
+
+    rest = jtu.tree_map(
+        lambda x, prev: replace if any(prev) else x,
+        pytree,
+        *filter_trees,
+    )
+
+    return *partitions, rest
 
 
 def _combine(*args):
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -143,8 +143,24 @@ def test_partition_and_combine(getkey):
 
 
 def test_partition_subtree():
-    a, b = eqx.partition([(1,), 2], [True, False])
-    eqx.combine(a, b)
+    pytree = [(1,), 2]
+
+    a, b = eqx.partition(pytree, [True, False])
+
+    assert eqx.combine(a, b) == pytree
+
+
+def test_partition_multi():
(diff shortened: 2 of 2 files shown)


equinox/_filters.py
```diff
@@ -135,21 +135,45 @@ def filter(
 def partition(
     pytree: PyTree,
     filter_spec: PyTree[AxisSpec],
+    *filter_specs: PyTree[AxisSpec],
```
knyazer, line 138: I dislike breaking backwards compatibility! Could you instead make ```filter_specs``` a keyword argument? This will be less ergonomic, since will force you to actually spell ```eqx.partition(..., filter_specs=())```, but it is worth it.

And, the _ordering_ of the filters matters in your implementation, so it is more 'logical' to have them as a sequence, and not as a variadic: in the current implementation it is not clear that order of filters matters a lot, but if it is another parameter, then it's pretty clear.

equinox/_filters.py
```diff
@@ -135,21 +135,45 @@ def filter(
...
     replace: Any = None,
     is_leaf: Optional[Callable[[Any], bool]] = None,
 ) -> tuple[PyTree, PyTree]:
     """Splits a PyTree into two pieces. Equivalent to
     `filter(...), filter(..., inverse=True)`, but slightly more efficient.
 
+    More generally, provide $N$ filter specifications to split the tree into $N + 1$
+    non-overlapping partitions.
+
```
knyazer, line 147: Could you slightly extend on the documentation? Before I looked at the code, it was not obvious what multiple partitioning does. My first assumption was that you would split things with some redundancy:
```
everything_that_is_array, everything_that_is_inexact_array, rest = eqx.filter(x, eqx.is_array, eqx.is_inexact_array)
```
I feel like this is more in line with the previous partitioning logic, at least that is how my brains understand the standard partitioning. 

Actually, I really like this way to specify things, but it does indeed break backwards compatibility :( 

I would probably implement it as a separate partition function, e.g. ```eqx.multi_partition```? This way the syntax could be wonderfully concise, and compatibility will be kept intact. Anyhow, that is for Patrick to decide, I'm just passing by.

equinox/_filters.py
```diff
@@ -135,21 +135,45 @@ def filter(
...
+        partitions.append(partition)
+
+    rest = jtu.tree_map(
+        lambda x, prev: replace if any(prev) else x,
+        pytree,
+        *filter_trees,
+    )
+
+    return *partitions, rest
```
knyazer, line 176: Especially considering how much new code is added into this 'simple' function, separating it into a distinct function makes sense to me.

equinox/_filters.py
```diff
@@ -135,21 +135,45 @@ def filter(
...
+    filter_trees = [
+        jtu.tree_map(_make_filter_tree(is_leaf), spec, pytree)
+        for spec in (filter_spec, *filter_specs)
+    ]
+
+    partitions = []
+
+    for i in range(len(filter_trees)):
+        partition = jtu.tree_map(
+            lambda x, curr, *prev: x if curr and not any(prev) else replace,
```
knyazer, line 162: Are you sure it should be ```any``` here, not ```jnp.any```? Could you also test this under jit, just to make sure it works correctly?

knyazer (commented):
Thanks for the contribution! I left a bunch of comments all over the place, would appreciate to hear your opinion on them.

equinox/_filters.py
patrick-kidger: FWIW if we made this a keyword then we'd need to allow not passing `filter_spec` itself in the case that it's used.

Hmm, this backward compatibility break may sink this whole endeavour.

equinox/_filters.py
patrick-kidger: This does also speak to another good point, how do overlapping filters work? It's not really clear which group things end up in.

The current 'only split into 2' scenario necessitates making this explicit.

patrick-kidger (commented):
@knyazer has spotted some meaningful flaws, I think.

For that reason I'm afraid I'm inclined not to take this PR -- neither the backward compatibility break nor the unclear semantics seem worth the trade-off for this feature, I'm afraid.

equinox/_filters.py
francois-rozet: When a leaf satisfies several filters, it is contained in the first one that is satisfied, as determined by the user-provided order.

equinox/_filters.py
francois-rozet: I don't understand your concern here. The filters should return `True` or `False` values, never arrays. In fact, you should never use booleans arrays for control flow.

equinox/_filters.py
francois-rozet: IMO forcing `replace` and `is_leaf` to be keyword arguments is actually an improvement as it could prevent user errors. It does break backward compatibility slightly, but I don't think there are many (if any) codebases using these arguments as positional. Note that `jax.tree_util.tree_map` forces `is_leaf` as a keyword argument.

Issue #810, opened 2024-08-22
Spectral Norm Features


lockwo:
The original spectral norm paper didn't accurately estimate spectral norms (e.g. in the convolution case they just reshape it then do normal matrix, but this isn't the actual transpose operation, see: https://arxiv.org/abs/2009.02773 for more) which is now the general default in neural network packages (equinox included). However, it's not that much LoC to get a spectral norm that can do one or both of the following (sources: https://arxiv.org/abs/1804.04368, https://arxiv.org/abs/1811.07457)
- actually approximates the correct spectral norm (and thus adheres to the 1-lipshitz cap, which original SN isn't super great at, see fig 3 of https://arxiv.org/abs/1802.04034)
- can be made numerically meaningful (by guaranteeing convergence to a certain threshold), at the cost of only being forward differentiable

I have some code that does both of these things that I could polish enough for equinox to really boost the spectral norm flexibility (of course keeping all existing features, these are valuable in research, but in actual performance doing a poor approximation to a spectral norm actually often performs better than the real norm lol see sec 4 of http://www.arxiv.org/abs/2009.02773).

PR #831, opened 2024-09-06
Add exact spectral norm feature

lockwo:
Addresses part of https://github.com/patrick-kidger/equinox/issues/810.

### DIFF
--- a/equinox/nn/_spectral_norm.py
+++ b/equinox/nn/_spectral_norm.py
@@ -1,4 +1,4 @@
-from typing import Generic, Optional, TypeVar
+from typing import Callable, Generic, Optional, TypeVar
 
 import jax
 import jax.lax as lax
@@ -6,18 +6,31 @@ import jax.numpy as jnp
 import jax.random as jr
 from jaxtyping import Array, Float, PRNGKeyArray
 
+from .._eval_shape import filter_eval_shape
 from .._module import field
 from .._tree import tree_at
 from ._sequential import StatefulLayer
 from ._stateful import State, StateIndex
 
 
-def _power_iteration(weight, u, v, eps):
+def _power_iteration_old(weight, u, v, eps):
     u = weight @ v
     u_norm = jnp.sqrt(jnp.sum(u**2))
     u = u / jnp.maximum(eps, u_norm)
 
     v = weight.T @ u
+    v_norm = jnp.sqrt(jnp.sum(v**2))
+    v = v / jnp.maximum(eps, v_norm)
+
+    return u, v
+
+
+def _power_iteration(forward, transpose, v_prev, eps):
+    u = forward(v_prev)
+    u_norm = jnp.sqrt(jnp.sum(u**2))
+    u = u / jnp.maximum(eps, u_norm)
+
+    v = transpose(u)[0]
     v_norm = jnp.sqrt(jnp.sum(v**2))
     v = v / jnp.maximum(eps, v_norm)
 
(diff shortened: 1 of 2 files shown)


lockwo:
Open to discussion on best interface. The current caution around bias is ok but could be a footgun (although the original paper doing the reshaping is already kind of a footgun imo lol), but maybe I can apply the linear transpose around the whole layer pytree and update the bias too? I'm not sure that would 100% generalize to all layers though.

equinox/nn/_spectral_norm.py
```diff
@@ -1,18 +1,19 @@
-from typing import Generic, Optional, TypeVar
+from typing import Callable, Generic, Optional, TypeVar
```
patrick-kidger, line 1: `from collections.abc import Callable`. This one is deprecated.

equinox/nn/_spectral_norm.py
```diff
@@ -1,18 +1,19 @@
...
 
+from .._eval_shape import filter_eval_shape
 from .._module import field
 from .._tree import tree_at
 from ._sequential import StatefulLayer
 from ._stateful import State, StateIndex
 
 
-def _power_iteration(weight, u, v, eps):
+def _power_iteration_old(weight, u, v, eps):
```
patrick-kidger, line 16: This function should be removeable? We can implement it in terms of the new one.

equinox/nn/_spectral_norm.py
```diff
@@ -90,6 +118,13 @@ def __init__(
             The `dtype` of the weight array of the `layer` input is applied to all
             parameters in this layer.
 
+
+        !!! Caution
+
+            If `exact` is true, it computes the transpose via `jax.linear_transpose` of
+            the layer. This includes all operations of the layer call, which means for
+            layers with a bias, this can result in the incorrect spectral value.
```
patrick-kidger, line 126: Maybe we should follow JAX's lead here and transpose the tangent pass of `jax.jvp`?

equinox/nn/_spectral_norm.py
```diff
@@ -98,17 +133,32 @@ def __init__(
...
-        weight = jnp.reshape(weight, (weight.shape[0], -1))
-        dtype = weight.dtype
-        u_len, v_len = weight.shape
         ukey, vkey = jr.split(key)
-        u0 = jr.normal(ukey, (u_len,), dtype=dtype)
-        v0 = jr.normal(vkey, (v_len,), dtype=dtype)
-        for _ in range(15):
-            u0, v0 = _power_iteration(weight, u0, v0, eps)
+
+        if not exact:
```
patrick-kidger, line 138: Nit: remove the `not`, flip the branches.

equinox/nn/_spectral_norm.py
```diff
@@ -98,17 +133,32 @@ def __init__(
...
+            self.reverse = None
+            u0 = jr.normal(ukey, (u_len,), dtype=dtype)
+            v0 = jr.normal(vkey, (v_len,), dtype=dtype)
+            for _ in range(15):
+                u0, v0 = _power_iteration_old(weight, u0, v0, eps)
+        else:
+            assert (
+                input_shape is not None
+            ), "Must specify `input_shape` to use exact spectral norm!"
+            assert isinstance(self.layer, Callable)
```
patrick-kidger, line 153: Assertions should be used for internal errors. User errors should be `ValueError`s. (This is just a normal Python thing.)

Also `isinstance(x, Callable)` -> `callable(x)`.

equinox/nn/_spectral_norm.py
lockwo: mypy leads me astray once again ("note: Did you forget to import it from "typing"? (Suggestion: "from typing import Callable")")

equinox/nn/_spectral_norm.py
lockwo: Hmmm, I tried to implemented what I thought you meant. This also means we could remove the "weight" flag for the exact case (maybe?) since we basically "determine" the weight through the jvp? 

Let me know if this is what you had in mind, or if I was totally off. Does seem like a lot of jvps tho.

Issue #832, opened 2024-09-06
Equinox seems to break cooperative multiple inheritance


NeilGirdhar:
```python
from typing import override

import equinox as eqx
import jax.numpy as jnp
from jax import Array


class AcceptsStreams(eqx.Module):
    def __post_init__(self) -> None:
        if hasattr(super(), '__post_init__'):
            super().__post_init__()  # pyright: ignore


class InputNode(AcceptsStreams):
    total_value_error: Array = eqx.field(init=False)

    @override
    def __post_init__(self) -> None:
        print("PI InputNode")  # noqa: T201
        super().__post_init__()
        self.total_value_error = jnp.zeros(())


class DeductionSource(AcceptsStreams):
    pass


class DistillationBase(DeductionSource, AcceptsStreams):
    @override
    def __post_init__(self) -> None:
        print("PI DistillationBase")  # noqa: T201
        super().__post_init__()


class InputPerception(DistillationBase, InputNode):
    pass


print(", ".join(x.__qualname__ for x in InputPerception.__mro__))  # noqa: T201
# InputPerception, DistillationBase, DeductionSource, InputNode, AcceptsStreams, Module, object
InputPerception()
# ValueError: The following fields were not initialised during __init__: {'total_value_error'}
```
For some reason, `InputNode.__post_init__` is never called.  If `DeductionSource` is removed, then it is.

patrick-kidger:
Thankyou for the report! This should be fixed in #834.

FWIW I generally recommend against co-operative multiple inheritance (I never see this done correctly except in code I own entirely), so these days I generally prefer the [abstract/final pattern instead](https://docs.kidger.site/equinox/pattern/). IMO this is more readable + is certainly more robust. This is probably contributing to not having bumped into this edge-case before.

NeilGirdhar:
> MO this is more readable + is certainly more robust. 

That would mean that every concrete class would need to duplicate the initialization code and init-variables of its superclasses.  Besides the repeated code, it also breaks separation of concerns.

Personally, I think it's better to just use inheritance properly.  It's a shame that Python inheritance confuses people.

> I generally recommend against co-operative multiple inheritance (I never see this done correctly except in code I own entirely

It's pretty straightforward in Equinox.  You define `__post_init__` whenever you need it, and make sure that it call supers with all keyword parameters.

All `eqx.Module`s can have:
* ordinary variables,
* keyword-only `InitVar`s, and
[...]

PR #833, opened 2024-09-06
Fix __post_init__ detection in Module

NeilGirdhar:
Fixes https://github.com/patrick-kidger/equinox/issues/832


patrick-kidger:
Thank you for the fix here too, by the way! That's awesome.
I initially tried the same thing, and found it needed a bit more tweaking to also work with `eqx.field(..., converter=...)` -- #834 is otherwise very similar.

NeilGirdhar:
Aha, didn't see that you were working on a fix.  Thanks for the quick turnaround!

PR #836, opened 2024-09-07
fix: RoPE embeddings do not preserve dtypes

knyazer:
RoPE embeddings use complex as an underlying representation for the frequencies, which leads to it enforcing ```float32``` dtype when applying to any float type that is lower in the promotion chain than ```float32``` (e.g. ```float16```/```bfloat16```). This PR fixes the issue by adding a new keyword argument, ```dtype```, and using the frequencies with the correct ```dtype``` throughout the implementation.

### DIFF
--- a/equinox/nn/_embedding.py
+++ b/equinox/nn/_embedding.py
@@ -3,6 +3,7 @@ from typing import Any, Optional
 import jax
 import jax.numpy as jnp
 import jax.random as jrandom
+from jax._src.dtypes import TypePromotionError
 from jaxtyping import Array, ArrayLike, Float, Int, PRNGKeyArray
 
 from .._caches import cache_clears
@@ -162,11 +163,7 @@ class RotaryPositionalEmbedding(Module, strict=True):
 
     embedding_size: int = field(static=True)
     theta: float = field(static=True, default=10_000.0)
-    dtype: Any = field(static=True, default=None)
-
-    def __post_init__(self):
-        if self.dtype is None:
-            self.dtype = default_floating_dtype()
+    dtype: Any = field(static=True, default_factory=default_floating_dtype)
 
     def __check_init__(self):
         if self.embedding_size < 0:
@@ -191,6 +188,7 @@ class RotaryPositionalEmbedding(Module, strict=True):
         t = jnp.arange(float(end))
         freqs_outer = jnp.outer(t, freqs)
 
+        # we assign the type at the very end to minimize the loss of precision
         return jnp.cos(freqs_outer).astype(dtype), jnp.sin(freqs_outer).astype(dtype)
 
     @jax.named_scope("eqx.nn.RotaryPositionalEmbedding")
@@ -241,7 +239,18 @@ class RotaryPositionalEmbedding(Module, strict=True):
         freqs_sin = jnp.tile(freqs_sin, (1, 2))
 
         rotate_x = self.rotate_half(x)
-        x_rope = (x * freqs_cos) + (rotate_x * freqs_sin)
+        try:
+            x_rope = (x * freqs_cos) + (rotate_x * freqs_sin)
+        except TypePromotionError as e:
+            inp_dtype = jnp.dtype(x.dtype)
+            rope_dtype = jnp.dtype(self.dtype)
+            raise TypePromotionError(
+                f"The type of the passed value differs from the type "
+                f"of the rotary embeddings ({inp_dtype} != {rope_dtype}), thus leading "
+                "to a conflict when numpy_dtype_promotion is set to strict. To avoid "
+                f"this error, either initialiaze RoPE module with {inp_dtype} "
+                f"dtype, or explicitly cast the input argument to {rope_dtype}."
+            ) from e
         return x_rope.astype(x.dtype)
 
 
--- a/equinox/nn/_embedding.py
+++ b/equinox/nn/_embedding.py
@@ -177,7 +177,7 @@ class RotaryPositionalEmbedding(Module, strict=True):
     @staticmethod
     def precompute_freqs_cis(
         embedding_size: int, end: int, theta: float
-    ) -> Complex[Array, "end half_of_embedding_size"]:
+    ) -> tuple[Float[Array, "end half_of_embedding_size"], Float[Array, "end half_of_embedding_size"]]:
         freqs = 1.0 / (
             theta
             ** (jnp.arange(0.0, embedding_size, 2)[jnp.newaxis, :] / embedding_size)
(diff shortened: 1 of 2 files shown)


knyazer:
Btw, another issue: the order of parameters is annoyingly the opposite of the order in ```nn.Embedding```, but we cannot change it now because backwards compatibility :sob: 

Though I suppose that there is about one person who were using RoPE anyways (since transformers in JAX are pain anyways), so maybe it is not critical. Idk. WDYT?

equinox/nn/_embedding.py
```diff
@@ -1,17 +1,17 @@
-from typing import Optional
+from typing import Any, Optional
 
 import jax
 import jax.numpy as jnp
 import jax.random as jrandom
-from jaxtyping import Array, ArrayLike, Complex, Float, Int, PRNGKeyArray
+from jaxtyping import Array, ArrayLike, Float, Int, PRNGKeyArray
```
knyazer, line 6: Reason to remove usage of Complex: if we are trying to preserve/use certain dtypes, complex is just not expressive enough: e.g. there are no complex-bf16, so it is impossible to use it without a forceful type conversion.

knyazer:
@Artur-Galstyan Could you also take a look? It was your implementation, after all.

equinox/nn/_embedding.py
```diff
@@ -217,24 +223,29 @@ def __call__(
             )
 
         with jax.ensure_compile_time_eval():
-            if embedding_size in internal_rope_embedding_cache:
-                freqs_cis = internal_rope_embedding_cache[embedding_size]
-                freqs_cis_seq_len, _ = freqs_cis.shape
+            if (embedding_size, self.dtype) in internal_rope_embedding_cache:
+                freqs_pair = internal_rope_embedding_cache[(embedding_size, self.dtype)]
+                freqs_cis_seq_len, _ = freqs_pair[0].shape
```
patrick-kidger, line 228: NitL this could be done more neatly as `freqs_cos, freqs_sin = ...`. This would also save you needing to index a few lines lower down.

(This is one of my common bugbears, to prefer unpacking over indexing -- much harder for bugs to slip through this way.)

patrick-kidger (commented):
Nit aside this LGTM! Thank you for the fix..

(Side note, I'm not sure what the minimum useful precision for RoPE is? Given the high-frequency behaviour, I'm wondering if we should expect to always compute part of this in higher precision, e.g. like how softmax should use float32 at minimum.)

knyazer:
Alright, so I looked at my last comment, and I figured that I can just straight up convert the output to the old dtype, similarly how it is done in normalization modules. So now we avoid manual conversion.

Artur-Galstyan:
@knyazer it LGTM too. The reason I implemented this using complex numbers was because it was actually slightly faster (though not by much if I remember correctly - there should be somewhere a speed comparison table in one of the closed PRs). 

FYI, there is currently [a subtle bug](https://github.com/patrick-kidger/equinox/pull/799#issuecomment-2334764741) in the way it's currently implemented. I have a PR open #799 to fix that, but it's not 100% done yet. It'd be nice to get this merged so that I can rebase on it in my own PR.

knyazer:
Thanks for the feedback, @Artur-Galstyan! Sorry about your PR, I didn't check beforehand, so you will probably have a ton of conflicts :sweat_smile: 

About the speed: yea, I would have expected complex to be slightly faster for a bunch of reasons, but I'm pretty sure using a smaller dtype leads to a more significant performance gain.

equinox/nn/_embedding.py
```diff
@@ -162,6 +162,11 @@ def process_heads(
 
     embedding_size: int = field(static=True)
     theta: float = field(static=True, default=10_000.0)
+    dtype: Any = field(static=True, default=None)
+
+    def __post_init__(self):
+        if self.dtype is None:
+            self.dtype = default_floating_dtype()
```
patrick-kidger, line 169: I think this could be `dtype: Any = field(static=True, default_factory=default_floating_dtype)`.

equinox/nn/_embedding.py
```diff
@@ -176,19 +181,17 @@ def rotate_half(x: Float[Array, "seq_length embedding_size"]):
...
             ** (jnp.arange(0.0, embedding_size, 2)[jnp.newaxis, :] / embedding_size)
         )
 
         t = jnp.arange(float(end))
         freqs_outer = jnp.outer(t, freqs)
-        with jax.numpy_dtype_promotion("standard"):
-            freqs_cis = jnp.cos(freqs_outer) + jnp.sin(freqs_outer) * 1j
 
-        return freqs_cis
+        return jnp.cos(freqs_outer).astype(dtype), jnp.sin(freqs_outer).astype(dtype)
```
patrick-kidger, line 194: Might be worth including a comment here, to emphasise that the usual rule of "prefer `some_constructor(..., dtype=...)` over `.astype(dtype)`" does not apply.

equinox/nn/_embedding.py
```diff
@@ -217,35 +220,39 @@ def __call__(
...
-        freqs_real = jnp.tile(freqs_cis.real, (1, 2))
-        freqs_imag = jnp.tile(freqs_cis.imag, (1, 2))
+        freqs_cos = jnp.tile(freqs_cos, (1, 2))
+        freqs_sin = jnp.tile(freqs_sin, (1, 2))
 
         rotate_x = self.rotate_half(x)
-        x_rope = (x * freqs_real) + (rotate_x * freqs_imag)
-        return x_rope
+        x_rope = (x * freqs_cos) + (rotate_x * freqs_sin)
+        return x_rope.astype(x.dtype)
```
patrick-kidger, line 245: I think this should be `x_rope = (x * freqs_cos.astype(x.dtype)) + (rotate_x * freqs_sin.astype(x.dtype))` ?

Right now you seem to be working around the lack of this with `with jax.numpy_dtype_promotion("standard"):` in the test.

equinox/nn/_embedding.py
knyazer: You are right.... I didn't have a default factory in my mental model, but according to docs it is indeed equivalent to my code: 

_If a field is excluded from [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) (using init=False) and the field also specifies default_factory, then the default factory function will always be called from the generated __init__() function. This happens because there is no other way to give the field an initial value._

Anyhow, thanks!

equinox/nn/_embedding.py
knyazer: Thanks!

knyazer:
@patrick-kidger PTAL

patrick-kidger:
Awesome! Thanks for the fix, this LGTM. :)

PR #837, opened 2024-09-07
build: bump up precommit hooks

knyazer:
Also added ```toml``` filetype to ruff scanning, since ruff has some checks for the pyproject.toml structure: [RUF200](https://docs.astral.sh/ruff/rules/invalid-pyproject-toml/)

### DIFF
--- a/equinox/nn/_conv.py
+++ b/equinox/nn/_conv.py
@@ -39,22 +39,23 @@ def _padding_init(
 ) -> Union[str, tuple[tuple[int, int], ...]]:
     if isinstance(padding, str):
         padding = padding.upper()
-        if padding in ("SAME", "SAME_LOWER", "VALID"):
-            return padding
-        raise ValueError(
-            "`padding` string must be `'SAME'`, `'SAME_LOWER'`, or `'VALID'`."
-        )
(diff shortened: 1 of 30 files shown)


knyazer:
Do you think it makes sense to wait for https://github.com/google/jax/pull/23492 to get merged, so that we can use one less ```# pyright: ignore```?

equinox/nn/_conv.py
patrick-kidger, line 1: Are the changes here coming from RUFF? If so, why?
(I think I mildly prefer the earlier implementation -- with only one `return` I think the control flow is easier to understand / easier to check there's no accidental fallthrough behaviour.)

patrick-kidger (commented):
No strong feelings about the pyright-ignores, happy to merge without that.

equinox/nn/_conv.py
knyazer: Oops, actually this is an artefact from me trying to come up with a way to fix pyright type inference here... Sorry :)

knyazer:
Fixed the typos, while I'm on it :wink:

patrick-kidger:
This LGTM! Thank you for all of the (many, many) fixes.

Gosh, I like writing 'argments' instead of 'arguments'.

Issue #842, opened 2024-09-11
StateIndex is a Module, but not a PyTree


NeilGirdhar:
Is this a correct usage of stateful programming?
```python
import equinox as eqx
import jax.numpy as jnp
from jax import jit


class Model(eqx.Module):
    x: eqx.nn.StateIndex = eqx.field(init=False)

    def __post_init__(self) -> None:
        self.x = eqx.nn.StateIndex(jnp.zeros(()))


@jit
def f(m: Model, state: eqx.nn.State) -> None:
    pass


model, state = eqx.nn.make_with_state(Model)()
f(model, state)
```
If so, why isn't `StateIndex` a PyTree?

NeilGirdhar:
It seems like in a variety of places, Equinox is torn between making modules py-trees (with all static fields marked as static) versus leaving some static fields as dynamic to allow using `tree_at`, etc. to modify those fields.

This is seems confusing to me (at least for now).  Would it be possible to either add a flag to `tree_at` so that it crawls over some static fields, which can be marked `crawlable`?  That way, Equinox modules are always pytrees, and you still have the behavior you want.

patrick-kidger:
I believe `StateIndex` is a pytree. It just subclasses `eqx.Module` in the usual way.

Looking at your PR it seems like you're trying to avoid having non-arrays in its PyTree structure. I can see that that's a small QoL improvement, which seems reasonable to me.

On your latter point: note that it's not possible to have a module declare its static fields at class definition time. Consider for example `eqx.nn.MLP.activation`, which may be a 'static' `jax.nn.relu` or a 'dynamic' `eqx.nn.PReLU`.

WDYT?

NeilGirdhar:
> I believe `StateIndex` is a pytree. It just subclasses `eqx.Module` in the usual way.

Hmm, I think you're mistaken.  The `init` field is dynamic can be set `object()`, which is not a pytree.  Therefore, `StateIndex` is not a pytree.

> you're trying to avoid having non-arrays in its PyTree structure.

Not exactly.  I'm avoiding have non-pytrees in dynamic parameters.  Another alternative would be to use a sentinel that's a pytree.

> On your latter point: note that it's not possible to have a module declare its static fields at class definition time. Consider for example `eqx.nn.MLP.activation`, which may be a 'static' `jax.nn.relu` or a 'dynamic' `eqx.nn.PReLU`.

[...]

PR #843, opened 2024-09-11
Allow StateIndex to be passed dynamically

NeilGirdhar:
Fixes https://github.com/patrick-kidger/equinox/issues/842

### DIFF
--- a/equinox/nn/_stateful.py
+++ b/equinox/nn/_stateful.py
@@ -18,6 +18,17 @@ _P = ParamSpec("_P")
 _T = TypeVar("_T")
 
 
+class _Sentinel(Module):
+    """A module for sentinels that can be passed dynamically."""
+
+    pass
+
+
+# Used as a sentinel in two ways: keeping track of updated `State`s, and keeping track
+# of deleted initial states.
+_sentinel = _Sentinel()
+
+
 class StateIndex(Module, Generic[_Value], strict=True):
     """This wraps together (a) a unique dictionary key used for looking up a stateful
     value, and (b) how that stateful value should be initialised.
@@ -43,10 +54,10 @@ class StateIndex(Module, Generic[_Value], strict=True):
     [`equinox.nn.BatchNorm`][] for further reference.
     """  # noqa: E501
 
-    # Starts off as an `object` when initialised; later replaced with an `int` inside
+    # Starts off as None when initialised; later replaced with an `int` inside
     # `make_with_state`.
     marker: Union[object, int] = field(static=True)
-    init: _Value
+    init: Union[_Value, _Sentinel]
 
     def __init__(self, init: _Value):
         """**Arguments:**
@@ -68,11 +79,6 @@ class StateIndex(Module, Generic[_Value], strict=True):
 
 def _is_index(x: Any) -> bool:
     return isinstance(x, StateIndex)
-
-
-# Used as a sentinel in two ways: keeping track of updated `State`s, and keeping track
-# of deleted initial states.
-_sentinel = object()
 
 
 _state_error = """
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
If this is acceptable, I'm happy to add some methods if you want to make the code more polished:
```python
class StateIndex:
  def initial_value(self):
    return self.init[0]

  def initial_value_deleted(self):
     return self.init == ()
```

patrick-kidger (commented):
As you suggest in #842, I think it'd be better to replace `()` with a sentinel value that is a empty pytree, just to help emphasise what's going on.

I don't think you need to wrap the initial value into a length-1 tuple? Just the original value should work, I think.

NeilGirdhar:
Before you accept, let me change the other sentinel too?

NeilGirdhar:
Errors fixed, but I have some questions about the code.  I don't understand why `StateIndex.marker`'s sentinel can't be changed to `None`.  Is it because of replacement with `eqx.combine`, etc.?   Also, in `State.set`, why can't we assert that `item.marker` is an integer?  I guess I don't understand what the code is doing.

patrick-kidger:
Indeed, I think `.marker` being an `object()` might be unnecessary. I think there is a potential footgun here if such a 'raw' `StateIndex` is passed to `State` though -- with `object()` then we'd at least get a unique dictionary key, but with `None` then they'd all overwrite each other. Probably the appropriate solution is indeedto explicitly raise an error if such a 'raw' `StateIndex` is passed to `State` -- i.e. your suggestion of asserting that it is an integer.

NeilGirdhar:
> then we'd at least get a unique dictionary key, but with `None` then they'd all overwrite each other.

Interesting, okay.  You may want to consider adding a comment if that's behavior that you're counting on in for some use cases.  (Sorry I've been struggling with COVID all week, and my brain's a bit slower than usual.)

> i.e. your suggestion of asserting that it is an integer.

I tried that, but couldn't get it to pass the tests.

Anyway, I do love how the interface of `State` protects users from setting twice on the same state, or trying to use an expired state.  Nice user-facing design.

patrick-kidger:
> You may want to consider adding a comment if that's behavior that you're counting on in for some use cases.

Honestly, I'm not sure that'd be *good* behaviour to rely on... ! It's certainly not the usual path. I'd be happy to change that.

Lmk where you land on all of this + when you want a review of this PR. (Once it's passing tests.)

NeilGirdhar:
> Honestly, I'm not sure that'd be good behaviour to rely on... !

Great!  If I have time, I'll take a look at this again.

> Lmk where you land on all of this + when you want a review of this PR. (Once it's passing tests.)

Hmmm, it passes the tests on my machine on Python 3.11 (the failing test), and 3.12.  I'm not sure how to debug this.  Do you have any insight into this by any chance?

patrick-kidger:
It does seem a bit weird!
I do note that JAX recently did a new release, which has apparently since been yanked. Possibly something to do with that new release?

NeilGirdhar:
Could be, I tested with Jax 0.4.31.

But the error
```
Closure-converted function called with different dynamic arguments to the example arguments provided.
```
is related to Equinox, right?  I'm not sure how closure conversion works since I haven't used it yet.

NeilGirdhar:
(Looks like this passes now.)

NeilGirdhar:
@patrick-kidger Do you have time to take a look at this?

patrick-kidger:
Yup, I do! Have been otherwise engaged this past week. I expect to have a look at this in the next couple of days :)

NeilGirdhar:
No worries, take your time :)

patrick-kidger:
Okay, LGTM -- merged!  Thank you for the contribution :)

PR #845, opened 2024-09-12
fix link from gan to diffusion

garymm:


garymm:
Ah nevermind... the link works once converted to HTML but not browsing in GitHub. I'm guessing WAI

PR #846, opened 2024-09-12
Add detail to make_with_state doc.

garymm:


patrick-kidger:
Hmm, I don't think I'm a fan of this I'm afraid. `delete_init_state` is deliberately not documented -- `make_with_state` is the 'safe' public API for this.

garymm:
Thanks for taking a look.

For context I ran into this error and was confused because I had not called delete_init_state:

https://github.com/patrick-kidger/equinox/blob/97ac55ad2f83695bdbbfad6591df695235ca7928/equinox/nn/_stateful.py#L122

If you don't like this, perhaps some rewording of that error message then?

patrick-kidger:
Sounds reasonable! I'll do that now and close this.

PR #848, opened 2024-09-13
Add [test] and update CONTRIBUTING

dlwh:
Makes it easier to set up dev env.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Now make your changes. Make sure to include additional tests if necessary.
 Next verify the tests all pass:
 
 ```bash
-pip install pytest
+pip install -e .[test]
 pytest
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Now make your changes. Make sure to include additional tests if necessary.
 Next verify the tests all pass:
 
 ```bash
-pip install -e .[test]
+pip install -e ".[tests]"
 pytest
 ```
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -34,7 +34,7 @@ Now make your changes. Make sure to include additional tests if necessary.
 Next verify the tests all pass:
 
 ```bash
-pip install pytest
+pip install -e .[test]
```
patrick-kidger, line 37: I think this will need quote marks to work in most shells.

pyproject.toml
```diff
@@ -55,3 +55,10 @@ reportIncompatibleMethodOverride = true
 reportIncompatibleVariableOverride = false  # Incompatible with eqx.AbstractVar
 reportFunctionMemberAccess = false
 include = ["equinox", "tests"]
+
+[project.optional-dependencies]
+test = [
```
patrick-kidger, line 60: Nit: can we call this `tests` instead?

patrick-kidger (commented):
This LGTM! I've been meaning to convert this over for a while. Thank you for tweaking this :)

Could you also (a) deleted `tests/requirements.txt` and (b) update https://github.com/patrick-kidger/equinox/blob/main/.github/workflows/run_tests.yml appropriately?

dlwh:
PTAL

patrick-kidger:
LGTM! Thanks for the improvement :)

dlwh:
Of course, I should probably do it in my own projects...

patrick-kidger:
Unfortunately I had to revert this one. The problem arises with https://github.com/patrick-kidger/equinox/blob/5a40d2ed61871b400674ff3b3c6e8f5ba410d899/.github/workflows/release.yml#L19 that we don't want to install Equinox itself here. That's already being done for us as a source or wheel distribution. And in fact we *can't* install Equinox using `pip install .`, as e.g. its `pyproject.toml` file may not even be present in the wheel. (In practice the first error was Hatch complaining about the lack of `LICENSE` file.)

Ah well.

PR #851, opened 2024-09-14
specify mkdocs-autorefs version

garymm:


patrick-kidger:
I encountered this issue already, I think it should already be updated?

PR #852, opened 2024-09-14
Add overloads to eqx.combine

NeilGirdhar:
In the common case wherein all trees have the same structure, the return should have the same structure.  Helps type checking a lot.


patrick-kidger:

PR #868, opened 2024-10-01
Forward compatibility with future versions of JAX errors.

patrick-kidger:


hawkinsp:
Thanks for doing this! I was just about to send you a PR doing the same...

equinox/_jit.py
```diff
@@ -108,12 +109,20 @@ def _postprocess(out):
...
-except Exception:
-    # Unused dummy
-    class XlaRuntimeError(Exception):
-        pass
+    # Added in JAX 0.4.34.
+    JaxRuntimeError = jax.errors.JaxRuntimeError  # pyright: ignore[reportAttributeAccessIssue]
+except AttributeError:
+    try:
+        # Forward compatibility in case they ever decide to fix the capitalization.
+        JaxRuntimeError = jax.errors.JAXRuntimeError  # pyright: ignore[reportAttributeAccessIssue]
```
hawkinsp, line 117: Well, that would have been a helpful comment before I changed many users to this naming. I'm inclined to leave it as is.

patrick-kidger: Haha! 😄 I just spotted that the capitalisation differed as compared to the other JAX errors and erred on the side of caution. No strong opinions from me :)

PR #873, opened 2024-10-08
Add memorax link to readme

smorad:
Adds a link to a library of recurrent models using equinox


patrick-kidger:
Thanks for the suggestion! The libraries I recommend here are actually an intentionally short list: for the most part they are ones that I've used myself.

However, I can totally see that it would be reasonable to add an 'Ecosystem' page to the docs on https://docs.kidger.site/equinox. That could be a useful place for people such as yourself to record downstream projects. (We had an attempt an awesome-list a while ago -- @matteoguarrera's https://github.com/matteoguarrera/awesome-equinox -- but I didn't really do enough to help promote it at the time.) I'd be happy to take a PR that adds such a page instead, including memorax on it.

PR #874, opened 2024-10-08
chore(docs): notify about `tree_at` bypassing checks

### LINKED ISSUE #872: `tree_at` silently skips `__check_init__`
Hi!

I guess this is probably a "feature" of `tree_at`, but I recently discovered that it does not run `__check_init__`, which means it can bypass checks, leading to malformed data.

I could not find any reference to this in the documentation, so (1) is it possible to still have `__check_init__` run after some `tree_at` surgery, and (2) would it be possible to document that in the docstring of the `tree_at` function?

As always, thanks for your great tool :-)

jeertmans:
Hi @patrick-kidger!

Here is a first proposal. I am not entirely convinced by my wording, but let me know what you think of it.

Closes #872

### DIFF
--- a/equinox/_tree.py
+++ b/equinox/_tree.py
@@ -68,6 +68,13 @@ def tree_at(
     The modified PyTree is returned and the original input is left unchanged. Make sure
     to use the return value from this function!
 
+    !!! info
+
+        The nodes are modifieds without any call to `__init__` or `__check_init__`.
+        As a result, any check or post-processing (e.g., similarly to what `field(converter=...)`
+        could do) logic should be performed, by you, after calling `tree_at`.
+        
+
     **Arguments:**
 
     - `where`: A callable `PyTree -> Node` or `PyTree -> tuple[Node, ...]`. It should


patrick-kidger:
I think this looks pretty good! I think the only thing I'd tweak is to express is (a) that there are good reasons for this and (b) what those good reasons are. (Otherwise it just sounds like we decided not to implement this for some reason! :D )

jeertmans:
> I think this looks pretty good! I think the only thing I'd tweak is to express is (a) that there are good reasons for this and (b) what those good reasons are. (Otherwise it just sounds like we decided not to implement this for some reason! :D )

Sure, I didn’t want to write those reasons myself as I didn’t want to be incorrect, and I also wanted to be concise and didn’t know the proper wording to this end :)

patrick-kidger:
Fair enough! I've modified this in #877 :)

jeertmans:
Nice, I guess I can close this now, I am leaving a few comments on the other PR :-)

PR #876, opened 2024-10-12
Fix computation in normalization layers

AakashKumarNain:
Irrespective of the dtype passed to the normalization layers, the calculations should be done with fp32 or higher dtype. Not doing that can cause instabilities in training which is too common when someone is training models in half-precision. Now, we can delegate this to the end user to do this explicitly, but it is an easy thing to miss. This PR takes care of that.

### DIFF
--- a/equinox/nn/_normalisation.py
+++ b/equinox/nn/_normalisation.py
@@ -140,6 +140,11 @@ class LayerNorm(Module, strict=True):
                 "`x.shape` ended with `shape`. However, this turned out to be a "
                 "frequent source of bugs, so we made the check stricter!"
             )
+        orig_dtype = x.dtype
+        with jax.numpy_dtype_promotion("standard"):
+            dtype = jnp.result_type(x.dtype, jnp.float32)
+
+        x = x.astype(dtype)
         mean = jnp.mean(x, keepdims=True)
         variance = jnp.var(x, keepdims=True)
         variance = jnp.maximum(0.0, variance)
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Looks like the current implementation doesn't pass under strict-dtype-promotion (see failing tests), but other than that this sounds good to me!

AakashKumarNain:
Thanks @patrick-kidger I have fixed it. Please let me know if you have any more suggestions on this one

equinox/nn/_normalisation.py
```diff
@@ -94,6 +94,10 @@ def __init__(
             )
         self.use_weight = use_weight
         self.use_bias = use_bias
+
+        with jax.numpy_dtype_promotion("standard"):
+            dtype = jnp.result_type(dtype, jnp.float32)
```
patrick-kidger, line 99: This one seems wrong to me. It's totally valid for a user to want to set the weight and bias to a lower dtype, if they really want to.

AakashKumarNain: Makes sense. Changed it.

patrick-kidger:
Alright, this LGTM! Final question before I merge this, do you have a reference for needing higher precision for these operations? (As much for me to read up on as anything else :D )

AakashKumarNain:
I think this is pretty much enough: https://pytorch.org/docs/stable/amp.html#autocast-op-reference

patrick-kidger:
Alright, looks good to me! Thanks for spotting this, and for the reference. Merged :)

PR #877, opened 2024-10-12
Improves docs for `tree_at` and `tree_check`.

### LINKED ISSUE #872: `tree_at` silently skips `__check_init__`

patrick-kidger:
Supersedes #874.
Fixes #872.

### DIFF
--- a/equinox/_tree.py
+++ b/equinox/_tree.py
@@ -116,6 +116,18 @@ def tree_at(
         new_mlp = eqx.tree_at(get_last_layer, mlp, new_linear)
         ```
         See also the [Tricks](../../tricks) page.
+
+    !!! Info
+
+        Constructing analogous PyTrees, with the same structure but different leaves, is
+        very common in JAX: for example when constructing the `in_axes` argument to
+        `jax.vmap`.
+
+        To support this use-case, the returned PyTree is constructed without calling
+        `__init__`, `__post_init__`, or
+        [`__check_init__`](./module/advanced_fields.md#checking-invariants). This allows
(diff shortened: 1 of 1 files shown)


equinox/_tree.py
```diff
@@ -406,15 +418,15 @@ def is_leaf(node):
...
 
 def tree_check(pytree: Any) -> None:
-    """Checks if the PyTree is well-formed: does it have no self-references, and does
-    it have no duplicate layers.
-
-    Precisely, a "duplicate layer" is any PyTree node with at least one child node.
-
-    !!! info
+    """Checks if the PyTree has no self-references, and if all non-leaf nodes unique
+    Python objects. (For example something like `x = [1]; y = [x, x]` would fail as `x`
```
jeertmans, line 422:
```suggestion
    """Checks if the PyTree has no self-references, and if all non-leaf nodes are unique
    Python objects. (For example something like `x = [1]; y = [x, x]` would fail as `x`
```

I think you missed a word here, no?

jeertmans (commented):
I think this is great! I just spot a possible typo.

patrick-kidger:
Thanks for the typo catch! Fixed, and merged. :)

jeertmans:
Thanks!

PR #878, opened 2024-10-12
Allow Python versions higher than 3.9

norpadon:
It is currently impossible to add equinox to a modern Python project using Poetry, because the current pyproject.toml says that it only supports Python 3.9

This PR fixes this

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "equinox"
 version = "0.11.7"
 description = "Elegant easy-to-use neural networks in JAX."
 readme = "README.md"
-requires-python ="~=3.9"
+requires-python =">=3.9"
 license = {file = "LICENSE"}
 authors = [
   {name = "Patrick Kidger", email = "contact@kidger.site"},


patrick-kidger:
I think this is correct as-written. `~=3.9` means that anything in the `3.x` series above `3.9` is acceptable.

See the PyPA documentation here: https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release

For some reason Poetry appears to use a nonstandard interpretation of this: https://python-poetry.org/docs/dependency-specification/#tilde-requirements

If my understanding is correct then I believe this is a bug in Poetry, and it should be reported there. (A quick search on their GitHub looks like this might have come up before: https://github.com/python-poetry/poetry/issues/380)

(These days I would strongly recommend using uv instead. Similar to poetry, fewer bugs, much faster.)

norpadon:
Oh, I think you are correct. The problem was on my side: I was specifying requirements as
```
python = ">=3.12"
```
But it theoretically includes `python = "4.0"`, which is technically not compatible with `python = "~=3.9"`.
So I fixed the problem on my side by changing python version requirements to "~=3.12"

This is indeed a technically correct behaviour, but I am not sure whether being "technically correct" is worth it, given that the likelihood of python4.x being released is negligibly small. Every other python package has ">=" in python version requirements, and equinox is the first package I had this issue with. Maybe switching to ">=3.9" on the equinox side is simply better for the user experience.

patrick-kidger:
Agreed. Whilst Poetry is definitely wrong here, it's a pragmatic choice for us to allow compatibility in this way. Merged! :)

PR #915, opened 2024-12-18
Upgrade jax and fix deprecation warnings

DrJessop:
Jax v0.4.38 deprecated many APIs in jax.core, and can be replaced instead by jax.extend.core.
Jax v0.4.38 also required me to upgrade python.

Fixes: https://github.com/patrick-kidger/equinox/issues/913

### DIFF
--- a/tests/test_primitive.py
+++ b/tests/test_primitive.py
@@ -4,6 +4,7 @@ import equinox as eqx
 import equinox.internal as eqxi
 import jax
 import jax.core
+import jax.extend.core
 import jax.interpreters.ad as ad
 import jax.interpreters.batching as batching
 import jax.interpreters.mlir as mlir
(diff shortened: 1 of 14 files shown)


tests/test_primitive.py
```diff
@@ -4,6 +4,7 @@
 import equinox.internal as eqxi
 import jax
 import jax.core
```
patrick-kidger, line 6: Thank you for the fix!

By the way, do we need to keep all of the `import jax.core` around? I'm guessing most of these can be removed.

DrJessop: Done

garymm: Seems you failed to push after removing the unused imports?

DrJessop: @garymm No, a fair amount of files still need jax.core.

patrick-kidger:
Alright, LGTM! Thank you for the upgrades :)

PR #924, opened 2025-01-05
Now using wadler_lindig pprint library.

patrick-kidger:


patrick-kidger:
@jakevdp -- [idk if Equinox is still in g3](https://github.com/patrick-kidger/equinox/pull/769#issuecomment-2187011069), but if so then heads-up that after this PR, it will have a [new (tiny) dependency](https://github.com/patrick-kidger/wadler_lindig).

(And you'll be pleased to know that this also cuts one of our dependencies on `jax._src`.)

PR #941, opened 2025-02-03
Remove parameters from discriminator optimizer update in DCGAN example

struan-robertson:
Hi,

Firstly, thanks for creating Equinox! Learning how to use it has been a lot of fun.

I noticed in your example of the DCGAN that when calling discriminator_optimizer.update the discriminator PyTree is passed as the optional params argument. This confused me, and after looking into it I think that this has no effect as the `adam` optimizer is being used. The reason I am creating this pull request, is that if the optimizer is changed to one that does require model parameters, such as `adamw`, passing the discriminator as is fails silently. When using such an optimizer, I had success using:
```python
updates, opt_state = discriminator_optimizer.update(
    grads, discriminator_opt_state, eqx.filter(discriminator, eqx.is_array),
)
```

So the current example of updating the discriminator optimizer is a bit misleading.


patrick-kidger:
LGTM! Agreed. Good catch, and thank you for the doc fix :)

PR #948, opened 2025-02-10
Batch BatchNorm

lockwo:
Revives https://github.com/patrick-kidger/equinox/pull/675, but ideally a smaller/simpler change (that doesn't require any math) that matches the patterns in flax/haiku (see: https://github.com/google-deepmind/dm-haiku/blob/main/haiku/_src/batch_norm.py#L42%23L206, https://github.com/google-deepmind/dm-haiku/blob/main/haiku/_src/moving_averages.py#L41%23L139). Has some hardcoded defaults (e.g. I don't support warmup iterations at all), but should help the users of batch norm. In addition to the stability in the initial example, here is an example on an AlphaZero model playing 9x9 Go: 

Screenshot 2025-02-09 at 9 37 30 PM

This screenshot (and code) also comes from this PR: https://github.com/sotetsuk/pgx/pull/1300

### DIFF
--- a/equinox/nn/_batch_norm.py
+++ b/equinox/nn/_batch_norm.py
@@ -1,9 +1,11 @@
+import warnings
 from collections.abc import Hashable, Sequence
+from typing import Literal
 
 import jax
 import jax.lax as lax
 import jax.numpy as jnp
-from jaxtyping import Array, Bool, Float, PRNGKeyArray
+from jaxtyping import Array, Bool, Float, Int, PRNGKeyArray
 
 from .._misc import default_floating_dtype
 from .._module import field
@@ -40,25 +42,70 @@ class BatchNorm(StatefulLayer, strict=True):
     statistics updated. During inference then just the running statistics are used.
     Whether the model is in training or inference mode should be toggled using
     [`equinox.nn.inference_mode`][].
+
+    With `mode = "batch"` during training the batch mean and variance are used
+    for normalization.  For inference the exponential running mean and unbiased
+    variance are used for normalization. This is in line with how other machine
+    learning packages (e.g. PyTorch, flax, haiku) implement batch norm.
+
+    With `mode = "ema"` exponential running means and variances are kept.  During
+    training the batch statistics are used to fill in the running statistics until
+    they are populated.  During inference the running statistics are used for
+    normalization.
+
+    ??? cite
+
+        [Batch Normalization: Accelerating Deep Network Training by Reducing
+         Internal Covariate Shift](https://arxiv.org/abs/1502.03167)
+
+        ```bibtex
+        @article{DBLP:journals/corr/IoffeS15,
+        author       = {Sergey Ioffe and
+                        Christian Szegedy},
+        title        = {Batch Normalization: Accelerating Deep Network Training
+                        by Reducing Internal Covariate Shift},
+        journal      = {CoRR},
+        volume       = {abs/1502.03167},
+        year         = {2015},
+        url          = {http://arxiv.org/abs/1502.03167},
+        eprinttype    = {arXiv},
+        eprint       = {1502.03167},
+        timestamp    = {Mon, 13 Aug 2018 16:47:06 +0200},
+        biburl       = {https://dblp.org/rec/journals/corr/IoffeS15.bib},
+        bibsource    = {dblp computer science bibliography, https://dblp.org}
+        }
+        ```
     """  # noqa: E501
 
     weight: Float[Array, "input_size"] | None
     bias: Float[Array, "input_size"] | None
-    first_time_index: StateIndex[Bool[Array, ""]]
-    state_index: StateIndex[
-        tuple[Float[Array, "input_size"], Float[Array, "input_size"]]
-    ]
+    ema_first_time_index: None | StateIndex[Bool[Array, ""]]
+    ema_state_index: (
+        None | StateIndex[tuple[Float[Array, "input_size"], Float[Array, "input_size"]]]
+    )
+    batch_counter: None | StateIndex[Int[Array, ""]]
+    batch_state_index: (
+        None
+        | StateIndex[
+            tuple[
+                tuple[Float[Array, "input_size"], Float[Array, "input_size"]],
+                tuple[Float[Array, "input_size"], Float[Array, "input_size"]],
+            ],
+        ]
+    )
     axis_name: Hashable | Sequence[Hashable]
     inference: bool
     input_size: int = field(static=True)
     eps: float = field(static=True)
     channelwise_affine: bool = field(static=True)
     momentum: float = field(static=True)
+    mode: Literal["ema", "batch"] = field(static=True)
 
     def __init__(
         self,
         input_size: int,
         axis_name: Hashable | Sequence[Hashable],
+        mode: Literal["ema", "batch", "legacy"] = "legacy",
         eps: float = 1e-5,
         channelwise_affine: bool = True,
         momentum: float = 0.99,
@@ -71,6 +118,7 @@ class BatchNorm(StatefulLayer, strict=True):
         - `axis_name`: The name of the batch axis to compute statistics over, as passed
             to `axis_name` in `jax.vmap` or `jax.pmap`. Can also be a sequence (e.g. a
             tuple or a list) of names, to compute statistics over multiple named axes.
+        - `mode`: The variant of batch norm to use, either 'ema' or 'batch'.
         - `eps`: Value added to the denominator for numerical stability.
         - `channelwise_affine`: Whether the module has learnable channel-wise affine
             parameters.
@@ -86,6 +134,16 @@ class BatchNorm(StatefulLayer, strict=True):
             `jax.numpy.float32` or `jax.numpy.float64` depending on whether JAX is in
             64-bit mode.
         """
+        if mode == "legacy":
+            mode = "ema"
+            warnings.warn(
+                "When mode is unspecified it defaults to 'ema'. This can have "
+                "substantial performance impacts, and the user is encouraged to "
+                "consider and pick which mode they need."
+            )
+        if mode not in ("ema", "batch"):
+            raise ValueError("Invalid mode, must be 'ema' or 'batch'.")
+        self.mode = mode
         dtype = default_floating_dtype() if dtype is None else dtype
         if channelwise_affine:
             self.weight = jnp.ones((input_size,), dtype=dtype)
@@ -93,12 +151,28 @@ class BatchNorm(StatefulLayer, strict=True):
         else:
             self.weight = None
             self.bias = None
-        self.first_time_index = StateIndex(jnp.array(True))
-        init_buffers = (
-            jnp.empty((input_size,), dtype=dtype),
-            jnp.empty((input_size,), dtype=dtype),
-        )
-        self.state_index = StateIndex(init_buffers)
+        if mode == "ema":
+            self.ema_first_time_index = StateIndex(jnp.array(True))
+            init_buffers = (
+                jnp.empty((input_size,), dtype=dtype),
+                jnp.empty((input_size,), dtype=dtype),
+            )
+            self.ema_state_index = StateIndex(init_buffers)
+            self.batch_counter = None
+            self.batch_state_index = None
+        else:
+            self.batch_counter = StateIndex(jnp.array(0))
+            init_hidden = (
+                jnp.zeros((input_size,), dtype=dtype),
+                jnp.ones((input_size,), dtype=dtype),
+            )
+            init_avg = (
+                jnp.zeros((input_size,), dtype=dtype),
+                jnp.ones((input_size,), dtype=dtype),
+            )
+            self.batch_state_index = StateIndex((init_hidden, init_avg))
+            self.ema_first_time_index = None
+            self.ema_state_index = None
         self.inference = inference
         self.axis_name = axis_name
         self.input_size = input_size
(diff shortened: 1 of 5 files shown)


lockwo:
Also, for anyone who wants to try out the comparisons, here's some example code that shows haiku vs equinox 

<details>
<summary> code </summary>

```python
import jax
from jax import numpy as jnp
import equinox as eqx
import haiku as hk


class Net(hk.Module):

    def __init__(
        self,
        name="net",
    ):
        super().__init__(name=name)

    def __call__(self, x, is_training):
        x = x.astype(jnp.float32)
        x = hk.BatchNorm(True, True, 0.9)(x, is_training)
        return x


class Neteqx(eqx.Module):
    norm: eqx.nn.BatchNorm

    def __init__(
        self,
        output_channels: int = 64,
    ):
        self.norm = eqx.nn.BatchNorm(
            output_channels, "batch", momentum=0.9, mode="eqx1"
        )

    def __call__(self, x, state):
        x = x.astype(jnp.float32)
        x = jnp.moveaxis(x, -1, 0)
        x, state = self.norm(x, state)
        x = jnp.moveaxis(x, 0, -1)
        return x, state


def forward_fn(x):
    net = Net()
    v = net(x, True)
    return v


def inf_fn(x):
    net = Net()
    v = net(x, False)
    return v


forward = hk.without_apply_rng(hk.transform_with_state(forward_fn))
inf = hk.without_apply_rng(hk.transform_with_state(inf_fn))

ins = jnp.array([1, 2, 3, 4]).reshape((4, 1, 1, 1)) * jnp.ones((4, 2, 2, 3))
params, state = forward.init(jax.random.key(0), jnp.zeros_like(ins))
# print(params)
# print(hk.data_structures.tree_size(params))
print(state)
jax.tree.map(lambda x: print(x), state)

for i in range(2):
    out, state = forward.apply(params, state, ins[2 * i : 2 * (i + 1)])
    jax.tree.map(lambda x: print(x), state)

inf.apply(params, state, ins)

eq, s = eqx.nn.make_with_state(Neteqx)(3)
eq = eqx.tree_at(
    lambda x: x.norm.weight, eq, params["net/batch_norm"]["scale"].squeeze()
)
eq = eqx.tree_at(
    lambda x: x.norm.bias, eq, params["net/batch_norm"]["offset"].squeeze()
)

# print(sum(x.size for x in jax.tree.leaves(eqx.filter(eq, eqx.is_inexact_array))))
jax.tree.map(lambda x: print(x), s)

for i in range(2):
    out_eqx, s = eqx.filter_vmap(
        eq, in_axes=(0, None), out_axes=(0, None), axis_name="batch"
    )(ins[2 * i : 2 * (i + 1)], s)
    jax.tree.map(lambda x: print(x), s)
    # print(out_eqx)

eqx.filter_vmap(
    eqx.nn.inference_mode(eq), in_axes=(0, None), out_axes=(0, None), axis_name="batch"
)(ins, s)
```
</details>

ZagButNoZig:
Ok so I don't know how useful this is, but I validated this implementation against the old ema impl and PyTorch's batch norm and `mode="ema"` has exactly the same output as the old version and `mode="batch"` is withing `10e-6` of the PyTorch implementation, which is exactly what we want?

Here is the script if anyone wants to reproduce it: 

```Python
import equinox as eqx
import jax.numpy as jnp
import jax
import numpy as np
import torch
import old_bn

N = 16
m = 0.1
mode = "ema"

bn_t = torch.nn.BatchNorm1d(N, momentum=m)
bn_e, s = eqx.nn.make_with_state(eqx.nn.BatchNorm)(input_size=N, axis_name="b", mode=mode, momentum=(1-m))
bn_eo, s2 = eqx.nn.make_with_state(old_bn.BatchNorm)(input_size=N, axis_name="b", momentum=(1-m))

for _ in  range(50):
    x = np.random.rand(N*N).reshape((N, N)).astype(dtype=np.float32)

    x_t = torch.tensor(x)
    x_j = jnp.array(x)
    
    o_t = bn_t(x_t)
    o_j, s = eqx.filter_vmap(bn_e, axis_name="b", in_axes=(0, None), out_axes=(0, None))(x_j, s)
    o_eo, s2 = eqx.filter_vmap(bn_eo, axis_name="b", in_axes=(0, None), out_axes=(0, None))(x_j, s2)

    o_t = np.array(o_t.detach().cpu())
    o_j = np.array(o_j)
    o_eo = np.array(o_eo)

    delta = np.abs(o_t - o_j)
    print('Torch vs. New: ', delta.max())
    
    delta = np.abs(o_j - o_eo)
    print('New vs. Old: ',delta.max())
```
Where  `old_bn.BatchNorm` is BatchNorm@main.

equinox/nn/_batch_norm.py
```diff
@@ -40,25 +41,69 @@ class BatchNorm(StatefulLayer, strict=True):
     statistics updated. During inference then just the running statistics are used.
     Whether the model is in training or inference mode should be toggled using
     [`equinox.nn.inference_mode`][].
+
+    With `mode = "batch"` during training the batch mean and variance are used
+    for normalization.  For inference the exponential running mean and ubiased
```
patrick-kidger, line 46: Typo: unbiased

equinox/nn/_batch_norm.py
```diff
@@ -40,25 +41,69 @@ class BatchNorm(StatefulLayer, strict=True):
...
+    variance are used for normalization. This is in line with out other JAX
```
patrick-kidger, line 47: Typo.

I'd probably also reference PyTorch instead, which is actually our usual point of reference for Equinox.

equinox/nn/_batch_norm.py
```diff
@@ -40,25 +41,69 @@ class BatchNorm(StatefulLayer, strict=True):
...
+        }
+        ```
     """  # noqa: E501
 
     weight: Float[Array, "input_size"] | None
     bias: Float[Array, "input_size"] | None
-    first_time_index: StateIndex[Bool[Array, ""]]
-    state_index: StateIndex[
-        tuple[Float[Array, "input_size"], Float[Array, "input_size"]]
+    ema_first_time_index: Optional[StateIndex[Bool[Array, ""]]]
```
patrick-kidger, line 81: Can you switch these over to `None | ...`?

equinox/nn/_batch_norm.py
```diff
@@ -40,25 +41,69 @@ class BatchNorm(StatefulLayer, strict=True):
...
     eps: float = field(static=True)
     channelwise_affine: bool = field(static=True)
     momentum: float = field(static=True)
+    mode: Literal["ema", "batch"] = field(static=True)
 
     def __init__(
         self,
         input_size: int,
         axis_name: Hashable | Sequence[Hashable],
+        mode: str = "ema",
```
patrick-kidger, line 106: Should we make the default `"legacy"` -- which emits a warning and then sets `"ema"` -- to encourage picking explicitly?


Also the type annotation here could be a Literal.

equinox/nn/_batch_norm.py
```diff
@@ -138,38 +203,85 @@ def __call__(
...
+                running_mean, running_var = state.get(self.ema_state_index)
+                momentum = self.momentum
+                running_mean = (1 - momentum) * batch_mean + momentum * running_mean
+                running_var = (1 - momentum) * batch_var + momentum * running_var
+                # since jnp.array(0) == False
+                running_mean = lax.select(first_time, batch_mean, running_mean)
+                running_var = lax.select(first_time, batch_var, running_var)
+                state = state.set(self.ema_state_index, (running_mean, running_var))
+
+            def _norm(y, m, v, w, b):
```
patrick-kidger, line 238: Am I misreading anything, or are the two copies of this function the same, so that it can be factored out above the `if` statement?

patrick-kidger (commented):
Okay gosh, I know this one has languished for a long time.

I think this basically LGTM! I've left some minor comments but I think I'm happy to merge this.

equinox/nn/_batch_norm.py
lockwo: Is this the new standard? I know it's allowed with PEP 604, but is it now the standard to do | over Union?

equinox/nn/_batch_norm.py
lockwo: fixed

lockwo:
```
The brownout dates and times are as follows:

April 1, 2025, 3 p.m. – 7 p.m. UTC
```

:(

equinox/nn/_batch_norm.py
patrick-kidger: This syntax became available in Python 3.10. I've now switched over Equinox to using this as it's now the minimum version we support.

ZagButNoZig:
Don't mean to be pushy, but are there any blockers left on this one?

Thanks again @lockwo for the amazing work!

equinox/nn/_batch_norm.py
```diff
@@ -138,38 +212,79 @@ def __call__(
...
+        else:
+            assert self.batch_state_index is not None and self.batch_counter is not None
+            if inference:
+                _, (mean, var) = state.get(self.batch_state_index)
+            else:
+                batch_mean, batch_var = jax.vmap(_stats)(x)
+                counter = state.get(self.batch_counter)
+                (hidden_mean, hidden_var), (running_mean, running_var) = state.get(
+                    self.batch_state_index
+                )
```
patrick-kidger, line 264: IIUC then `running_mean` and `running_var` don't actually need to be in the state -- they're never consumes on the `inference=False` branch. 

I think we could afford to have just the `hidden_mean, hidden_var` in the state, and then do the zero-debiasing on the `inference=True` branch? This would save on both memory and compute.

WDYT?

patrick-kidger (commented):
Fortunately this is on my list of PRs to review today :)

I have just one question, where I might be missing something?

equinox/nn/_batch_norm.py
ZagButNoZig: If your use case is training heavy then yes. But if you mostly run inference then this adds extra compute as we have to recompute the `running` data every time?

patrick-kidger: So for inference then it's already pretty common to want additional optimisations: for example, fusing the weights/biases of batch norm with the parameters of adjacent linear layers.

I think this should probably come under the same banner. Ideally such behaviour should be obtained via constant propagation in the XLA compiler (or similar).

Exactly how this is usually done in practice I've not dug into, as I don't typically run inference-heavy workloads, but ideally we should provide something that *can* be optimized to do that, rather than the current implementation which unconditionally takes the slower training path?

ZagButNoZig: Ah right. Makes sense, thank you!

patrick-kidger:
Okay, so following on the from above discussion, I've gone ahead and adjusted the above. (I'm sympathetic that you must be a bit tired of this PR at this point!) I've:

- Rebased on the new `dev` and squashed all your commits together.
- Updated the zero scaling to happen at inference time, rather than training time.
- Performed various stylistic simplifications.

@lockwo @ZagButNoZig take a look and let me know what you think?

I've also tried running an adapted version of @ZagButNoZig's script, to test agreement between PyTorch and our `mode="batch"`, and between our `mode="batch"` and the version of `BatchNorm` currently available on `main`, and get agreement throughout.

<details><summary><h3>Click for code</h3></summary>

```python
import equinox as eqx
import equinox.nn.old_bn
import jax.numpy as jnp
import numpy as np
import torch


N = 16
m = 0.1

bn_torch = torch.nn.BatchNorm1d(N, momentum=m)
bn_batch, s_batch = eqx.nn.make_with_state(eqx.nn.BatchNorm)(
    input_size=N, axis_name="b", mode="batch", momentum=(1 - m)
)
bn_ema, s_ema = eqx.nn.make_with_state(eqx.nn.BatchNorm)(
    input_size=N, axis_name="b", mode="ema", momentum=(1 - m)
)
bn_old, s_old = eqx.nn.make_with_state(equinox.nn.old_bn.BatchNorm)(
    input_size=N, axis_name="b", momentum=(1 - m)
)

for _ in range(50):
    x = np.random.rand(N * N).reshape((N, N)).astype(dtype=np.float32)

    x_t = torch.tensor(x)
    x_j = jnp.array(x)

    o_torch = bn_torch(x_t)
    o_batch, s_batch = eqx.filter_vmap(
        bn_batch, axis_name="b", in_axes=(0, None), out_axes=(0, None)
    )(x_j, s_batch)
    o_ema, s_ema = eqx.filter_vmap(
        bn_ema, axis_name="b", in_axes=(0, None), out_axes=(0, None)
    )(x_j, s_ema)
    o_old, s_old = eqx.filter_vmap(
        bn_old, axis_name="b", in_axes=(0, None), out_axes=(0, None)
    )(x_j, s_old)

    o_torch = np.array(o_torch.detach().cpu())
    o_batch = np.array(o_batch)
    o_ema = np.array(o_ema)
    o_old = np.array(o_old)

    delta = np.abs(o_torch - o_batch)
    print("Torch vs. batch: ", delta.max())

    delta = np.abs(o_ema - o_old)
    print("ema vs. old: ", delta.max())
```

</details>

lockwo:
Yea I think that all looks good. Moving to inference I think is totally reasonable (and cleaning things up is always good, like deleting key and what not). I just made a minor tweak to update one of the comments to match the language

ZagButNoZig:
Awesome! LGTM :)

patrick-kidger:
Awesome, this LGTM then, so merged! Thank you for your help both of you :)

PR #950, opened 2025-02-14
Allow zero shapes in nn.Linear

aseyboldt:
`math.sqrt(0)` throws an exception, which prevents `eqx.nn.Linear(0, 4)` from working, even though it is well defined.


patrick-kidger:
LGTM! Thank you for the edge case fix :)

PR #953, opened 2025-02-20
Add AlexNet + CLU example

Artur-Galstyan:
Hi there! 

I was in the middle of creating a tutorial on implementing AlexNet in JAX and I thought it might be useful to add that as an introductory example. It differs from the CNN example by using `tensorflow_datasets` and also showcases how to use `CommonLoopUtils` with Eqx. 

As a side note: do you think adding the `LayerResponseNormalisation` to the API makes sense? On the one hand, it doesn't hurt and perhaps some people would find that useful, on the other hand, it's been mostly overshadowed by `BatchNorm`. WDYT?


mkdocs.yml
Artur-Galstyan, line 1: Whoops, seems like my editor had some strong opinions on formatting. The only change I made was on line 106.

patrick-kidger:
Thanks for putting this together!

So I think this might actually fit best as a member of our new [awesome-list](https://docs.kidger.site/equinox/awesome-list/), by linking to the tutorial you're putting together. I'm not completely sold on either tfd or clu, so for our hosted examples I try to avoid using them. WDYT?

> As a side note: do you think adding the LayerResponseNormalisation to the API makes sense? 

I think I'm inclined not to include it. Every extra API surface is an extra thing to maintain, and as you note this one isn't really interesting to a broad audience :)

Artur-Galstyan:
I'm not so sure if it's a **good** fit to the list because it's actually a couple of YouTube tutorials that I made/am making. But if you think YT tutorials are ok, then I'll make a PR with a link to the tutorials once they are done :) In that case, feel free to close this PR.

As for CLU: When I noticed that I was literally writing the same code for the metrics over and over again I started using CLU and I've grown to like it, which is why I wanted to share it xd 

Regarding the LRN part: you're right! One idea I just had is creating a repo called something like `equinox_extended` or something where I'd dump all these layers that I'm writing (most of those are "clones" from Pytorch just in JAX style). That way Equinox core can remain svelte if people need some extra fancy layers or something, they could use the other repo :D Now that would be a good fit for the awesome-list! 😆

patrick-kidger:
The YT / extended stuff both sound reasonable! Send PRs with links once you have them :)

Issue #973, opened 2025-03-13
Overhead of filter_jit


ZagButNoZig:
Hey, 

I have noticed that methods annotated with `@eqx.filter_jit` have an additional overhead compared to `@jax.jit`. I am using this to compare the times: 

```Python
import equinox as eqx
import jax
import jax.numpy as jnp


model = eqx.nn.Identity()
x = jnp.zeros((1,1))

@eqx.filter_jit
def fwd_eqx(m, x):
    return m(x)

@jax.jit
def fwd_jax(m, x):
    return m(x)

# Make sure everything is jit compiled
for _ in range(10):
    fwd_eqx(model, x)
    fwd_jax(model, x)
```

Which gives me:

```Python
%timeit fwd_jax(model, x).block_until_ready()
53.2 μs ± 744 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

```Python
%timeit fwd_eqx(model, x).block_until_ready()
423 μs ± 4.07 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

I also traced 3000 calls to `fwd_eqx` using Pyinstrument: 

![Image](https://github.com/user-attachments/assets/1fa2f2bf-26ca-49a1-9224-1002002fedf3)

# Questions 

- Is what I'm measuring actually equinox related?
- Is this overhead relevant enough to be relevant to you @patrick-kidger ?
- What could be done to reduce this? (Looking for "good first issues" here :) )

ZagButNoZig:
The first small win I could find, was replacing `currently_jitting` with:

```Python
def currently_jitting():
    return isinstance(jnp.array([]), jax.core.Tracer)
```

which gives: 

```Python
%timeit fwd_eqx(model, x).block_until_ready()
286 μs ± 5.25 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

johannahaffner:
There is a small additive overhead, e.g. discussed at length here: https://github.com/patrick-kidger/equinox/issues/928

However, it looks like you are getting a little more slowdown than I would expect. Which versions of JAX and Equinox are you using?

ZagButNoZig:
I have equinox==0.11.12 and can reproduce it with both jax-nightly and jax==0.5.0. (jax 0.5.2 crashes because it can't load some xla .so file)

Is the slowdown not present on your end? @johannahaffner

johannahaffner:
[...]

PR #976, opened 2025-03-14
Avoid calling currently jitting

ZagButNoZig:
Hey,

this is a patch to reduce the overhead of filter jit, as discussed in #973.

This patch avoids calling `currently_jitting` and only computes whether we are in a jit region, when needed. 

**Note: This is very much just a proof of concept and I'm open to suggestions!**

All tests are passing (Linux CPU backend).

image

I also (micro)benchmarked this on cpu backend (which is why the numbers differ from the ones in the issue):

Old: 

```Python
%timeit fwd_eqx(model, x).block_until_ready()
208 μs ± 4.89 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

New:

```Python
%timeit fwd_eqx(model, x).block_until_ready()
121 μs ± 1.6 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

### DIFF
--- a/equinox/_errors.py
+++ b/equinox/_errors.py
@@ -327,14 +327,20 @@ def branched_error_if_impl(
             else:
                 return x
 
+    # We don't know where to cut the stack trace here, so compute all possible stack traces
+    # and let filter_jit determine which trace to use
     tb = None
+    tb_format = []
     for f, lineno in traceback.walk_stack(None):
-        ejf = f.f_locals.get("_ejf")
-        if ejf and not ejf():
-            break
+        stack_marker = f.f_locals.get("_stack_marker", None)
         if traceback_util.include_frame(f):
             tb = types.TracebackType(tb, f, f.f_lasti, lineno)
-    stack = "".join(traceback.format_tb(tb)).rstrip()
+            formatted_stack = "".join(traceback.format_tb(tb)).rstrip()
+            tb_format.append(_jit.StackCheckpoint(trace=formatted_stack, stack_marker=stack_marker))
+        else:
+            tb_format.append(_jit.StackCheckpoint(trace=None, stack_marker=stack_marker))
+            
+    stack = tb_format
     dynamic_x, static_x = partition(x, is_array)
     flat = jtu.tree_leaves(dynamic_x)
     if len(flat) == 0:
--- a/equinox/_errors.py
+++ b/equinox/_errors.py
@@ -329,7 +329,8 @@ def branched_error_if_impl(
 
     tb = None
     for f, lineno in traceback.walk_stack(None):
-        if f.f_locals.get("__equinox_filter_jit__", False):
+        ejf = f.f_locals.get("_ejf")
+        if ejf and not ejf():
             break
         if traceback_util.include_frame(f):
             tb = types.TracebackType(tb, f, f.f_lasti, lineno)
--- a/equinox/_errors.py
+++ b/equinox/_errors.py
@@ -327,8 +327,8 @@ def branched_error_if_impl(
             else:
                 return x
 
-    # We don't know where to cut the stack trace here, so compute all possible stack traces
-    # and let filter_jit determine which trace to use
+    # We don't know where to cut the stack trace here, so compute all
+    # possible stack traces and let filter_jit determine which trace to use
     tb = None
     tb_format = []
     for f, lineno in traceback.walk_stack(None):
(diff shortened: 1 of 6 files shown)


johannahaffner:
Nice speedup! You should definitely not get a 100 tests failing, though. No tests fail on MacOS for me, with Python 3.13.1 and JAX 0.5.2.

ZagButNoZig:
Oh, I was running jax-nightly and got a bunch of failures because of some deprecation on nightly. Then I downgraded to 0.5.0 and got about a million cuda version issues, so I ran the tests on cpu. All tests are passing now.

equinox/_errors.py
```diff
@@ -329,7 +329,8 @@ def branched_error_if_impl(
 
     tb = None
     for f, lineno in traceback.walk_stack(None):
-        if f.f_locals.get("__equinox_filter_jit__", False):
+        ejf = f.f_locals.get("_ejf")
+        if ejf and not ejf():
```
patrick-kidger, line 333: It's not super obvious to me that this is doing the right thing. In the current implementation the JITing-or-not is checked at the moment of wrapping.

Here it is being checked much later, at the moment we're calling `branched_error_if`.

I think that means that if we're currently in a JIT'd region then all `ejf()` checks will return True, and we will never `break` here. Correspondingly the tracebacks will be much too large, as they'll include all of the out-of-JAX calling code.

patrick-kidger: So `ejf` as written, doesn't close over anything at all, nor attempt to reset the JAX interpreter stack to the state it was in at the point the function was defined.

Note that JAX defines whether something is JIT'ing based on dynamic context (`with` statements), so it's the state of the program at the point the function is called that matters. (Not when it was defined.)

ZagButNoZig: Makes sense.

Corret me if I'm wrong, but the only time is when this would not behave as expected is when `currently_jitting` would evaluate to True and `ejf` evaluates to `False` (for the reasons you have just explained) but can that happen? Because if the outer function is called with different arguments, it would also mean the inner function is re-traced?

Could you provide me a snipped where this PR fails to break at the correct spot, so I can try to fix it? I was thinking, maybe I could also call currently_jitting [here](https://github.com/patrick-kidger/equinox/blob/8191b113df5d985720e86c0d6292bceb711cbe94/equinox/_jit.py#L253) and clean up the stack trace there.

patrick-kidger: Sure. On this snippet:
```python
import equinox as eqx
import jax.numpy as jnp

def f2(x):
    return eqx.error_if(x, x > 0, "hi")

@eqx.filter_jit
def f1(x):
    return f2(x)

def g():
    f1(jnp.array(1.))

g()
```

Then on current `main` we get:
```
  File ".../file.py", line 14, in <module>
    g()
  File ".../file.py", line 12, in g
    f1(jnp.array(1.))
EquinoxRuntimeError: Above is the stack outside of JIT. Below is the stack inside of JIT:
  File ".../file.py", line 9, in f1
    return f2(x)
           ^^^^^
  File ".../file.py", line 5, in f2
    return eqx.error_if(x, x > 0, "hi")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
equinox.EquinoxRuntimeError: hi
```

whilst with this PR we get:
```
  File ".../file.py", line 14, in <module>
    g()
  File ".../file.py", line 12, in g
    f1(jnp.array(1.))
EquinoxRuntimeError: Above is the stack outside of JIT. Below is the stack inside of JIT:
  File ".../file.py", line 14, in <module>
    g()
  File ".../file.py", line 12, in g
    f1(jnp.array(1.))
  File ".../file.py", line 9, in f1
    return f2(x)
           ^^^^^
  File ".../file.py", line 5, in f2
    return eqx.error_if(x, x > 0, "hi")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
equinox.EquinoxRuntimeError: hi
```

ZagButNoZig: Thank you!

I have updated the code to compute where to cut the stack lazily (only when an error is thrown). This should now be equivalent to main, while avoiding the currently_jitting call.

The way this works:

- Every filter jit sets a `stack_marker`
- When walking the stack trace, `branched_error_if` compute every possible stack (since it can't know where `currently_jitting` is False
- When filter_jit catches the error, it selects the stack that would have been chosen, by computing the point where `if f.locals.get(...)` would have evaluate to true

I hope this is not too chaotic of a patch 😅

ZagButNoZig:
Ok I merged upstream/main into this branch. 

I hope it's fine that this is not rebased - I messed up my commit history a bit, so a merge seemed saner.

If you want I can do a proper rebase.

ZagButNoZig:
Closed in favor of #983.

PR #980, opened 2025-03-20
Avoid calling _wrap_method.__get__ on _JitWrapper

ZagButNoZig:
This is the second (and last) patch related to #973 and should hopefully be a little simpler than #976.

The idea here is that for certain wrapped methods (mainly `_JitWrapper._call`) we recompute the BoundMethod every time. This patch caches the BoundMethod.

This should be safe, as the cached entry is used if and only if the instance is the same Module and since Modules are immutable the BoundMethod is the same.

With this patch equinox is within a few tens of milliseconds of jax (will update PR later with proper numbers).

### DIFF
--- a/equinox/_jit.py
+++ b/equinox/_jit.py
@@ -204,75 +204,10 @@ class _JitWrapper(Module):
     def __wrapped__(self):
         return hashable_combine(self._dynamic_fun, self._static_fun)
 
-    def _call(self, is_lower, args, kwargs):
-        __tracebackhide__ = True
-        # Used by our error messages when figuring out where to stop walking the stack.
-        jitting = currently_jitting()
-        if not jitting:
-            __equinox_filter_jit__ = True  # noqa: F841
-        info = (
-            self._signature,
-            self._dynamic_fun,
-            self._static_fun,
-            self.donate_first,
-            self.donate_rest,
-        )
-        dynamic_donate, dynamic_nodonate, static = _preprocess(  # pyright: ignore
-            info, args, kwargs, return_static=True
-        )
-        if is_lower:
-            return Lowered(
-                self._cached.lower(dynamic_donate, dynamic_nodonate, static),
-                info,
-                _preprocess,  # pyright: ignore
-                _postprocess,  # pyright: ignore
-            )
-        else:
-            filter = _FilterCallback()
-            callback_logger = logging.getLogger("jax._src.callback")
-            callback_logger.addFilter(filter)
-            try:
-                if self.filter_warning:
-                    with warnings.catch_warnings():
-                        warnings.filterwarnings(
-                            "ignore", message="Some donated buffers were not usable*"
-                        )
-                        marker, _, _ = out = self._cached(
-                            dynamic_donate, dynamic_nodonate, static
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Thanks for this! The basic intent of this looks reasonable to me. However note that:

(a) I believe this will introduce a memory leak: `self.cache -> _method -> instance`, so that every module instance persists forever;

(b) it is possible for `id(...)`s to be reused by the Python interpreter after the previous value using that memory location has been deallocated. In particular this is a failure mode seen in practice when repeatedly allocating and deallocating objects of the same type. For this reason it is not suitable as a dictionary key here. (Something to do with weakreferences is probably more appropriate.)

Why is the overhead here a concern for you? Assuming this is happening inside a JIT context then the overhead should be zero after compilation. (And it can always be placed inside of a JIT context, if required.)

ZagButNoZig:
So from my unterstanding and measurements, given this: 

```Python
@eqx.filter_jit
def step(m, x):
    l, g = compute_loss(m, x)
    model = eqx.apply_updates(m, g)
    return model
```

invoking step goes JitWrapper -> _call (calls `wrap_method.__get__` here) outside of the jit and thus recomputes the wrapped method. It is also visible in the trace in #973.

ZagButNoZig:
I have addressed (b) but I'm not sure if (a) is a problem, as Python should be able to detect the cyclic reference, no?

patrick-kidger:
> So from my unterstanding and measurements, given this:
> 
> ```python
> @eqx.filter_jit
> def step(m, x):
>     l, g = compute_loss(m, x)
>     model = eqx.apply_updates(m, g)
>     return model
> ```
> 
> invoking step goes JitWrapper -> _call (calls `wrap_method.__get__` here) outside of the jit and thus recomputes the wrapped method. It is also visible in the trace in #973.

Ah, good catch. Indeed that is the case.

In that case, instead of this PR, let's move `_call` to being a global function?
```python
def __call__(self, /, *args, **kwargs):
    ...
    return _call(self, False, args, kwargs)
```
If you think that sounds reasonable I'd be happy to take a PR on this. (Including a comment on why it is laid out this way!)

ZagButNoZig:
Wow, that's so much more elegant. I updated this commit and also merged this into #976. Both PRs should be ready now, but I'm not sure if you are happy with #976 yet.

Anyways, this reduces the overhead of each filter_jit call from >300μs to ~30μs on my machine.

patrick-kidger:
Okay, this LGTM! Looks like pre-commit is failing -- take a look at [CONTRIBUTING.md](https://github.com/patrick-kidger/equinox/blob/main/CONTRIBUTING.md) for how to set this up so that this is automatically checked when you commit. Happy to merge this once that is fixed :)

As for #976 I'll take a look at the new logic you've got and see what I think. (To minimise the diff, can you rebase it on top of latest `main` after this PR is merged.)

ZagButNoZig:
Awesome! 

I fixed the pre-commit issues and will rebase #976 on top of main when this is merged.

patrick-kidger:
Awesome! LGTM, so merged :D

PR #989, opened 2025-04-02
Minor optimization in hashable_partition

ZagButNoZig:
This is just a very small change to `hashable_partition` that avoids calling `filter_fn` twice for every leaf.

On my machine this saves 200μs on the MLP example (~260 leaves).

All tests are passing and there should be no behavior change.

Old: 

```Python
%timeit -r7 -n500 fwd_eqx(dyn_flat, x).block_until_ready()
2.36 ms ± 58.1 μs per loop (mean ± std. dev. of 7 runs, 500 loops each)
```

New:

```Python
%timeit -r7 -n500 fwd_eqx(dyn_flat, x).block_until_ready()
2.16 ms ± 47 μs per loop (mean ± std. dev. of 7 runs, 500 loops each)
```

### DIFF
--- a/equinox/_compile_utils.py
+++ b/equinox/_compile_utils.py
@@ -20,9 +20,17 @@ def hashable_filter(pytree: PyTree, filter_fn: Callable):
 
 def hashable_partition(pytree: PyTree, filter_fn: Callable):
     leaves, treedef = jtu.tree_flatten(pytree)
-    dynamic_leaves = tuple(x if filter_fn(x) else None for x in leaves)
-    static_leaves = tuple(None if filter_fn(x) else x for x in leaves)
-    return dynamic_leaves, (static_leaves, treedef)
+    length = len(leaves)
+    a = [None] * length
+    b = [None] * length
+
+    for i, x in enumerate(leaves):
+        if filter_fn(x):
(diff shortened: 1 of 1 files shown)


equinox/_compile_utils.py
```diff
@@ -20,9 +20,17 @@ def hashable_filter(pytree: PyTree, filter_fn: Callable):
...
+    a = [None] * length
+    b = [None] * length
+
+    for i, x in enumerate(leaves):
+        if filter_fn(x):
+            a[i] = x
+        else:
+            b[i] = x
+
+    return tuple(a), (tuple(b), treedef)
```
patrick-kidger, line 33: I think we might be able to skip the `tuple(...)` and return lists instead. I think the output here is only ever consumed by `hashable_combine`. Does that improve things a little further?

ZagButNoZig: On my machine this throws some strange errors (`ValueError: Non-hashable static arguments are not supported. An error occurred while trying to hash an object of type <class 'tuple'>`) and I measured some tuple to list conversions (for 1000 elements) and it never takes more than 1μs, so not sure if it's even worth it.

Issue #961, opened 2025-02-26
Slow Initialization


nstarman:
Hi @patrick-kidger. I've recently been speed testing some code and found that Equinox is around 2x slower than a custom pytree.

```python
import jax
import equinox as eqx

def func(x, y):
    return (x - y) / (x + y)

x = jnp.linspace(0.0, 1, 10)
y = jnp.linspace(1.0, 2, 10)

print(jax.make_jaxpr(func)(x, y))
# { lambda ; a:f32[10] b:f32[10]. let
#     c:f32[10] = sub a b
#     d:f32[10] = add a b
#     e:f32[10] = div c d
#   in (e,) }

f = jax.jit(func)
f(x, y)
# %timeit f(x, y)
# 2.87 µs ± 95 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Now with a custom Module

```python
class MyEqxArray(eqx.Module):
    array: Array

    def __add__(self, other):
        return jax.tree.map(jnp.add, self, other)

    def __sub__(self, other):
        return jax.tree.map(jnp.subtract, self, other)

    def __truediv__(self, other):
        return jax.tree.map(jnp.divide, self, other)


mx = MyEqxArray(x)
my = MyEqxArray(y)

func(mx, my)

print(jax.make_jaxpr(func)(mx, my))  # same jaxpr 👍 
# { lambda ; a:f32[10] b:f32[10]. let
#     c:f32[10] = sub a b
#     d:f32[10] = add a b
#     e:f32[10] = div c d
#   in (e,) }

f = jax.jit(func)
f(mx, my)
%timeit f(mx, my)
# 8.4 µs ± 969 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Now with a custom PyTree

```python
from dataclasses import dataclass

@jax.tree_util.register_pytree_node_class
@dataclass
class CustomArray:
    array: Array

    def __add__(self, other):
        return jax.tree.map(jnp.add, self, other)

    def __sub__(self, other):
        return jax.tree.map(jnp.subtract, self, other)

    def __truediv__(self, other):
        return jax.tree.map(jnp.divide, self, other)

    def tree_flatten(self) -> tuple[tuple[Any], Any]:
        return (self.array,), None

    @classmethod
    def tree_unflatten(cls, aux_data: Any, children: tuple[Any]) -> "CustomArray":
        return cls(*children)

f = jax.jit(func)
f(mx, my)
[...]

PR #994, opened 2025-04-10
perf: speed up module (un)flattening

nstarman:
Partly addresses #961.

I don't see a benchmarking suite, e.g. using https://pypi.org/project/pytest-codspeed/, so in the attached notebook (attached as a ZIP because GH doesn't support directly attaching ipynb) I'm finding ~25% speedup towards matching the raw jax (around a 10% speedup overall).
[overhead.ipynb.zip](https://github.com/user-attachments/files/19694256/overhead.ipynb.zip)

This PR:

1. removes `_flatten_module` in favor of one defined in `_ActualModuleMeta.__new__` allowing for many aspects of the flattening to be pre-computed.
2. add slots to `_FlattenedData`, which should provide a minuscule speedup and reduction in memory use. Didn't hurt so 🤷.
3. Combined the `wrapper_field_names` and `wrapper_field_values` into a single field of `_FlattenedData` since it's only  ever used as a zipped combo. Also changed to pre-determined size for `wrapper_field_names_and_values` by storing the sentinel value. The sentinel is then used for skipping assignment when unflattening. In testing this did provide a speedup. 
4. Same as 3. for the static field names and values.

For future PRs...
- figure out how to not filter on `__dict__` so that `dynamic_fs = data_fs` and tuple sizes are pre-determined.
- try a namedtuple `_FlattenedData` to see if that speeds things up further.
- JAX dataclass-related speedups.

### DIFF
--- a/equinox/_module.py
+++ b/equinox/_module.py
@@ -8,18 +8,19 @@ import abc
 import dataclasses
 import functools as ft
 import inspect
+import itertools
 import types
 import warnings
 import weakref
 from collections.abc import Callable
-from typing import Any, cast, Protocol, TYPE_CHECKING, TypeVar
+from typing import Any, cast, Final, Protocol, TYPE_CHECKING, TypeVar
 from typing_extensions import dataclass_transform, ParamSpec
 
 import jax
 import jax._src.traceback_util as traceback_util
 import jax.tree_util as jtu
 import numpy as np
-from jaxtyping import Array, Bool, PyTreeDef
+from jaxtyping import Array, Bool, PyTree, PyTreeDef
 
 from ._better_abstract import ABCMeta, dataclass
 from ._caches import cache_clears
@@ -132,6 +133,8 @@ def field(
 # Part 2: Modules!
 # This is the core of Equinox.
 #
+
+_flatten_sentinel: Final = doc_repr(object(), "-")
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


equinox/_module.py
```diff
@@ -478,11 +480,44 @@ def __new__(
...
+        )
+        static_fs = tuple(
+            f.name for f in dataclasses.fields(cls) if f.metadata.get("static", False)
+        )
+
+        def flatten_module(
+            obj: "Module", /, with_keys: bool
+        ) -> tuple[tuple[PyTree, ...], _FlattenedData]:
+            # Static metadata, placed in aux.
+            static_vs = [getattr(obj, name) for name in static_fs]
```
patrick-kidger, line 498: I think static fields could also fail to be initialised at times, e.g. when partway through `__init__`.

Also, I think this needs to be `obj.__dict__.get(name)` to avoid looking up through the class and the class hierarchy, in case of shadowing. We only want to grab instance attributes.

equinox/_module.py
```diff
@@ -478,11 +480,44 @@ def __new__(
...
+            # Subnodes in the PyTree (filtering on `__dict__` to avoid picking
+            # up `property` and uninitialized values).
+            dynamic_fs = [name for name in data_fs if name in obj.__dict__]
+            if with_keys:
+                dynamic_vs = [(jtu.GetAttrKey(k), getattr(obj, k)) for k in dynamic_fs]
+            else:
+                dynamic_vs = [getattr(obj, k) for k in dynamic_fs]
```
patrick-kidger, line 505: Since speed is a concern, note that this is doing two lookups into `obj.__dict__`, once in `name in obj.__dict__` and another in `getattr`. I think the former could be skipped by replacing the latter with some kind of `obj.__dict__.get(k, _some_sentinel_value)`.

(I also speculate that a small speedup could be arranged by assigning to a local variable:
```python
get = obj.__dict__.get
dynamic_vs = [get(k, _some_kind_of_sentinel) for k in dynamic_fs]
```
)

patrick-kidger (commented):
This looks good to me! I've left some minor comments but I like the look of this.

equinox/_module.py
nstarman: Done!

nstarman:
I refactored the module flattening for further optimizations.
I think you need to enable the tests to run.

equinox/_module.py
nstarman: https://github.com/patrick-kidger/equinox/pull/994/commits/53e829297c55c675cc49b3472a091219d141fc52 only uses a single loop for getting from the dict.

equinox/_module.py
```diff
@@ -478,11 +480,77 @@ def __new__(
...
+
+        def flatten_module(
+            obj: "Module", /
+        ) -> tuple[tuple[PyTree, ...], _FlattenedData]:
+            """Flattens a module into a flattened representation."""
+            d = obj.__dict__
+            sntl = _flatten_sentinel
+
+            # Static fields, filtered on property / init state
+            static_vs = [v for k in static_fs if (v := d.get(k, sntl)) is not sntl]
```
patrick-kidger, line 502: I think this will end up being of variable length if we get a sentinel value. That will then produce the wrong behaviour at unflattening time.

We could either similarly filter `static_fs` here, or skip pass sentinels in the unflattening. Wild guess that the latter is more efficient as in general sentinels will be uncommon.

patrick-kidger (commented):
Okay, this looks good to me! I think I spot one bug but otherwise I'm happy to merge this.

equinox/_module.py
nstarman: Done! I don't see any speedup, but %timeit is an imprecise tool.
This also afforded some further simplifications in `_FlattenedData` and then in `_unflatten_module`

equinox/_module.py
```diff
@@ -478,11 +481,69 @@ def __new__(
...
+                v = d.get(k, sntl)
+                if v is sntl:
+                    continue
+                dynamic_fs.append(k)
+                dynamic_vs.append(v)
+
+            aux = _FlattenedData(
+                tuple(dynamic_fs),
+                tuple([(k, d.get(k, sntl)) for k in static_fs]),
+                tuple([(k, d.get(k, sntl)) for k in _wrapper_field_names]),
```
patrick-kidger, line 516: Nit: this is materialising a list and then converting it to a tuple. I think a direct `tuple((k, d.get(k, sntl) for k in static_fs)` should presumably save a small amount.

nstarman: One would think. But I've timed it out for different stuff and haven't found that to be the case. Here's a quick test:

CleanShot 2025-04-13 at 17 44 27@2x

patrick-kidger: Mysterious! Well fair enough then :)

patrick-kidger:
Alright, awesome -- merged! 😁

PR #996, opened 2025-04-11
feat: converter = None as the sentinel

nstarman:
1. Easier for users to access instead of a private sentinel.
2. simplifies later performance-related logic changes.
3. Broadens support to allow `dataclasses.field(metadata=dict(converter=None))`, not just `eqx.field`.

Follow up to #994.

### DIFF
--- a/equinox/_module.py
+++ b/equinox/_module.py
@@ -300,27 +300,28 @@ class _ActualModuleMeta(ABCMeta):
         # checkers.
         # Note that mutating the `__init__.__annotations__` is okay, as it was created
         # by the dataclass decorator on the previous line, so nothing else owns it.
-        for f in dataclasses.fields(cls):
-            if f.name not in cls.__init__.__annotations__:
-                continue  # Odd behaviour, so skip.
+        if has_dataclass_init:
+            for f in dataclasses.fields(cls):
+                if not f.init:
+                    continue
 
-            if (converter := f.metadata.get("converter")) is None:
-                continue  # No converter, so skip.
+                if (converter := f.metadata.get("converter")) is None:
+                    continue  # No converter, so skip.
 
-            try:
-                signature = inspect.signature(converter)
-            except ValueError:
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Looks reasonable to me! Let's get #994 in and then this one.

nstarman:
@patrick-kidger. Ready!

equinox/_module.py
```diff
@@ -305,25 +303,24 @@ def __new__(
         for f in dataclasses.fields(cls):
             if f.name not in cls.__init__.__annotations__:
                 continue  # Odd behaviour, so skip.
+
```
patrick-kidger, line 306: Actually, I think I'm noticing a bug in the existing code here -- I think we should only be running all of this if `has_dataclass_init`? Worth fixing up here, if you agree?

patrick-kidger (commented):
One comment only, but otherwise LGTM!

equinox/_module.py
nstarman: Done.

equinox/_module.py
```diff
@@ -324,7 +322,7 @@ def __new__(
                         converter_annotation = parameters[0].annotation
                         if converter_annotation is inspect.Signature.empty:
                             converter_annotation = Any
-                cls.__init__.__annotations__[f.name] = converter_annotation
+                    cls.__init__.__annotations__[f.name] = converter_annotation
```
patrick-kidger, line 325: I think this is an unintentional indent?

nstarman: unintentional!

nstarman: Fixed

patrick-kidger:
Okay, great! All looks good me, so merged. :)

Heads-up for other PRs in the near future,I've just spun up a `dev` branch to merge this into (as changes to `main` immediately appear in our documentation, and here the `eqx.field` signature has changed).

PR #998, opened 2025-04-11
refactor: collect Module (un)flattening

nstarman:
A followup to #994.
AFAICT there's no performance impact, just a cleaner definition.
Still in draft mode because I'm trying to see if there's a way to add speed ups, e.g. define the un/flattening in C++ or whatnot. Seems a separate-enough discussion from #994.

### DIFF
--- a/equinox/_module.py
+++ b/equinox/_module.py
@@ -300,8 +300,9 @@ class _ActualModuleMeta(ABCMeta):
         # checkers.
         # Note that mutating the `__init__.__annotations__` is okay, as it was created
         # by the dataclass decorator on the previous line, so nothing else owns it.
+        fields = dataclasses.fields(cls)
         if has_dataclass_init:
-            for f in dataclasses.fields(cls):
+            for f in fields:
                 if not f.init:
                     continue
 
@@ -394,7 +395,7 @@ class _ActualModuleMeta(ABCMeta):
                 if (base_num_fields > 0) or (not _has_dataclass_init[base]):
                     # If we've added more fields, or added a custom init method, then
                     # error.
-                    if len(dataclasses.fields(cls)) != base_num_fields:
+                    if len(fields) != base_num_fields:
                         raise TypeError(
                             "For readability, any custom `__init__` method, and all "
                             "fields, must all be defined on the same strict Module. "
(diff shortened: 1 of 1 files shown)


nstarman:
Thoughts on this PR:
- An alternative to this PR is to define `unflatten` on `ActualModuleMeta`. Then it doesn't need the `ft.partial`.
- Or `ModuleUnflattener` could be written in C++ with pybind11, like JAX does. That eliminates many Python overheads.
- Or compilation with `mypyc` (https://mypyc.readthedocs.io/en/latest/native_classes.html#inheritance) to likewise eliminate overheads.

But I'm hesitant to sink a lot of time into this if it isn't going to move the needle performance-wise. If you have tips for figuring out where Module is slowest, that would be great. I'm having a tough time doing a timing breakdown. I've messed with `__hash__` a bit but haven't seen performance gains.

[overhead.ipynb.zip](https://github.com/user-attachments/files/19720272/overhead.ipynb.zip)

patrick-kidger:
> Thoughts on this PR:
> 
> * An alternative to this PR is to define `unflatten` on `ActualModuleMeta`. Then it doesn't need the `ft.partial`.

I'd like to avoid that -- we deliberately don't define any of methods on `eqx.Module` (or its metaclass). This keeps it as a base class that is completely agnostic to what the user does, e.g. if they were to define a method called `unflatten` for some other purpose.

> * Or `ModuleUnflattener` could be written in C++ with pybind11, like JAX does. That eliminates many Python overheads.
> * Or compilation with `mypyc` (https://mypyc.readthedocs.io/en/latest/native_classes.html#inheritance) to likewise eliminate overheads.

I'd like to avoid compilation/C++ approaches as this becomes substantially more difficult to maintain. Pure Python works with anything. The moment these enter the story then it becomes a matter of recompiling for the various combinations, worrying about manylinux specs, ABI stability, ... I've been down this road before (with libtorch) and it's not something I want to take on again.

> But I'm hesitant to sink a lot of time into this if it isn't going to move the needle performance-wise. If you have tips for figuring out where Module is slowest, that would be great. I'm having a tough time doing a timing breakdown. I've messed with `__hash__` a bit but haven't seen performance gains.

My usual favourite profiler is [py-spy](https://github.com/benfred/py-spy/), although I've not checked if it'd be useful on this problem. Given the fairly small time intervals involved here you could maybe increase the density of sampling hits by running it in Monte-Carlo fashion, i.e.
```python
for _ in range(big number):
    flatten()
    unflatten()
```

nstarman:
> I'd like to avoid that ...  If they were to define a method called `unflatten` for some other purpose.

The actual method names would be `tree_flatten`, `tree_flatten_with_keys`, and `tree_unflatten`. These are presumably pretty rare names to be defined on a class except for JAX purposes!
But if even these methods are to be avoided then this PR is worth doing, for concision if not speed. I'll have to rebase after #994 is in.

nstarman:
> My usual favourite profiler is [py-spy](https://github.com/benfred/py-spy/), although I've not checked if it'd be useful on this problem. Given the fairly small time intervals involved here you could maybe increase the density of sampling hits by running it in Monte-Carlo fashion, i.e.

I'll take a look into this soon, because even with all the optimizations in the past couple PRs, Equinox Modules are ~25% points slower than an equivalent JAX PyTree and ~60% slower than a JAX array. If we can close the 25% gap then any further optimizations would have to involve C code for the (un)flattening / improvements to JAX itself.

equinox/_module.py
```diff
@@ -904,19 +850,86 @@ def __repr__(self):
...
+
+    def flatten(self, obj: "Module", /) -> tuple[tuple[PyTree, ...], _FlattenedData]:
+        """Flattens a module into a flattened representation."""
+        d = obj.__dict__
+
+        # Build dynamic keys, values in a single pass
+        dynamic_fs = []
+        dynamic_vs = []
+        for k in self.dynamic_fs:
+            v = d.get(k, _flatten_sentinel)
```
patrick-kidger, line 876: Btw, does replacing this with
```
get = obj.__dict__.get
...
v = get(k, _flatten_sentinel)
```
help performance here at all? (If only by some very minor amount.)

patrick-kidger (commented):
Changes here seem pretty minor, so LGTM!

equinox/_module.py
nstarman: It should...

nstarman: Done.

nstarman: 🤷 hard to tell what diff fit made. Well within the run-to-run differences for `%timeit`. Hopefully #1001 will help and I can go back and A/B test this more rigorously.

patrick-kidger:
Great, this LGTM! I've just updated the target branch to our new dev and it looks like there's a merge conflict. Modulo that happy to merge.

patrick-kidger:
Ty! Merged :)

PR #999, opened 2025-04-14
refactor: simplifications in ModuleMeta __call__

nstarman:


nstarman:
@patrick-kidger I think this is ready for review.

nstarman:
Some possible ideas, inspired from [`dataclassish`](https://github.com/GalacticDynamics/dataclassish/blob/main/src/dataclassish/_src/converters.py)

1. Always define a `__dataclass_init__` method that is constructed by `Module` and which is the init method created by dataclass if `init=True`. This is very useful when defining a custom init method to be able to do

```python
class MyClass
    def __init__(self, ...):
        ... # Whatever manual processing
        self.__dataclass_init__(...)  # don't have to worry about setting in frozen vs not, descriptors vs not, etc.
```

Of course it's totally optional to use, it's just very convenient!

2. Always define a `__converter_init__` method that is constructed by `Module` and which does 

```python
def __converter_init__(self, ...):
    ba = run_converters(...)
    self.__dataclass_init__(**ba)
```

When `dataclass_init=True` this is swapped in for `__init__` so that input conversion happens early, which should be more efficient (e.g. saving 2x assignment).
When `dataclass_init=False` then users can simplify their `__init__` methods by doing

```python
class MyClass
    def __init__(self, ...):
        ... # Whatever manual processing
        self. __converter_init__(...)  # handles conversion, and all complexities of setting.
```

Of course it's totally optional to use, it's just very convenient!

3. If `dataclass_init` is True and the `run_converters` skip args that are Tracers then the `__init__` method satisfies JAX's assumptions about dataclasses and we can use the faster pytree registration.

nstarman:
SGTM. I think we can also do the converter loop as part of the `__init__` trick when `dataclass_init=True` as an internal optimization. That should reduce setting from 2x to 1x.

PR #1009, opened 2025-04-21
Clarify Lambda.fn behavior in PyTrees

adonath:
This PR just adds a short sentence, clarifying that `Lambda.fn` is treated as a node of the PyTree. The behavior seemed surprising to me at first.


patrick-kidger:
LGTM! Thank you for the doc improvement :)

PR #1020, opened 2025-05-09
add esmj (and joltz and boltz-binder-design :) ) to awesome-list.md

nboyd:
Hi Patrick,

Adding ESM-C as requested! I also shamelessly slipped in joltz and boltz-binder-design -- can remove these if you like 

Nick


patrick-kidger:
Awesome! LGTM. Sneaking in extra fun biology sounds good to me :D

PR #1026, opened 2025-05-16
feat: TreeWLCustom for pretty printing.

nstarman:
Get the WL doc used by `tree_pformat`.
This is useful if you want to make a WL similar to how Equinox/Module does it, but you need to modify the WL before stringifying it.
I also added the `respect_pdoc` kwarg so that `tree_pdoc` can be applied to ~/ used within~ objects with a `__pdoc__`.

Alternatively, if `custom` were made public then `tree_pdoc` wouldn't need to exist. For example,

```python
@dataclasses.dataclass(slots=True)
class TreeWLCustom:
    truncate_leaf: Callable[[Any], bool]
    short_arrays: bool
    struct_as_array: bool

    def __call__(self, obj: Any) -> wl.AbstractDoc:
        if self.truncate_leaf(obj):
            return wl.TextDoc(f"{type(obj).__name__}(...)")

        ...

def tree_pformat(...):
    return wl.pformat(..., custom= TreeWLCustom(...), ...)
```

### DIFF
--- a/equinox/_pretty_print.py
+++ b/equinox/_pretty_print.py
@@ -25,17 +25,23 @@ def _false(_):
 
 @dataclasses.dataclass(slots=True)
 class TreeWLCustom:
-    """Wadler-Lindig customisation for PyTree pretty-printing."""
+    """Wadler-Lindig customisation for PyTree pretty-printing.
+
+    **Arguments:**
+    - `truncate_leaf`: A function `Any -> bool`. Applied to all nodes in the
+        PyTree; all truthy nodes will be truncated to just "<class_name>(...)".
+    - `short_arrays`: If `True`, arrays are abbreviated to their shape and
+      dtype (with weak dtype info prepended to the dtype name).
+    - `struct_as_array`: If `True`, JAX ShapeDtypeStructs are abbreviated to
+      their shape and dtype.
+
+    """
 
     truncate_leaf: Callable[[Any], bool]
-    """Function to determine if a leaf should be truncated."""
 
     short_arrays: bool
-    """If `True`, JAX arrays are abbreviated to their shape and dtype."""
 
     struct_as_array: bool
-    """If `True`, JAX ShapeDtypeStructs are abbreviated to their shape and
-    dtype."""
 
     def __call__(self, obj: Any) -> wl.AbstractDoc | None:
         if self.truncate_leaf(obj):
--- a/equinox/_pretty_print.py
+++ b/equinox/_pretty_print.py
@@ -23,6 +23,42 @@ def _false(_):
     return False
 
 
+@dataclasses.dataclass(slots=True)
+class TreeWLCustom:
+    """Wadler-Lindig customisation for PyTree pretty-printing."""
+
+    truncate_leaf: Callable[[Any], bool]
+    """Function to determine if a leaf should be truncated."""
+
+    short_arrays: bool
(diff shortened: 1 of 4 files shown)


equinox/_pretty_print.py
```diff
@@ -23,20 +23,22 @@ def _false(_):
...
-    width: int = 80,
     indent: int = 2,
     short_arrays: bool = True,
     struct_as_array: bool = False,
     truncate_leaf: Callable[[Any], bool] = _false,
-) -> str:
-    """Pretty-formats a PyTree as a string, whilst abbreviating JAX arrays.
+    respect_pdoc: bool = True,
+) -> wl.AbstractDoc:
+    """Pretty-formats a PyTree as a Wadler-Lindig document.
```
patrick-kidger, line 36: Hmm, I'm not 100% sold on making WL part of the Equinox public API. For example I could see we might want the flexibility to make a breaking change in WL somewhere, and that could have impact on Equinox users who might not otherwise expect it.

Our `custom` method is not that big, you could probably simply copy-paste it into your application? Alternatively if you're particularly concerned that the two implementations might diverge over time then we could move it to the top-level of the module so you could import (though still a private import as again I don't think this is something I'd like Equinox to commit to in its public API)?

nstarman: I'm open to any option. I just find myself copying it in a few different places and want to just keep the code in sync.
If you're more amenable to exposing the `custom` method, than maybe as a class in `eqxi` ?

nstarman: Ah. circular imports means `_pretty_print` cannot use anything in `internal`.

nstarman: Ok. Done. `TreeWLCustom`. Not in the docs.

nstarman: Though maybe it would still be preferable to have the `TreeWLCustom` exposed via `equinox.internal`?

```
# _pretty_print.py
class TreeWLCustom: ...

# internal/__init__.py
from .._pretty_print import TreeWLCustom
```

patrick-kidger: Yup, I like the idea of putting it in eqxi. Let's do that.

On circularity: `eqxi` is allowed to depend on `eqx`, but not the other way around. It's meant as a place to factor out a lot of private helpers that would otherwise be copy-pasted around the Diffrax/Optimistix/etc libraries.

nstarman: Done!

mkdocs.yml
```diff
@@ -84,6 +84,7 @@ plugins:
             - jax.extend.core.ClosedJaxpr
             - jax.Device
             - jax.sharding.Sharding
+            - wadler_lindig.AbstractDoc
```
patrick-kidger, line 87: This line probably isn't necessary any more?

nstarman: True!

mkdocs.yml
```diff
@@ -83,7 +83,7 @@ plugins:
             - jax.ShapeDtypeStruct
             - jax.extend.core.ClosedJaxpr
             - jax.Device
-            - jax.sharding.Sharding
+            - jax.sharding.Shardingwadler_lindig.AbstractDoc
```
patrick-kidger, line 86: I don't think so 😁

nstarman: 😆

PR #1028, opened 2025-05-18
New cleaner Module implementation.

patrick-kidger:
In this PR I've refactored the definition of `eqx.Module`, with the goal of maintaining the same behaviour whilst meaningfully simplifying the implementation. (Which has grown rather a lot of cruft over the years.) Hopefully the result is back to being fairly readable + understandable on what it's actually doing.

The main changes are:

- No longer swapping out `__class__` during initialisation with an unfrozen one. A lot of our previous code was around handling the weird edge cases that arise out of that. Now we're just adding a marker in the dynamic context instead.
- No longer using a descriptor to ensure that bound methods are pytrees. Instead we check this during `Module.__getattribute__`.
- Dropped support for `class Foo(eqx.Module, strict=True)`. In the end 'strict modules' were never documented nor graduated into the public API, as I wasn't comfortable with the idea of a two-tier 'module' vs 'strict module' system. If I was to write Equinox again from scratch then they'd be the default, but as it is this feature isn't useful enough to justify a breaking change, nor to keep around on an optional basis for all five of us who really want this.

I've already ran the jaxtyping+Lineax+Optimistix+Diffrax test suites against these changes and they all pass.

I'm tagging @nstarman @NeilGirdhar @ZagButNoZig @jeertmans @lockwo as power users who have previously contributed to the `eqx.Module` implementation -- in case you'd like to try out these changes on your own use-cases, and make sure that I haven't accidentally broken you!


NeilGirdhar:
Thanks for the ping.  This is a nice improvement to maintainability of `equinox.Module`.  I love how you broke up the monolithic file.

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+    dynamic_fs: tuple[str, ...]
+    static_fs: tuple[str, ...]
+
+    def __init__(self, fields: tuple[dataclasses.Field[Any], ...]):
+        self.dynamic_fs = tuple(
+            [f.name for f in fields if not f.metadata.get("static", False)]
+        )
+        self.static_fs = tuple(
+            [f.name for f in fields if f.metadata.get("static", False)]
+        )
```
nstarman, line 69: Just wondering if this is faster:

```suggestion
        dynamic_fs = []
        static_fs = []
        for f in fields:
            if f.metadata.get("static", False):
                dynamic_fs.append(f.name)
            else:
                static_fs.append(f.name)
        self.dynamic_fs = tuple(dynamic_fs)
        self.static_fs = tuple(static_fs)
```

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+
+        return cls
+
+    def __call__(cls, *args, **kwargs):  # noqa: N805
+        __tracebackhide__ = True
+        if cls in _abstract_module_registry:
+            # Any other is-abstract checks will be handled in super().__call__.
+            raise TypeError("Cannot instantiate abstract `equinox.Module`.")
+        if _has_dataclass_init[cls]:
```
nstarman, line 405:
```suggestion
        if cls.__dataclass_params__.init:
```

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+                _currently_initialising.remove(self)
+        assert not is_abstract_module(cls)  # pyright: ignore[reportArgumentType]
+
+        fields = dataclasses.fields(cls)  # pyright: ignore[reportArgumentType]
+        missing_names = {
+            f.name
+            for f in fields
+            # Not `vars` or `__dict__`, to allow for `property`s overwriting a field.
+            # Not recommended, but allowable for backward compatibility.
+            if f.name not in dir(self)
```
nstarman, line 423: I think this makes `dir` every time?

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+                        stacklevel=2,
+                    )
+
+        for parent_cls in cls.__mro__:
+            try:
+                check_init = parent_cls.__dict__["__check_init__"]
+            except KeyError:  # noqa: PERF203
+                pass
+            else:
+                check_init(self)
```
nstarman, line 448:
```suggestion
        for parent_cls in cls.__mro__:
            if (check_init := parent_cls.__dict__.get("__check_init__", False)):
                check_init(self)
```

nstarman (commented):
This is looking great!
Some preliminary comments. If you want to do typing stuff in a different PR then ignore about half of these comments 🤣

equinox/_module/_module.py
nstarman: Can be hard to tell, esp with Jax, without running some tests. And I'm in an airport about to board 😅.

equinox/_module/_prebuilt.py
```diff
@@ -0,0 +1,100 @@
...
+    def __post_init__(self):
+        for field_name in wrapper_field_names:
+            try:
+                value = getattr(self.__func__, field_name)
+            except AttributeError:
+                pass
+            else:
+                setattr(self, field_name, value)
+
+    def __call__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
NeilGirdhar: Why not dump the `__call__` completely?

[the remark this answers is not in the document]
NeilGirdhar: Why not use type parameters and `ParamSpec`?

equinox/_module/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+- Equinox has a lot of nice error messages for every kind of edge case we can think of:
+  - Accidentally marking arrays as static fields;
+  - Accidentally creating cycles via method assignment;
+  - Accidentally closing over arrays;
+  - Accidentally ignoring `__post_init__` because an `__init__` method already exists.
+- Good behaviour around edge cases:
+  - Bound methods like `foo = SomeModule().some_method` are also pytrees, so that things like `jax.jit(...)(foo)` work correctly.
+  - Subclassing propagates `__init__` methods, `__doc__`strings.
+  - Mutating attributes during `__init__` is totally fine because we don't do any `__setattr__` magic.
```
jeertmans, line 17: Wouldn't be better to include, for each line (at least non-trivial ones), a small code example?

I think that not all users may see how we could create cycles via method assignments, or *closing over arrays* (especially for non-native English speakers, like me, I had to check on SO to know what it meant ^^').

equinox/_module/README.md
jeertmans, line 1: Is this document meant to be included in the online documentation too? (I am not a mkdocs user, so I don't know what's getting included by default.)

equinox/_module/_field.py
```diff
@@ -0,0 +1,90 @@
...
+        metadata["converter"] = converter
+    if static:
+        metadata["static"] = True
+    return dataclasses.field(metadata=metadata, **kwargs)
+
+
+def static_field(**kwargs):
+    warnings.warn(
+        "`equinox.static_field` is deprecated in favour of `equinox.field(static=True)`"
+    )
```
jeertmans, line 89: Shouldn't it raise a `DeprecationWarning` specifically? Also, I think it is [recommended to specify the stacklevel (and set it to 2)](https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/).

equinox/_module/_prebuilt.py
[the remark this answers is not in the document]
jeertmans: > Why not use type parameters and `ParamSpec`?

Then, I think, the class should be generic over some `ParamSpec` P, like so:

```python
class BoundMethod(Module):
    """Just like a normal Python bound method... except that this one is a PyTree!
    This stores `__self__` as a subnode.
    """

    __func__: Callable[P, R] = field(static=True)
    __self__: Module
```

but then we use `Callable` instead of `FunctionType`, and I'm not sure what are the differences between the two.

equinox/_module/_prebuilt.py
```diff
@@ -0,0 +1,100 @@
...
+class BoundMethod(Module):
+    """Just like a normal Python bound method... except that this one is a PyTree!
+
+    This stores `__self__` as a subnode.
+    """
+
+    __func__: types.FunctionType = field(static=True)
+    __self__: Module
+
+    def __post_init__(self):
```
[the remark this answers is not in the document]
jeertmans: `-> None` and `: Any` aren't used anywhere in the codebase, so I am not sure if it's real an added-value to add them just here.

tests/test_module.py
jeertmans, line 1: Suggestion: add a new test to validate that `static_field` raises a `DeprecationWarning`?

jeertmans (commented):
Hi @patrick-kidger, thanks for tagging me on this!

I reviewed your PR and left a few comments. I also ran the tests (more than 500) on my main tool that heavily depends on Equinox. A few of them are now breaking, but I think it is because of type-checking is working *better* (probably due to a combined upgrade of Beartype, jaxtyping and Equinox) and

```python
class Test(eqx.Module)
  x: Array = eqx.field(converter=jnp.asarray)

Test(x=np.array([1,2,3]))
```
will now raise an error if `x` is not exactly a JAX array.

It makes sense from the type-checking point-of-view, but it would be great if the `__ini__` method generated by `eqx.Module` could use a different type-hint than the class attributes themselves (mostly because we use converters). (But I guess it is probably a limitation inherited from [`dataclass` itself](https://stackoverflow.com/questions/62753881/how-to-annotate-the-type-of-field-in-dataclass-to-be-different-from-the-type-of).)

Otherwise, I didn't see any other error.

equinox/_module/_prebuilt.py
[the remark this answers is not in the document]
patrick-kidger: Yup, by convention I leave off `-> None` on `__init__` and `__post_init__` as they can never return anything else.

equinox/_module/_prebuilt.py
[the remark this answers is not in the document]
patrick-kidger: FWIW this is an internal-only class that's only made available via `eqx.Module.__getattribute__`. I have no strong feelings on adding type information here or not, but I do note that it won't be checked by any type checker!

equinox/_module/_better_abstract.py
```diff
@@ -236,7 +240,7 @@ def __call__(cls, *args, **kwargs):
 
 
 @dataclass_transform()
-def dataclass(**kwargs):
+def better_dataclass(**kwargs):
```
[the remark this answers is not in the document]
patrick-kidger: It already is on the line above I think?

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+                    # Non-Module superclasses.
+                    if kls.__init__ is not object.__init__:
+                        has_dataclass_init = False
+                        break
+                else:
+                    break
+            else:
+                assert name == "Module"
+                has_dataclass_init = True  # eqx.Module itself
+        _has_dataclass_init[cls] = has_dataclass_init
```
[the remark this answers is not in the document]
patrick-kidger: Good point. I'll see if I can remove this.

equinox/_module/_module.py
patrick-kidger: Haha! From past experience (but without having benchmarked it here) I suspect your change will actually be slightly faster.

equinox/_module/_module.py
patrick-kidger: Makes sense!

equinox/_module/_module.py
patrick-kidger: Hmm, this is relying on an implicit-cast-to-bool (in this case that functions are truthy). I try to avoid ever using this, I regard it as a misfeature / an easy source of bugs.

equinox/_module/README.md
patrick-kidger: I considered this and left it off, in the interests of a smaller document.

This document isn't meant to be public anywhere, it's just meant as an easter egg summary for anyone curious enough to go poking through the codebase!

equinox/_module/_field.py
patrick-kidger: IIUC deprecation warnings do not surface by default except in tests. For that reason they often get missed.

Good point about the stacklevel, I'll use that!

patrick-kidger:
Thank you for the quick feedback everyone!

> It makes sense from the type-checking point-of-view, but it would be great if the `__ini__` method generated by `eqx.Module` could use a different type-hint than the class attributes themselves

I think it should already be pulling in the type annotations from the converters:

https://github.com/patrick-kidger/equinox/blob/adaaf161d9bff64f16dc924f61bc6884a884ca7a/equinox/_module/_module.py#L364-L386

Is this not working for your use-case?

equinox/_module/_field.py
jeertmans: Indeed, from [`warnings`](https://docs.python.org/3/library/warnings.html)
> category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in `__main__`).

But, in the meantime, this is what Python recommends doing for such types of warnings. JAX itself does that.

Missing such warning is not a big deal from the user point of view (as you should use strict requirements). Developer should run tests with `-Werror` :-)

equinox/_module/_better_abstract.py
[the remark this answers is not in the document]
nstarman: lol. Completely missed that.

equinox/_module/_module.py
nstarman: What about an is not False check than?

jeertmans:
> Thank you for the quick feedback everyone!
> 
> > It makes sense from the type-checking point-of-view, but it would be great if the `__ini__` method generated by `eqx.Module` could use a different type-hint than the class attributes themselves
> 
> I think it should already be pulling in the type annotations from the converters:
> 
> https://github.com/patrick-kidger/equinox/blob/adaaf161d9bff64f16dc924f61bc6884a884ca7a/equinox/_module/_module.py#L364-L386
> 
> Is this not working for your use-case?

I did not know you were reading types from the converter!
But it does not seem to work (beartype uses the type from the class attribute), see:

```python
import equinox as eqx
import jax.numpy as jnp
import numpy as np
from beartype import beartype as typechecker
from jaxtyping import Array, ArrayLike, Float, jaxtyped


def asarray(x: ArrayLike) -> Array:
    return jnp.asarray(x)


@jaxtyped(typechecker=typechecker)
class Test(eqx.Module):
    x: Float[Array, "3"] = eqx.field(
        converter=asarray
    )  # Using jnp.asarray also fails with np.array


# print(Test(x=jnp.array([1.0, 2.0, 3.0])))  # This works
print(Test(x=np.array([1.0, 2.0, 3.0])))  # This fails
```

outputs:

```
signature = <Signature (x: Union[jax.Array, numpy.ndarray, numpy.bool_, numpy.number, bool, int, float, complex]) -> jax.Array>
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 449, in wrapped_fn_impl
    param_fn(*args, **kwargs)
  File "<@beartype(__main__.Test) at 0x7bba57ff8540>", line 29, in Test
beartype.roar.BeartypeCallHintParamViolation: Method __main__.Test() parameter x="array([1., 2., 3.])" violates type hint <class 'jaxtyping.Float[Array, '3']'>, as this value is not an instance of the underlying array type <class 'jax.Array'>.

During handling of the above exception, another exception occurred:

  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 814, in _get_problem_arg
    fn(*args, **kwargs)
  File "<@beartype(__main__.check_single_arg) at 0x7bba57ff9760>", line 29, in check_single_arg
beartype.roar.BeartypeCallHintParamViolation: Method __main__.check_single_arg() parameter x="array([1., 2., 3.])" violates type hint <class 'jaxtyping.Float[Array, '3']'>, as this value is not an instance of the underlying array type <class 'jax.Array'>.

The above exception was the direct cause of the following exception:

  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 454, in wrapped_fn_impl
    argmsg = _get_problem_arg(
             ^^^^^^^^^^^^^^^^^
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 817, in _get_problem_arg
    raise TypeCheckError(
jaxtyping.TypeCheckError: 
The problem arose whilst typechecking parameter 'x'.
Actual value: f64[3](numpy)
Expected type: <class 'Float[Array, '3']'>.

The above exception was the direct cause of the following exception:

  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/test.py", line 20, in <module>
    print(Test(x=np.array([1.0, 2.0, 3.0])))  # This fails
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/equinox/_module/_module.py", line 412, in __call__
    self = super().__call__(*args, **kwargs)  # pyright: ignore[reportAttributeAccessIssue]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/equinox/_module/_better_abstract.py", line 232, in __call__
    self = super().__call__(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 319, in __init__
    _check_dataclass_annotations(self, typechecker)
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 624, in _check_dataclass_annotations
    f(self, **values)
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 562, in wrapped_fn
    return wrapped_fn_impl(args, kwargs, bound, memos)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/export/home/eertmans/repositories/DiffeRT/DiffeRT/.venv/lib/python3.11/site-packages/jaxtyping/_decorator.py", line 483, in wrapped_fn_impl
    raise TypeCheckError(msg) from e
jaxtyping.TypeCheckError: Type-check error whilst checking the parameters of __main__.Test.
The problem arose whilst typechecking parameter 'x'.
Actual value: f64[3](numpy)
Expected type: <class 'Float[Array, '3']'>.
----------------------
Called with parameters: {'self': Test(...), 'x': f64[3](numpy)}
Parameter annotations: (self: Any, x: Float[Array, '3']) -> Any.
```

equinox/_module/_module.py
nstarman: ... Or `parent_cls.__dict__.get("__check_init__")) is not None:`

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+                "hold, then consider using `__check_init__` instead. This is an "
+                "Equinox-specific extension that is always ran. See here for more "
+                "details: "
+                "https://docs.kidger.site/equinox/api/module/advanced_fields/#checking-invariants",  # noqa: E501
+                stacklevel=2,
+            )
+        if has_dataclass_init:
+            init_doc = cls.__init__.__doc__
+
+        cls = better_dataclass(eq=False, repr=False, init=has_dataclass_init)(cls)
```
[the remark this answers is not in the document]
jeertmans: I am not sure whether this is possible, but the ideal solution (to me) would be that typechecker can both check that (1) the input value of the converter is of correct type and (2) the output value of the converter matches the type defined by the class attribute. After some testings, it looks like only (1) is performed. The issue is that converters like `jnp.asarray` are very loosely typed, which does not perform the intended checks.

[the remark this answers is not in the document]
NeilGirdhar: (1) is by far more important, right?

The most popular converters are probably:
* `jnp.asarray`, and
* various boxing equinox modules that have one field and mark the subtree in some way.

In both cases, the output value is probably right.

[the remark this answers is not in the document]
jeertmans: > (1) is by far more important, right?

I don't agree with this, especially as converters like `jnp.asarray` accept anything, i.e., `Any`, and return `Array`.

Type annotations on class attributes should describe the exact type the class will hold, so the type checker should check that the final results, i.e., the return value of the converter, matches this type (in my opinion).

My current solution is to write a custom converter for each class attribute, e.g.,

```python
def asarray(x: Any) -> Float[Array, "3"]:
    return jnp.asarray(x)


@jaxtyped(typechecker=typechecker)
class Test(eqx.Module):
    x: Float[Array, "3"] = eqx.field(
        converter=asarray
    )
```
but this adds quite a lot of lines of code.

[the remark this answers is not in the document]
NeilGirdhar: > I don't agree with this, especially as converters like jnp.asarray accept anything, i.e., Any, and return Array.

That probably should be fixed in the Jax repo.

equinox/_module/_module.py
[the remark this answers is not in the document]
patrick-kidger: So I've just checked https://docs.python.org/3/library/dataclasses.html and I don't see `__dataclass_params__` listed in the public API. I'm tempted to avoid using it for that reason. Do you have any reference for this?

patrick-kidger:
@jeertmans which library versions are you using to get the error in https://github.com/patrick-kidger/equinox/pull/1028#issuecomment-2890868558 ? With `equinox` set to this branch, `jaxtyping==0.3.2`, `beartype==0.20.2` then I don't get any error with the snippet you've attached.

equinox/_module/_prebuilt.py
```diff
@@ -0,0 +1,100 @@
...
+
+        - `*args`: the arguments to apply. Passed after those arguments passed during
+            `__init__`.
+        - `**kwargs`: any keyword arguments to apply.
+
+        **Returns:**
+
+        The result of the wrapped function.
+        """
+        return self.func(*self.args, *args, **kwargs, **self.keywords)
```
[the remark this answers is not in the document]
nstarman: Was this meant to be marked as resolved?

lockwo:
> in case you'd like to try out these changes on your own use-cases

So far so good on my end (although I think most of my use cases are applications of modules rather than anything too fancy within modules themselves)

> Dropped support for class Foo(eqx.Module, strict=True). In the end 'strict modules' were never documented nor graduated into the public API, as I wasn't comfortable with the idea of a two-tier 'module' vs 'strict module' system. If I was to write Equinox again from scratch then they'd be the default, but as it is this feature isn't useful enough to justify a breaking change, nor to keep around on an optional basis for all five of us who really want this.

me_irl: https://www.youtube.com/watch?v=17KmNrG9pE4

I was a big enough fan that I even had some code for doing the same thing but for non equinox modules (which I just pushed, so maybe 4 other people can use it https://github.com/lockwo/ihoop, very rough draft, but I have it working with my all `strict=True` library https://github.com/lockwo/distreqx).

NeilGirdhar:
> I was a big enough fan that I even had some code for doing the same thing but for non equinox modules.

Can you not just implement this in a subclass, and use that instead?

equinox/_module/_module.py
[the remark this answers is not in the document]
johannahaffner: > My current solution is to write a custom converter for each class attribute, e.g.,

We also do this here: 

https://github.com/patrick-kidger/optimistix/blob/9927984fb8cbec77f9514fad7af076dce64e3993/optimistix/_solver/learning_rate.py#L12

This was motivated not by type checker complaints, but by wanting cleaner documentation. Before the custom converter, our docs looked like `learning_rate: Union[jnp.ndarray, np.ndarray, int, float, bool, ...]`. I'm not sure how related docgen and type checking are here, but just in case this gets tweaked here, I wanted to mention it.

equinox/_module/_prebuilt.py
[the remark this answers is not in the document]
patrick-kidger: Yup! I just pushed a new commit with this in.

lockwo:
> Can you not just implement this in a subclass, and use that instead?

I just did the same meta class approach: https://github.com/lockwo/ihoop/blob/main/ihoop/strict.py

> And honestly I think it's reasonable that this not be the default

That's true, I think the frozen classes are already a bit much for some people lol

patrick-kidger:
From your library:

> Do you like python, but wish it is was more like Julia? Do like Julia, but wish it was a language other people used? Did you read Design Patterns: Elements of Reusable Object-Oriented Software, but wish you didn't?

Pahahaha 😁

jeertmans:
> @jeertmans which library versions are you using to get the error in [#1028 (comment)](https://github.com/patrick-kidger/equinox/pull/1028#issuecomment-2890868558) ? With `equinox` set to this branch, `jaxtyping==0.3.2`, `beartype==0.20.2` then I don't get any error with the snippet you've attached.

Yep you're right, I probably did something wrong with package versions, it is working as expected (i.e., the type checker now checks the type from the first parameter of the converter).

As discussed in https://github.com/patrick-kidger/equinox/pull/1028#discussion_r2097159524, we still have the issue that neither the output type of the converter function, nor the declared type of the attribute are checked (see below). However, I am not sure if we can do something about this here.

```python
# /// script
# dependencies = [
#   "beartype==0.20.2",
#   "equinox@git+https://github.com/patrick-kidger/equinox.git@new-module",
#   "jaxtyping==0.3.2",
# ]
# ///

import equinox as eqx
import jax.numpy as jnp
import numpy as np
from beartype import beartype as typechecker
from jaxtyping import Array, ArrayLike, Float, jaxtyped


def asarray(x: ArrayLike) -> Array:
    return x  # We should transform x into an Array


@jaxtyped(typechecker=typechecker)
class Test(eqx.Module):
    x: Float[Array, "3"] = eqx.field(
        converter=asarray
    )


print(Test(x=jnp.array([1.0, 2.0, 3.0])))  # This works
# prints: Test(x=f32[3])
print(Test(x=np.array([1.0, 2.0, 3.0])))  # This works,
# but should fail as the return value of `asarray` is not an Float[Array, "3"], not even an Array
# prints: Test(x=f64[3](numpy))
```

patrick-kidger:
Okay, great! I think that means everything outstanding on this has been tackled, so merged. (Any late-breaking test failures do please still report them here, we haven't merged dev->main yet.)

As for checking the attributes on assignment, this is really a jaxtyping thing. (Equinox has no opinion about type-checking.) This is something that changed in jaxtyping a few releases ago: it used to be that we did not check `__init__` arguments and did check attributes, but we then switched that around. The reason is essentially code simplicity; checking attributes involved some complicated monkey-patching of the `__init__` method and did not gel well with `eqx.field(converter=...)`. I don't think checking that is something that's on the roadmap I'm afraid!

johannahaffner:
What is your timeline for the next equinox release? Optimistix uses `strict=True` all over the place :)

patrick-kidger:
> What is your timeline for the next equinox release?

'When it's ready' 😁 

> Optimistix uses strict=True all over the place :)

Yup, and that should be fine -- this is backwards-compatible with those, it just ignores them now.

equinox/_module/_module.py
```diff
@@ -0,0 +1,617 @@
...
+
+    def __repr__(self):
+        return tree_pformat(self)
+
+    def __hash__(self) -> int:
+        return hash(
+            tuple(
+                (f.name, getattr(self, f.name)) for f in dataclasses.fields(type(self))
+            )
+        )
```
leademeule, line 476: I updated from equinox 0.12.1 to 0.13 and found myself confronted to an exception as I tried to `jax.jit` a `Classifier` that subclasses `equinox.Module`.  I traced this to a nested module within `Classifier` that possesses unhashable attributes. Once `Classifier` is partitioned into parameters and static arguments and passed to `jax.jit`, the static part fails to hash in the highlighted code. 

I would suggest adding a warning or a more informative exception message concerning unhashable attributes since they were allowed in previous versions, and since they do not trigger an exception message that identifies the guilty attributes the moment. Now, calling `jax.jit` simply triggers an exception of the form `Non-hashable static arguments are not supported. An error occurred while trying to hash an object of type Classifier`, followed by the entire module pretty-printed, by a traceback leading to the highlighted code, and by a `TypeError` stating _only_ the type of the offending attribute.

patrick-kidger: Can you provide a MWE, including the version of JAX you are using?

(I don't think it's due to the highlighted section here, as unhashable attributes have never been permitted.)

leademeule: Here is a minimal reproduction for the issue:

```
from typing import Dict, Tuple
import sys
import jax
import equinox

print(f"Version of Python:  {sys.version}")
print(f"Version of Equinox: {equinox.__version__}")
print(f"Version of JAX:     {jax.__version__}")

class Problematic(equinox.Module):
    nested: Dict[int, equinox.Module]

    def __init__(self, nested: Dict[int, equinox.Module]):
        super().__init__()
        self.nested = nested

    def __call__(self, array: jax.Array) -> jax.Array:
        for key in sorted(self.nested.keys()):
            array = self.nested[key](array)

        return array


model = Problematic({0: equinox.nn.Linear(10, 10, key=jax.random.key(0))})

array = jax.numpy.ones((10,))

try:
    _ = model(array)
    print("Calling the model directly succeeded.")
except Exception as exception:
    print("Calling the model directly failed:")
    print(exception)


def split(
    model: equinox.Module,
) -> Tuple[equinox.Module, equinox.Module]:
    return equinox.partition(model, equinox.is_array)


def merge(
    array: jax.Array, parameters: equinox.Module, static: equinox.Module
) -> jax.Array:
    return equinox.combine(parameters, static)(array)


try:
    _ = jax.jit(merge, static_argnames="static")(array, *split(model))
    print("Calling the model after filtering and jitting succeeded.")
except Exception as exception:
    print("Calling the model after filtering and jitting failed:")
    print(exception)
```

And below are two outputs across `equinox 0.12.1` and `equinox 0.13.0` showing the regression:

```
Version of Python:  3.12.9 (main, Feb  5 2025, 18:58:23) [Clang 19.1.6 ]
Version of Equinox: 0.12.1
Version of JAX:     0.5.0
Calling the model directly succeeded.
Calling the model after filtering and jitting succeeded.
```

```
Version of Python:  3.12.9 (main, Feb  5 2025, 18:58:23) [Clang 19.1.6 ]
Version of Equinox: 0.13.0
Version of JAX:     0.5.0
Calling the model directly succeeded.
Calling the model after filtering and jitting failed:
Non-hashable static arguments are not supported. An error occurred while trying to hash an object of type <class '__main__.Problematic'>, Problematic(
  nested={
    0:
    Linear(
      weight=None, bias=None, in_features=10, out_features=10, use_bias=True
    )
  }
). The error was:
  File "/Users/lea/Code/equinox-issue/unhashable.py", line 49, in <module>
  File "/Users/lea/Code/equinox-issue/.venv/lib/python3.12/site-packages/equinox/_module/_module.py", line 516, in __hash__
TypeError: unhashable type: 'dict'
```

lockwo: Looks like because now the hash is based on the value of the fields, whereas before it was a hash based on the leaves of the pytree (https://github.com/patrick-kidger/equinox/blob/0-4-33-errors-fix/equinox/_module.py#L1073)

Also here's an even simpler MVP

```python
import sys
import jax
import equinox as eqx

print(f"Version of Python:  {sys.version}")
print(f"Version of Equinox: {eqx.__version__}")
print(f"Version of JAX:     {jax.__version__}")

class Problematic(eqx.Module):
    nested: dict

    def __init__(self, nested):
        self.nested = nested

    def __call__(self):
        return 1.0


model = Problematic({0: 0.0})


try:
    _ = model()
    print("Calling the model directly succeeded.")
except Exception as exception:
    print("Calling the model directly failed:")
    print(exception)


def merge(parameters, static):
    return eqx.combine(parameters, static)()

arr, static = eqx.partition(model, eqx.is_array)

try:
    _ = jax.jit(merge, static_argnames="static")(arr, static)
    print("Calling the model after filtering and jitting succeeded.")
except Exception as exception:
    print("Calling the model after filtering and jitting failed:")
    print(exception)
```

PR #1039, opened 2025-06-16
Make parameters of jaxpr equations hashable.

hawkinsp:
In an upcoming JAX change we will add caching around abstract evaluation rules in JAX. This requires that parameters to jaxpr equations have a hash and equality defined.

Here, use tuples instead of dicts in jaxpr parameters.


hawkinsp:
I think the pyright errors aren't from my PR?

patrick-kidger:
LGTM, thank you! Merged. (To `dev`, LMK if you need a new release on `main` in advance of the new JAX changes.)

Do you have a JAX commit with the new behaviour, that I can try running the Lineax+Optimistix+Diffrax tests on top of? A lot of ecosystem here calls [`lineax.linear_solve_p`](https://github.com/patrick-kidger/lineax/blob/66b7b5327a44e4b944a8ce9242773150e8a8d811/lineax/_solve.py#L308), which is complicated enough that I could imagine something going wrong.

hawkinsp:
> LGTM, thank you! Merged. (To `dev`, LMK if you need a new release on `main` in advance of the new JAX changes.)
> 
> Do you have a JAX commit with the new behaviour, that I can try running the Lineax+Optimistix+Diffrax tests on top of? A lot of ecosystem here calls [`lineax.linear_solve_p`](https://github.com/patrick-kidger/lineax/blob/66b7b5327a44e4b944a8ce9242773150e8a8d811/lineax/_solve.py#L308), which is complicated enough that I could imagine something going wrong.

Sure, the PR is https://github.com/jax-ml/jax/pull/29498 . 

(In general we're doing this to improve trace times. No need to run abstract eval or lowering rules repeatedly if you're doing the exact same thing.)

This PR seemed to be enough for the tests I was able to run internally but I could imagine that maybe not all of your downstream tests are run or not all of the versions are up to date.

patrick-kidger:
> This PR seemed to be enough for the tests I was able to run internally but I could imagine that maybe not all of your downstream tests are run or not all of the versions are up to date.

Ty! I didn't know if you still had Diffrax/Optimistix/Lineax checked in internally. (?) I'll give things a run just to be sure.

hawkinsp:
I see you merged this into your `dev` branch. It turns out it would actually be rather helpful if you would release this in some form soonish either on a more stable branch or as a numbered release. I'm guessing `dev` isn't safe for end users?

patrick-kidger:
`dev` should be safe to use. I just use it when I have new features leading to doc updates; changes to `main` are configured to eagerly update the docs (so as to make between-releases doc fixes easy).

johannahaffner:
Lineax and Optimistix test suites run without errata. Diffrax has lots of warnings, but they are unrelated and predate this change (Tsit5 not marked as converging to Ito solution, complex numbers). So seems we're all set and forward compatible!

patrick-kidger:
Awesome - thank you @johannahaffner !

Issue #1050, opened 2025-07-19
Improving readability of state indices by using key paths


fhchl:
Hej! How about improving the readability of states when printed? Currently, if I print a state object after `eqx.nn.make_with_state`, the keys are just some arbitrary hex of the id of an integer, e.g.

    State(0xa42648=weak_i32[])

If there are a lot of them, it's impossible to understand which are which. To look at the values, one can look at the `state._state` dictionary with `int`s as keys, e.g.

    {0: Array(0, dtype=int32, weak_type=True)}

That also won't help figuring out which state has which value. Of course, one can use `state.get(some_index)` to get a specific value for some specific state.

But how about replacing that integer key with a stringified KeyPath? Then the result could look like

    State(TopModule.stateful_mod.counter_index=weak_i32[])
    # or 
    {'TopModule.stateful_mod.counter_index': Array(0, dtype=int32, weak_type=True)}

Much easier to read!

Here is a colab with the idea: https://colab.research.google.com/drive/1L1dk1UpORf-DrIUHPuBHDUJRF8SSSFVm?usp=sharing

I'd be happy to make a PR for this.

patrick-kidger:
I like this idea! I'd be very happy to take a PR on it.

patrick-kidger:
Fixed in #1052!

PR #1052, opened 2025-07-24
use keypaths instead of ints for state keys

fhchl:
See #1050.

With this change

```python
import equinox as eqx

class StatefulModule(eqx.Module):
    counter_index: eqx.nn.StateIndex

    def __init__(self):
        counter_init = 0
        self.counter_index = eqx.nn.StateIndex(counter_init)

    def __call__(self, x, state):
        counter = state.get(self.counter_index)
        counter = counter + 1
        state = state.set(self.counter_index, counter)
        return x, state


class TopModule(eqx.Module):
    stateful_mod: eqx.Module

    def __init__(self):
        self.stateful_mod = StatefulModule()

    def __call__(self, x, state):
       return self.stateful_mod(x, state)

proc, state = eqx.nn.make_with_state(TopModule)()
state, state._state, state.get(proc.stateful_mod.counter_index)
```

prints

```
(State(TopModule.stateful_mod.counter_index=weak_i32[]), {'TopModule.stateful_mod.counter_index': Array(0, dtype=int32, weak_type=True)}, Array(0, dtype=int32, weak_type=True))
```

### DIFF
--- a/equinox/nn/_stateful.py
+++ b/equinox/nn/_stateful.py
@@ -55,9 +55,9 @@ class StateIndex(Module, Generic[_Value]):
     [`equinox.nn.BatchNorm`][] for further reference.
     """  # noqa: E501
 
-    # Starts off as None when initialised; later replaced with an `int` inside
-    # `make_with_state`.
-    marker: object | int = field(static=True)
+    # Starts off as object when initialised; later replaced with stringified KeyPath
+    # inside `make_with_state`.
+    marker: object | str = field(static=True)
     init: _Value | _Sentinel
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Awesome! Thanks for the PR. It looks like the pre-commit hooks are failing (see CONTRIBUTING.md), but otherwise this basically LGTM :)

fhchl:
Woops, fixed!

patrick-kidger:
Awesome, LGTM! Thank you for the contribution :)

PR #1054, opened 2025-07-25
Crash fix for JAX 0.7.0

patrick-kidger:


johannahaffner:
Where did this jaxplosion occur?

patrick-kidger:
See test failures for #1053 for example.

(Besides which the tests here are failing due to https://github.com/jax-ml/jax/issues/30517, which presumably we'll need to wait for JAX to update to fix.)

patrick-kidger:
Merging as the content of this PR is itself correct, and we're just waiting on JAX to fix their segfault.

PR #1065, opened 2025-08-04
chore(docs): improve `tree_at` documentation

### LINKED ISSUE #1061: Doc. issue / feature request: `tree_at` (and related) do not return a deep copy of the input tree
Hi!

I recently noticed that `equinox.tree_at` (and thus `equinox.nn.inference_mode`) does not actually return a true deep copy of the objects. While I think this is fine and should be the default, it can be surprising when used with `jax.jit` and donating arguments, as you suggest in some examples (e.g., `equinox.filter_jit(donate="all")`).

Therefore, I think the documentation is misleading by saying, "Returns: A copy of ..." because one might expect a real copy.

In JAX's documentation, `jax.tree.map` is said to return a "new object," which is slightly more accurate (as the object's attributes can be a view of the original object).

If I may, I have two suggestions:

1. Add an optional keyword argument, e.g., `deepcopy: bool = False`, to allow users to request an actual copy. Alternatively, document that copying a PyTree is as simple as
   `jax.tree.map(lambda x: x.copy() if equinox.is_array(x) else x, pytree)`.
2. Second, rephrase the function documentation, e.g., instead of
"Returns: A copy of the input PyTree with the appropriate modifications."
write
"Returns: A new PyTree with the same structure as the input PyTree and the appropriate modifications.
Note that this only performs a shallow copy of the input PyTree. If you need to return an actual copy that you can safely pass to a function with donated arguments, specify `deepcopy=True`."

The second part of the paragraph would be useful if an optional deepcopy argument is implemented as part of `equinox.tree_at`.

What do you think?

[^1]: Naming it `deepcopy` might not be the best choice if it only copies arrays and not static attributes, but that can be changed, of course.

jeertmans:
Closes #1061

### DIFF
--- a/equinox/_tree.py
+++ b/equinox/_tree.py
@@ -92,7 +92,17 @@ def tree_at(
 
     **Returns:**
 
-    A copy of the input PyTree, with the appropriate modifications.
+    A new PyTree with the same structure as the input PyTree and the appropriate modifications.
+
+    !!! Important
+
+        The returned PyTree is only a shallow copy of the input PyTree, with the
+        specified nodes replaced. This means that the input PyTree cannot be reused
+        after using the output PyTree as a *donated* argument to a JAX transform.
+
+        To reuse the input PyTree, you must make a copy of its arrays first, e.g.:
+        `copied_pytree = jax.tree.map(lambda x: x.copy() if eqx.is_array(x) else x, pytree)`,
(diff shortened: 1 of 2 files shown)


patrick-kidger:
Awesome, LGTM! Thank you for the contribution :)

PR #1072, opened 2025-08-13
Enable pyright to catch errors on accessing non-existent attributes of a Module.

nisheethlahoti:
Since equinox 0.13, the custom `__getattribute__` and `__setattr__` methods on `Module` cause pyright to decide it doesn't know what attributes can be accessed anymore.  e.g. if we have the code:
```
import equinox as eqx

class Foo(eqx.Module):
    bar: int

foo = Foo(3)
foo.nonexistent
foo.nonexistent = 4
```

Then until 0.12, pyright would show error diagnostics on the last two lines. Now it doesn't.

This PR fixes this by hiding these custom methods from static type-checkers.

### DIFF
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -6,7 +6,7 @@ import types
 import warnings
 import weakref
 from collections.abc import Callable, Hashable
-from typing import Any, cast, Final, ParamSpec, TypeVar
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Awesome! This LGTM, and thank you for the fix :)

Issue #1077, opened 2025-08-16
in_shardings and out_shardings cannot be passed to eqx.filter_jit


mjo22:
Hi, I’ve been discussing in #1076 and #1067 some thoughts for supporting automatic parallelization via `eqx.filter_jit` with something like the following interface:

```python
s, r = sharding, replicated

@eqx.filter_jit(donate=“all”, in_shardings=(eqx.if_array(r), eqx.if_array(s)), out_shardings=(eqx.if_array(r),)
def train_step(model_opt_state, x_y):
    model, opt_state = model_opt_state
    x, y = x_y
    …
```

I wanted to open a separate issue because I noticed that `eqx.filter_pmap` documentation says “using [equinox.filter_jit](https://docs.kidger.site/equinox/api/transformations/#equinox.filter_jit) with sharded inputs is now recommended over filter_pmap.”

I tried this and got some strange errors, and from looking at the code I realized this can’t actually be done in practice right now because the order of arguments passed ultimately to `jax.jit` is shuffled because of how the donate interface works. To achieve this recommendation, I think an explicit update is required. And further, I think it would be great to add the niceness of `eqx.filter_vmap` `AxisSpec`s.

patrick-kidger:
Picking up the conversation from #1076: this was intentional in the design of `eqx.filter_jit` - as `{in,out}_shardings` were considered soft-deprecated, it didn't matter that there was no way to pass them.

If that's changed then we may need to adjust this.

michael-0brien:
Closing this as per discussion in #1076

Issue #1067, opened 2025-08-06
Update autoparallelism tutorial


michael-0brien:
Hi, wanted to report it looks like some features used in the [autoparallelism tutorial](https://docs.kidger.site/equinox/examples/parallelism/#autoparallelism-eg-multi-gpu) are deprecated in JAX 0.7.0, namely the `PositionalSharding`.

patrick-kidger:
Thanks for the spot! The sharding capabilities aren't something I use super often, so if you're already familiar then I'd be happy to take a PR updating this?

michael-0brien:
I am just starting to learn about this; I can give updating a shot sometime in the next few weeks!

michael-0brien:
I just gave this a shot, and it's not totally clear to me what the best thing to do is. It's difficult to exactly reproduce the tutorial because `sharding.replicate()` method does not exist anymore and I'm struggling to find the alternative. However, I think this is symptomatic that JAX's current recommendations are no longer focused on `jax.lax.with_sharding_constraint`.

Keeping in line with their API, it may be best for this to be a larger PR expanding on #978, once the dust settles with all of the sharding API changes (it seems like this may be the case now)!

From looking through the JAX documentation, it seems to me an equinox-like extension to JAX sharding would be

1) Add support to `eqx.filter_jit` for filtering `jax.jit`'s `in_shardings` and `out_shardings` args (could follow a similar pattern as `eqx.filter_vmap`?). This encompasses `eqx.filter_shard` for most users.
2) Add a `eqx.filter_device_put`. This is the "eager" version of what is currently `eqx.filter_shard`.
3) Add a `eqx.filter_shard_map`.

Of course there are many important considerations here. For example, it seems like `eqx.filter_shard` may accomplish 1) and 2), however JAX's recommendations say to use `device_put` in eager regions and `jax.lax.with_sharding_constraint` inside jitted ones *only* to control intermediate arguments (i.e. not the inputs or the outputs of the function, see (https://docs.jax.dev/en/latest/_autosummary/jax.lax.with_sharding_constraint.html)). Otherwise, the recommendation is to use `jax.jit`'s `in_sharding` and `out_sharding` arguments.

Of course there are other important considerations too, like backwards compatibility and whether or not the developers want to maintain a `eqx.filter_*` API for sharding, etc etc. My main observation is that if an API is to be maintained, based on recent developments `eqx.filter_shard` seems less in line with `equinox`'s philosophy. Let me know what you think!

[...]

Issue #1076, opened 2025-08-14
Double use of 'eqx.filter_shard' required?


miltonllera:
First of all, great job with the library. I find it much more readable than the alternative.

To my question, in the auto-parallelism example, `eqx.filter_shard` is used both outside and inside the jitted `train_step` function. Is this really needed?

Here is the link: https://docs.kidger.site/equinox/examples/parallelism/

I ran my code without using the sharding-related calls inside `train_step` and it works, so not sure if I am missing something here.

michael-0brien:
Hi, see #1067. I am currently working on updating the tutorial (some features are deprecated in JAX 0.7.0). I am not totally sure but it seems to me that `eqx.filter_shard` was implemented when there was a fast-changing API / recommendations for sharding.

Under the hood, `eqx.filter_shard` calls `jax.lax.with_sharding_constraint`. JAX's recommendations now are to only use this function for intermediate arguments inside jitted regions. The recommendations for the behavior in the notebook now seem to be through `jax.jit`'s `in_shardings` and `out_shardings` arguments (additionally with `jax.device_put` outside of jitted regions for more control). Though if I understand correctly, `eqx.filter_shard` can be used to achieve the same thing as `in_shardings` and `out_shardings` inside jitted regions.

Anyone please correct me if there are any details here that are not accurate!

michael-0brien:
If `jax.lax.with_sharding_constraint` outside of jitted regions does the same thing as `jax.device_put`, then I don't think there is a reason to call `eqx.filter_shard` both outside jitted regions and at the beginning of the `train_step` function. See here for the reason: https://github.com/jax-ml/jax/discussions/30869. After `jax.device_put` is called outside jitted regions, arrays are committed to that device and jit will maintain this. I am not totally sure if `eqx.filter_shard` in its current state does the same thing as `jax.device_put` though. I haven't fully tested this though; this is just my current understanding of things.

miltonllera:
Okay based on your comments I went and ran a simple example:

```
sharded = jax.lax.with_sharding_constraint(inputs, jax.sharding.NamedSharding(jax.make_mesh((2,), ('batch',)), jax.sharding.PartitionSpec('batch',)))
sharded.sharding
```

which gives as output:
> NamedSharding(mesh=Mesh('batch': 2, axis_types=(Auto,)), spec=PartitionSpec('batch',), memory_kind=device)

[...]

PR #1078, opened 2025-08-18
Autoparallelism tutorial update

michael-0brien:
I've updated the autoparallelism tutorial (#1067) to remove the deprecated syntax as of JAX 0.7.0. I had some questions about how exactly to update the tutorial in this issue, but I decided to go ahead and do it for what seems to be correct as per JAX for the current state of equinox. The one thing I am not sure about is if `eqx.filter_shard` via `jax.lax.with_sharding_constraint` can really be used as `jax.device_put`. Therefore, in the updated tutorial I only call `eqx.filter_shard` inside the jitted region.

If we wanted to instead apply the sharding in eager regions, then I would recommend we add a `eqx.filter_device_put`. If we do this, it is not necessary to call `eqx.filter_shard` on the inputs of the `train_step` function (just the outputs).

It seemed to me that addressing #1077 would really be the best thing to do. However, from looking through the code it looks like there would be a relatively large refactor of the `eqx.filter_jit` internals to do this; the `_filter_jit_cache` is called before the arguments are preprocessed, but it would have to be the other way around in order to implement what I proposed. I imagine there is a good reason why this is the case too, so I wanted to leave it alone.

### DIFF
--- a/equinox/_sharding.py
+++ b/equinox/_sharding.py
@@ -1,7 +1,6 @@
 from typing import Any
 
 import jax
-import jax.core
 import jax.lax as lax
 from jaxtyping import PyTree
 
--- a/tests/test_sharding.py
+++ b/tests/test_sharding.py
@@ -5,23 +5,19 @@ import jax.random as jr
 import jax.sharding as jshard
 import jax.tree_util as jtu
 
-jax.config.update("jax_platform_name", "cpu")
-jax.config.update("jax_num_cpu_devices", 2)
-
-
 
 def test_sharding_no_inside_jit():
-    is_committed = lambda pytree: jax.tree.reduce(lambda x, y: x and y, jax.tree.map(lambda x: x.committed, pytree))
-    is_sharded = lambda pytree, sharding: jax.tree.reduce(lambda x, y: x and y, jax.tree.map(lambda x: eqx.tree_equal(x.sharding, sharding), pytree))
-
     mlp = eqx.nn.MLP(2, 2, 2, 2, key=jr.PRNGKey(0))
 
-    num_devices = 2
+    num_devices = 1
     mesh = jax.make_mesh((num_devices,), ("x",))
     sharding = jshard.NamedSharding(mesh, jshard.PartitionSpec("x"))
     sharded_mlp = eqx.filter_shard(mlp, sharding)
-    assert is_committed(eqx.filter(sharded_mlp, eqx.is_array))
-    assert is_sharded(eqx.filter(sharded_mlp, eqx.is_array), sharding)
+    # Known bug, not resolved until JAX 0.7.2. Eager behavior of `eqx.filter_shard`
+    # does not commit arrays to the sharding if does not change the original
+    # array (i.e. it does not result in a copy): https://github.com/jax-ml/jax/issues/31199
+    # assert is_committed(eqx.filter(sharded_mlp, eqx.is_array))
+    # assert is_sharded(eqx.filter(sharded_mlp, eqx.is_array), sharding)
 
     @eqx.filter_jit
     def f(x):
@@ -30,22 +26,20 @@ def test_sharding_no_inside_jit():
         x = eqx.combine(a, b)
         return x
 
-    out = f(sharded_mlp)
+    _ = f(sharded_mlp)
 
-    assert is_sharded(eqx.filter(out, eqx.is_array), sharding)
+    # assert is_sharded(eqx.filter(out, eqx.is_array), sharding)
+
 
 def test_sharding_only_inside_jit():
-    is_sharded = lambda pytree, sharding: jax.tree.reduce(lambda x, y: x and y, jax.tree.map(lambda x: x.sharding == sharding, pytree))
-
     # Make sharding
-    devices = jax.local_devices(backend="cpu")
-    num_devices = len(devices)
+    num_devices = 1
     mesh = jax.make_mesh((num_devices,), ("x",))
     sharding = jshard.NamedSharding(mesh, jshard.PartitionSpec("x"))
     # Make dummy pytree
-    shape = (num_devices,)
+    shape = (10 * num_devices,)
     pytree = ((jnp.zeros(shape), 4), (jnp.zeros(shape),), "something_static")
-    assert not is_sharded(eqx.filter(pytree, eqx.is_array), sharding)
+    # assert not _is_sharded(eqx.filter(pytree, eqx.is_array), sharding)
 
     @eqx.filter_jit
     def f(x):
@@ -55,6 +49,18 @@ def test_sharding_only_inside_jit():
         return eqx.filter_shard(x, sharding)
 
     # Evaluate jitted computation
-    out = f(pytree)
+    _ = f(pytree)
 
-    assert is_sharded(eqx.filter(out, eqx.is_array), sharding)
+    # assert is_sharded(eqx.filter(out, eqx.is_array), sharding)
+
+
+def _is_sharded(pytree, sharding):
+    return jax.tree.reduce(
+        lambda x, y: x and y, jax.tree.map(lambda x: x.sharding == sharding, pytree)
+    )
+
+
+def _is_committed(pytree):
+    return jax.tree.reduce(
+        lambda x, y: x and y, jax.tree.map(lambda x: x.committed, pytree)
+    )


michael-0brien:
Not sure the best way to process as per conversations in #1076 and #1077. It seems like some clarification is required for JAX 0.7.0, but if we want a minimal change for now we could simply fix the deprecated parts of the tutorial!

michael-0brien:
Hi, see updates to `test_sharding.py`. Looks like `eqx.filter_shard` as implemented now does not work as expected in line with this discussion: https://github.com/jax-ml/jax/discussions/30869. I've tested if `x.committed` is `True` after running `eqx.filter_shard`. This test passes when `jax.device_put` replaces `jax.lax.with_sharding_constraint`. I suspect that if `x.committed = False`, then shardings are not respected by the compiler. This is consistent with JAX's documentation to only use `jax.lax.with_sharding_constraint` inside jitted regions: https://docs.jax.dev/en/latest/_autosummary/jax.lax.with_sharding_constraint.html.

The JAX version for this test is v0.7.0.

michael-0brien:
Okay here is the following observation:

```python
import jax
import numpy as np
import jax.numpy as jnp

sharding = jax.sharding.SingleDeviceSharding(jax.devices()[0])

x = np.arange(10)  # numpy array
y_wsc = jax.lax.with_sharding_constraint(x, sharding)
y_put = jax.device_put(x, sharding)
assert y_wsc.committed == y_put.committed == True  # This passes

x = jnp.arange(10)  # jax array
y_wsc = jax.lax.with_sharding_constraint(x, sharding)
y_put = jax.device_put(x, sharding)
assert y_wsc.committed == y_put.committed == True  # This fails; y_wsc.committed = False
```

miltonllera:
This is interesting. What would be the implication of the `y_wsc` not being committed?

michael-0brien:
See the discussion I link to above in the JAX discussions page. Strangely, this does pass:

```python
import jax
import jax.sharding as jshard
import numpy as np
import jax.numpy as jnp

jax.config.update("jax_num_cpu_devices", 2)

num_devices = len(jax.devices())
mesh = jax.make_mesh((num_devices,), ("batch",))
sharding = jshard.NamedSharding(mesh, jshard.PartitionSpec("batch"))

x = np.arange(10)  # numpy array
y_wsc = jax.lax.with_sharding_constraint(x, sharding)
y_put = jax.device_put(x, sharding)
assert y_wsc.committed == y_put.committed == True  # This passes

x = jnp.arange(10)  # jax array
y_wsc = jax.lax.with_sharding_constraint(x, sharding)
y_put = jax.device_put(x, sharding)
assert y_wsc.committed == y_put.committed == True  # This passes too!
```

Whether or not this is okay I think depends on whether or not there really is a notion of a "compiler hint" in eager mode. I am not sure about that.

miltonllera:
Okay, this makes sense since, when I ran a similar experiment, I found that the `wsc`-sharded array was also committed. See [here](https://github.com/patrick-kidger/equinox/issues/1076#issuecomment-3193477987).

michael-0brien:
I've opened this question in JAX discussions... it seems appropriate to ask. Perhaps they do not have the behavior of wsc outside jitted regions clearly defined or tested. Or, maybe it really is a "hint" outside jitted regions.

patrick-kidger:
From #1076:

> “For most operations these rules are simple and obvious because there’s only one reasonable choice. But for some operations it’s unclear how to shard the result. In that case we ask the programmer to provide an out_sharding argument explicitly and we throw a (trace-time) error otherwise.”

I don't think this is referring to `jax.jit`; this is referring to the new API (that we're not using here and are happy to keep not using here).

On the topic of adding `eqx.filter_reshard` etc: I'm not sure. As this discussion has highlighted it's actually a pretty tricky thing to maintain! I suspect a lot of the more advanced sharding APIs may be things that we do not offer a filtered equivalent for in Equinox; anyone with opinions there could easily implement it themselves with partition+combine anyway.

On the topic of what is committed... no idea! This doesn't seem clear, I agree.

michael-0brien:
It will be interesting to see what we learn… regardless, on the topic of how far should the equinox sharding support go, it seems to me most use cases as per the docs are implemented with `jax.device_put` on the inputs! 

I went ahead and got some information on the behavior of with_sharding_constraint in eager mode: https://github.com/jax-ml/jax/discussions/31127#discussioncomment-14143589. It is implemented as something like `jax.jit(lambda x: x, out_shardings=sharding)`. To me this doesn’t really explain the behavior in the example I showed above, but I think it is an argument for using `jax.device_put`: the two functions are not the same, and there could be unexpected behavior using JAX not as they advertise (here with the example being whether or not arrays are committed, see the discussion https://github.com/jax-ml/jax/discussions/30869 if this is not clear or the relevant place in the JAX docs https://docs.jax.dev/en/latest/faq.html#controlling-data-and-computation-placement-on-devices).

I've pushed a change that is in line with @patrick-kidger's plan in the discussion in #1076, but I adjust `eqx.filter_shard` to call `jax.lax.with_sharding_constraint` if the computation is being traced and `jax.device_put` otherwise. I think I’ve implemented this correctly but I have not tried this before, so if you have please take a look.

miltonllera:
I don't have a strong opinion on what should be done, but isn't this even less intuitive than before? You would now have one function calling two functions depending on context, one of which it is not even clear how it works internally. Seems like it would be much more intuitive to add a `filter_put` function that is clearly a filtered version of `device_put` (you could even add the device to the name) and make it explicit in the docs `filter_shard` is a filtered version of `wcs` (can't remember if that's the case as I only found out about this after it was mentioned in the thread). This is less likely to fool users in the case that something changes in the future on Jax's part, which seems likely.

michael-0brien:
The reason I've implemented this is 1) The equinox documentation does actually state that it is a function that combines the use of `jax.device_put` and `jax.lax.with_sharding_constraint`. I learned that it was calling wsc from reading the source code. So to me this puts it in line with its description! 2) It does so in a way that is now consistent with JAX; device_put is intended only for eager regions and wsc is intended only for jitted regions. I agree that there is an argument for two separate functions, but the current framework also has its advantages and I think this is the way to implement things in that framework. This is my first contribution so I'll defer to Patrick and the other equinox developers!

On “one of which isn’t clear how it works internally”: it is clear how `with_sharding_constraint` works inside jitted regions! The unclear part of the example I was testing earlier was only because I was calling the function outside jitted regions. 

I think the tests are passing... I'm getting this seg fault on any call to `eqx.filter_jit` in tests, maybe there's something weird with my install or I'm not set up to properly run tests.

michael-0brien:
Some clarification here: https://github.com/jax-ml/jax/discussions/31105. I'm interpreting the the eager behavior of `jax.lax.with_sharding_constraint` just as a choice designed to get some reasonable behavior under `jax.disable_jit`.

michael-0brien:
Hi, replying to #1076 here @miltonllera @lockwo

> > Just noting there are places where JAX docs do use in/out sharding, I'm not up to date on the 0.7.0 stuff, but e.g. https://docs.jax.dev/en/latest/notebooks/layout.html and something I've been looking at recently FFI https://docs.jax.dev/en/latest/ffi.html...
> 
> The issue with both of those examples is that the first is clearly experimental and the second one is from `0.4.23` which is several versions ago. Jax has some impressive features (and I really like the functional style) but the documentation is not the best maintained, which is a shame.

See https://github.com/jax-ml/jax/discussions/31105#discussioncomment-14154823; I am not totally sure but I think this means both APIs will be maintained.

I’ll summarize my conclusions here from the docs and various threads:

- Most use cases in the scope of `equinox` currently are covered by `jax.device_put` on the inputs.
- `jax.lax.with_sharding_constraint` is documented / discussed as for constraining intermediate values inside a jitted regions, while the primary use case in equinox will likely be using it for input/outputs. If this remains we should be mindful that this could cause confusion and perhaps should add something in the docs. 
- `in_shardings` or `out_shardings` are not soft-deprecated as per the thread linked above (this doesn’t explicitly confirm this but that is my conclusion for now). My interpretation remains that this will be a more commonly used API for constraining inputs/outputs, then users may invoke `jax.lax` for finer control. Through `jax.experimental.layout` there is currently functionality only available through `in_shardings` and `out_shardings` (i.e. Layout.AUTO), but we will have to see how this evolves to see how equinox should follow!

mjo22:
Hi, see my bug report in JAX here: https://github.com/jax-ml/jax/issues/31199#issuecomment-3208075423.

To resolve this it looks like `jax.lax.with_sharding_constraint` will behave just as an identity and will no longer make any device transfer at all.

@patrick-kidger if the current plan makes sense to you, we may be ready to merge.

michael-0brien:
As per #1084, just tested the tutorial and test_sharding.py on JAX 0.6.2 and everything is passing. @patrick-kidger if you think this plan is a good one and the `eqx.filter_shard` implementation doesn't have any sharp edges, I think things are ready for merge. Let me know if anything else is required.

FYI I've added some additional descriptions and links to the JAX docs in the tutorial for clarity. I noticed the tutorial was stating it would use 8 devices but did not force this in the JAX config, so we now force 8 CPU devices.

michael-0brien:
Okay, sorry for the spam but I've returned things to the original behavior of using `with_sharding_constraint` only.

I was misunderstanding some things about JAX and indeed learned there are sharp edges to what I was proposing branching on `device_put` and `with_sharding_constraint`: https://github.com/jax-ml/jax/issues/31237. I had thought it was best to stick to the patterns in the JAX docs and not use `with_sharding_constraint` if using eagerly, but indeed the JAX developers have accounted for this by making `jax.lax.with_sharding_constraint` flexible. There was a small bug in the eager behavior of `with_sharding_constraint` where arrays are not committed if wsc is just an identity, but now its behavior should match `device_put`: https://github.com/jax-ml/jax/issues/31199.

The original development of `eqx.filter_shard` was probably aware of all this already, so thank you for your patience!

miltonllera:
> Okay, sorry for the spam but I've returned things to the original behavior of using with_sharding_constraint only.

I actually think this has been quite illustrative. It is a bit of a shame that the Jax docs are a bit hard to follow at times. The framework already has a higher learning curve than Torch, which makes having a solid documentation all the more important.

For future readers of this thread, here is a filtered version of `device_put` in case you wish to use that instead of `eqx.filter_shard`:

``` python
import jax
import equinox as eqx
from jaxtyping import PyTree

def filter_put(model: PyTree, sharding: jax.sharding.NamedSharding):
    params, static = eqx.partition(model, eqx.is_array)
    params = jax.device_put(params, sharding)
    return eqx.combine(params, static)
```

tests/test_sharding.py
```diff
@@ -1,18 +1,51 @@
 import equinox as eqx
 import jax
+import jax.numpy as jnp
 import jax.random as jr
+import jax.sharding as jshard
 import jax.tree_util as jtu
-from jax.sharding import Mesh, PartitionSpec
 
+jax.config.update("jax_platform_name", "cpu")
+jax.config.update("jax_num_cpu_devices", 2)
```
patrick-kidger: I think this will potentially mess with the rest of our tests (in particular the pmap tests), since it's a global change. Better to use the existing single-CPU sharding I think?

equinox/_sharding.py
```diff
@@ -1,6 +1,7 @@
 from typing import Any
 
 import jax
+import jax.core
```
patrick-kidger: This seems spurious?

equinox/_jit.py
```diff
@@ -381,11 +381,11 @@ def f(x, y):  # both args traced if arrays, static if non-arrays
...
     deprecated_0_10(jitkwargs, "out")
     if any(
-        x in jitkwargs for x in ("static_argnums", "static_argnames", "donate_argnums")
+        x in jitkwargs for x in ("static_argnums", "static_argnames", "donate_argnums", "in_shardings", "out_shardings")
     ):
         raise ValueError(
-            "`jitkwargs` cannot contain 'static_argnums', 'static_argnames' or "
-            "'donate_argnums'"
+            "`jitkwargs` cannot contain 'static_argnums', 'static_argnames', "
+            "'donate_argnums', 'in_shardings', or 'out_shardings'."
```
patrick-kidger, line 388: Nice!

docs/examples/parallelism.ipynb
patrick-kidger, line 1: The changes here LGTM!

patrick-kidger (commented):
Okay, nitty comments only, I like the look of this! Thank you for battling through the JAX details to figure out what was going on here.

tests/test_sharding.py
michael-0brien: Ah that makes sense. So the first test will fail due to this: https://github.com/jax-ml/jax/issues/31199. I could see if setting up a mesh with 1 device would pass.

michael-0brien:
Ok addressed comments, however now we do not pass tests due to the bug (should be fixed for JAX 0.7.2) https://github.com/jax-ml/jax/issues/31199. The TLDR of this is that the eager behavior of `jax.lax.with_sharding_constraint` will not shard or commit arrays if the specified sharding doesn't actually require changing anything about the array.

There are edge cases where this would create unexpected behavior diverging from `jax.device_put`, but it shouldn't impact programs for the most part. I can address this further somehow if desired, or we can merge.

tests/test_sharding.py
patrick-kidger: Yup, a mesh with 1 device sounds good to me.

patrick-kidger:
Okay! I'm happy to wait until JAX 0.7.2 to land this change. Both JAX 0.7.0 and JAX 0.7.1 have anyway been unusable for the Equinox ecosystem, so we're pretty much waiting for the next release regardless :)

miltonllera:
Out of curiosity, what issues have you encountered? Because for me it works perfectly...

patrick-kidger:
JAX 0.7.0 causes a segfault on interpreter shutdown, due to a double-free error. (https://github.com/jax-ml/jax/issues/30517)
JAX 0.7.1 causes some `O(n)` programs to run in `O(n^2)` time. (https://github.com/jax-ml/jax/issues/31284)

The former affects all usage of `equinox.filter_jit`, the latter affects approximately all usage of Diffrax.

miltonllera:
Interesting. Is it an issue on TPU? I have had no issue whatsoever with `equinox.filter_jit` on either JAX 0.7.0 or 0.7.1.

patrick-kidger:
Mysterious! I got the segfault on CPU.

Anyway if it's working for you, great! :D

patrick-kidger:
Okay, JAX 0.7.2 is out and so this is now merged! As a practical matter I've done so via #1103, so as to squash the commits together (whilst retaining your authorship).

Thank you for all your effort investigating this, which I know took some work. :)

michael-0brien:
Great! Thanks. FYI in test_sharding.py the assert statements I wrote are still commented out—I had done this before we decided we’d wait for 0.7.2. They should all pass now.

Issue #1081, opened 2025-08-20
`eqxi.filter_primitive_bind` and JAX 0.7.1 seems to throw error


mjo22:
I had run a `lineax.linear_solve` and got the following error in `equinox.internal.filter_primitive_bind` as of JAX 0.7.1:

```
     def filter_primitive_bind(prim: jax.extend.core.Primitive, *args) -> PyTree:
        """Calls a primitive that has had its rules defined using the filter
        functions above.
        """
        assert prim.multiple_results
        # If `args` contains a Jaxpr or ClosedJaxpr in its leaves, then it ends up as a
        # member of the `static` tuple. This is important to ensure that jaxpr-rewriting
        # passes are able to find it.
        # (E.g. if `eqx.filter_closure_convert(...)` is an argument and we apply
        # `jax.core.jaxprs_in_params`.)
        flat, treedef = jtu.tree_flatten(args)
        dynamic = [x for x in flat if is_array(x)]
        static = tuple(_missing_dynamic if is_array(x) else x for x in flat)
        flatten = Flatten()
        flat_out = prim.bind(*dynamic, treedef=treedef, static=static, flatten=flatten)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       TypeError: cannot create weak reference to 'Flatten' object

```

The JAX [changelog](https://github.com/jax-ml/jax/releases/tag/jax-v0.7.1) for 0.7.1 says:

"Several dozen internal APIs have been deprecated from `jax.interpreters.ad`, `jax.interpreters.batching`, and `jax.interpreters.partial_eval`; they are used rarely if ever outside JAX itself, and most are deprecated without any public replacement."

johannahaffner:
Interesting, this just came up in the context of https://github.com/patrick-kidger/lineax/issues/169 as well.

patrick-kidger:
Away from my laptop ATM, but does it suffice to add `__weakref__` to the the `__slots__`?

johannahaffner:
Yes, looks like it! I had been trying that. PR with the fix up in a second! Not 100 % sure if it fixes the other Lineax issue yet.

patrick-kidger:
Closing as fixed in #1082, thank you @johannahaffner !

PR #1082, opened 2025-08-20
Add support for creating weak references to Flatten()

johannahaffner:
This fixes the error reported in https://github.com/patrick-kidger/equinox/issues/1081. 

This seems to fix https://github.com/patrick-kidger/lineax/issues/169 as well: 

- With the Equinox fix, using `eqx.debug.assert_max_traces`: no error thrown for repeated calls to `lx.linear_solve`
- Without the Equinox fix, `assert_max_traces` reports recompilation when using JAX 0.7.0 for `lx.linear_solve`

### DIFF
--- a/equinox/internal/_primitive.py
+++ b/equinox/internal/_primitive.py
@@ -98,7 +98,7 @@ def _make_spec(x, y):
 
 
 class Flatten:
-    __slots__ = ("treedef_out", "static_out")
+    __slots__ = ("treedef_out", "static_out", "__weakref__")
 
     def called(self):
         return hasattr(self, "treedef_out")


patrick-kidger:
Tests are failing because `jax.interpreters.batching.NotMapped` is deprecated. Digging into this, replacing this with `None` seems to be correct. That's simple, at least.

Concerningly, however, I am finding (after that fix) that the following three tests fail on JAX 0.7.1:
```
tests/test_while_loop.py::test_speed_buffer_while[False] - assert 15.108638292062096 < 1
tests/test_while_loop.py::test_speed_buffer_while[True] - assert 15.343655332922935 < 1
tests/test_while_loop.py::test_speed_grad_checkpointed_while - assert 2.089675624971278 < 0.5
```
This indicates that something has shifted in JAX/XLA that completely nukes the performance of our while loop; likely that our in-place updates are now happening out-of-place. This will probably render Diffrax unusable.

I'm going to be honest, it's pretty unlikely I can find the time any time soon to dig into whatever is going on here, but bisecting through jaxlib nightlies would be a reasonable place to determine what has gone wrong. I'm going to tag @lockwo on this one in case you have any appetite/interest on this one. (Or at the very least a heads-up that this is happening.)

johannahaffner:
> completely nukes the performance of our while loop; likely that our in-place updates are now happening out-of-place. This will probably render Diffrax unusable.

This cannot be allowed to happen :) 

> bisecting through jaxlib nightlies

This part at least is also easy :) using Tuesday's 20250819 nightly JAX release, we get this specific test failure starting form jaxlib nightly 20250814. Going down to jaxlib nightly 20250813 allows the tests for this function to pass again.

Depending on exactly when their nightly releases are built and when the commits were made, this means we are looking at jaxlib commits from the 13th/14th of August. Judging purely from its title and description, this one stands out to me (the others are also mostly labeled GPU/TPU):

- https://github.com/jax-ml/jax/commit/da357bbcf33765fd09398217b6ee7c01ae076555

lockwo:
>https://github.com/jax-ml/jax/commit/da357bbcf33765fd09398217b6ee7c01ae076555

I haven't dug into jaxlib that much, is the socket stuff done here like some way JAX does interface?

Maybe also something with a linearize (since buffer has it and grad test is as well?) https://github.com/jax-ml/jax/commit/45a3fe62d6b567014e4fd742d1e38f02e16baefa 

I will see about making a pure JAX snippet and see if the jax devs are able to point to anything (in JAX or XLA). 

>This indicates that something has shifted in JAX/XLA that completely nukes the performance of our while loop; likely that our in-place updates are now happening out-of-place. This will probably render Diffrax unusable.

Thankfully we already had an intern rewrite our codebase to avoid while loops and just manually unroll them in code (for some reason he hasn't halted coding yet).

patrick-kidger:
> I will see about making a pure JAX snippet and see if the jax devs are able to point to anything (in JAX or XLA).

Okay, awesome!

Given the tests I have a strong suspicion that it will be that a pattern of the form `loop(some combination of reading from and writing from buffers)` has become `O(n^2)` instead of `O(n)`.

When we originally fixed this I wrote this benchmark a couple if years ago, that may provide a starting point:

https://github.com/jax-ml/jax/blob/79c2d48b5771f8bb005880c2856905346d8b3c86/benchmarks/api_benchmark.py#L798

johannahaffner:
> > [jax-ml/jax@da357bb](https://github.com/jax-ml/jax/commit/da357bbcf33765fd09398217b6ee7c01ae076555)
> 
> I haven't dug into jaxlib that much, is the socket stuff done here like some way JAX does interface?
> 

I haven't dug into it either, so I can't really say. From what I gather this commit is about transferring data into existing arrays (?) by creating some kind of buffer of future values (?), with the possible motivation that this avoids creation of new arrays (?). The PR does not have a description beyond the title, and received no review or comments.

But also - I reverted this commit and built jaxlib from source, and I can report that this did not fix the issue, so it is probably something else.

> Maybe also something with a linearize (since buffer has it and grad test is as well?) [jax-ml/jax@45a3fe6](https://github.com/jax-ml/jax/commit/45a3fe62d6b567014e4fd742d1e38f02e16baefa)
> 
> I will see about making a pure JAX snippet and see if the jax devs are able to point to anything (in JAX or XLA).

That would be great! I'm going to run the hourly investigation suggested [here](https://docs.jax.dev/en/latest/investigating_a_regression.html).

Issue #1095, opened 2025-09-16
TypeError with JAX 0.7.2: non-hashable parameter in noinline primitive


johannahaffner:
JAX 0.7.2 just dropped, so we can move ahead and do a release! ...not so fast, I'm getting one new error:

```
E       TypeError: As of JAX v0.7, parameters to jaxpr equations must have __hash__ and __eq__ methods. 
In a call to primitive noinline, the value of parameter static was not hashable: 
(LiteralArray(7, dtype=int32), 
<function noinline.<locals>.abstract_fn at 0x33b12a160>, 
<function _impl_transform at 0x1091eca40>, 
<function _jvp_transform at 0x1091ed080>, 
ShapedArray(float32[1]), 0, 0, 
<object object at 0x108068d10>, 
<object object at 0x108068d10>, 
<object object at 0x108068d10>, 
<object object at 0x108068d10>, 
<object object at 0x108068d10>, 
<object object at 0x108068d10>, 
<jax._src.custom_derivatives.custom_jvp object at 0x10898be30>, 
<function <lambda> at 0x1093c22a0>, 
<object object at 0x108068d10>, 
<object object at 0x108068d10>)
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

equinox/internal/_primitive.py:271: TypeError
```

I'm going to investigate this, looks like this function needs an update similar to https://github.com/patrick-kidger/equinox/pull/1039.

johannahaffner:
This is purely due to the `LiteralArray`, which is a new private JAX type that is introduced in 0.7.2: https://github.com/jax-ml/jax/releases/tag/jax-v0.7.2. It is not hashable, of course. It seems that the hashability requirements for jaxpr parameters is only now fully "on" (or was only partially introduced in earlier 0.7x releases), since we always supported partitioning arrays into the static component of a pytree and did not get errata so far.

Wrapping it naively in `eqxi.Static` breaks other code, due to the requirement to call `Static.value` later on. Since any change made here would percolate through the library, I wanted to raise this as a design question. How should we treat arrays that are part of the static component of some pytree? It is clear to me that we want to enable things like custom filter specs, so we will need to wrap these in something or convert them, and then unwrap or convert back as appropriate.

lockwo:
[...]

PR #1098, opened 2025-09-17
Remove a ruff rule that no longer exists from the list of rules we ignore

johannahaffner:
We previously had this rule deactivated. In the current version of ruff, this no longer exists and only raises:

```bash
warning: The following rules have been removed and ignoring them has no effect:
    - UP038
```

I figured that we can remove it on our end as well. 

Or should we keep it in case some contributors use older versions of ruff?

### DIFF
--- a/equinox/nn/_conv.py
+++ b/equinox/nn/_conv.py
@@ -2,7 +2,6 @@ import itertools as it
 import math
 from collections.abc import Callable, Sequence
 from typing import cast, TypeVar
-from typing import TypeVar, cast
 
 import jax.lax as lax
 import jax.numpy as jnp


patrick-kidger:
Happy to remove it! In particular as the version of ruff that is used should be pinned for all contributors via `.pre-commit-config.yaml`.

On which note, I note that our CI is failing because it is indeed using an older version of ruff. 😄

johannahaffner:
Haha :D I have updated the versions for ruff and pyright, fixed what ruff complained about (white-space related formatting) and ignored what pyright complained about (overlapping overloads + some union types). 

Now we still get two actual test failures - they are what I started to investigate in https://github.com/patrick-kidger/equinox/issues/1095.

johannahaffner:
This branch is now ready to be merged. The two remaining errata in the tests will be fixed by https://github.com/patrick-kidger/equinox/pull/1099, which should be merged after.

patrick-kidger:
Done!

Issue #1097, opened 2025-09-17
Error in All of Equinox doc


manuelhaussmann:
The filtering example  (https://docs.kidger.site/equinox/all-of-equinox/#2-filtering) currently breaks for me when running Option 1, i.e.,
```python
@ft.partial(jax.jit, static_argnums=1)  # `static` must be a PyTree of non-arrays.
@jax.grad  # differentiates with respect to `params`, as it is the first argument
def loss(params, static, x, y):
    model = eqx.combine(params, static)
    pred_y = jax.vmap(model)(x)
    return jax.numpy.mean((y - pred_y) ** 2)

params, static = eqx.partition(model, eqx.is_array)
loss(params, static, x, y)
```
returns a Type error (`TypeError: unhashable type: 'list'`)

Using option two and equinox' `eqx.filter_{fix,grad}` works perfectly.

Equinox version 0.13.0
Jax both 0.7.2 and <0.7

johannahaffner:
I think this is another variant of https://github.com/patrick-kidger/equinox/issues/1095, I'll take a look at it.

patrick-kidger:
Ah, this is actually the same issue as #1089. We fixed a minor bug in how Equinox modules are hashed.

I've just updated the docs to switch this example from using a list to using a tuple.

PR #1099, opened 2025-09-18
Adding `LiteralArray` to the types recognised by `is_array`

johannahaffner:
This is a new, private JAX type that can break instance checks, although it does otherwise duck type as a numpy array. I'm adding it to the union of types for which `is_array` returns `True`, since their suggested workaround of using `np.asarray` is not useful in our case. 

This will fix https://github.com/patrick-kidger/equinox/issues/1095 until the next release of JAX, since `LiteralArray` has already been renamed to `TypedNdArray`. 

Builds on top of https://github.com/patrick-kidger/equinox/pull/1098 to pass the CI.


patrick-kidger:
I note that `isinstance(..., jax.Array)` is already overriden to pass, e.g. for both tracers and for underlying `ArrayImpl` (=`type(jnp.array(...))` outside of JIT)`. Perhaps this would best be upstreamed into JAX there?

If they decide not to then this LGTM, though!

johannahaffner:
> If they decide not to then this LGTM, though!

Let's merge this - then we are compatible with 0.7.2 and can do a release with all the early 0.7.0/0.7.1 shenanigans behind us.

What exactly do you mean by upstreaming? As in, where is the override defined? I'm happy to open a PR upstreaming this, and to retire our workaround here asap.

patrick-kidger:
> Let's merge this - then we are compatible with 0.7.2 and can do a release with all the early 0.7.0/0.7.1 shenanigans behind us.

Makes sense to me!

We have quite a few different fixes in flight (ruff, pyright, JAX 0.7.2 here, Python 3.14). Ideally we'd keep CI green as we merge each of these in turn. What order do you want to merge stuff in? (pyright + JAX 0.7.2 first I think?)

(Keeping it green really isn't essential if it's easier for you though, happy with this choice as your call.)

> What exactly do you mean by upstreaming? As in, where is the override defined? I'm happy to open a PR upstreaming this, and to retire our workaround here asap.

I'm not sure where JAX have this defined. I think it might be in the C++ class, judging from this line: https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jax/_src/basearray.py#L174

johannahaffner:
I think the easiest would be if you merged your Python 3.14 compatibility fix first, then I rebase the ruff + pyright thing, and then we add this one last. 

The reason is that your compatibility fix includes a better way to deal with the pyright errors that I'm just ignoring, so I rebase on that one first.

(Edit: this will result in two red CIs, followed by this final green one.)

johannahaffner:
> I'm not sure where JAX have this defined. I think it might be in the C++ class, judging from this line: https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jax/_src/basearray.py#L174

Haha this is a fun little trail: [1](https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jax/_src/basearray.py#L174), [2](https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jax/_src/basearray.py#L23), [3](https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jax/_src/lib/__init__.py#L89), [4](https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jaxlib/xla_client.py#L324), [5](https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jaxlib/xla_client.py#L28), [6](https://github.com/jax-ml/jax/blob/4c5411b8878ef0ea53222e1a2f2194888560d546/jaxlib/_jax/__init__.pyi#L621). Unfortunately we leave the land of well-defined types with an `Any`.

patrick-kidger:
> I think the easiest would be if you merged your Python 3.14 compatibility fix first, then I rebase the ruff + pyright thing, and then we add this one last.

Sg! Merged the 3.14 one then :)

johannahaffner:
This PR is ready to be merged and should get in after https://github.com/patrick-kidger/equinox/pull/1098, at which point its diff to main will be reduced to about twenty lines in one single file that only pertain to compatibility with the new private JAX types `LiteralArray`, which now exists under the moniker `TypedNdArray` on the current `main` branch of JAX. The change is backward and non-breaking. As long as `TypedNdArray` does not change its name again, it is also forward-compatible with the next JAX release.

Re: upstreaming: I think this fix will need to stay in for compatibility with 0.7.2, even if it becomes unnecessary again for future versions of JAX.

johannahaffner:
This change does not resolve the unrelated issue https://github.com/patrick-kidger/equinox/issues/1097, so that one will still need to be dealt with before we're ready for a release.

patrick-kidger:
Okay!
Heads-up that unfortunately GitHub does not really support 'stacked PRs' (a long-standing issue), so you'll need to manually rebase this branch on top of `main` first.

johannahaffner:
Done.

patrick-kidger:
And merged!

Issue #1096, opened 2025-09-16
Why can (abstract) attributes without default values be declared after attributes that do have default values?


johannahaffner:
Looking into https://github.com/patrick-kidger/optimistix/issues/168, I found one of the culprits [here](https://github.com/patrick-kidger/lineax/blob/6bf1a0e17c6352b96bdf52e72772fac63eb641df/lineax/_solver/cg.py#L53). Lineax can be imported + tested if I move the declaration of the private abstract attributes up, so that it occurs before the `norm` is declared with its default values.

So far, so good. But what I don't understand is why this worked in the first place, given that a minimal example 

```python
import equinox as eqx


class AbstractSomething(eqx.Module):
    a: int = 3
    b: eqx.AbstractVar[bool]

class ConcreteSomething(AbstractSomething):
    b: False

a = ConcreteSomething() 
```

raises a `TypeError`, as expected. Optimistix has the same issue in one of the root finders.

patrick-kidger:
Your MWE uses `AbstractVar` rather than `AbstractClassVar`, which is the difference to the Optimistix issue. :)

(I'm going to close this issue as it's now been traced upstream to the Python 3.14 prerelease.)

PR #1100, opened 2025-09-18
Python 3.14 compatibility

patrick-kidger:
Python 3.14 has deprecated the use of `cls.__dict__["__annotations__"]` in favour of either `inspect.get_annotations` (Python 3.10+) or `annotationlib.get_annotations` (Python 3.14+).

This is documented in [this ruff rule](https://docs.astral.sh/ruff/rules/access-annotations-from-class-dict/) in [PEP 749](https://peps.python.org/pep-0749/#new-annotationlib-module). See also this CPython issue: https://github.com/python/cpython/issues/139140

IIUC the above CPython issue (+PEP 749) correctly then the `annotationlib` function is the 'recommended' option, but as a practical matter they'll become the same function, and the `inspect` option is already available on older Pythons.

Fixes https://github.com/patrick-kidger/equinox/issues/1096 and https://github.com/patrick-kidger/optimistix/issues/168, hopefully.

### DIFF
--- a/equinox/_module/_better_abstract.py
+++ b/equinox/_module/_better_abstract.py
@@ -8,6 +8,7 @@ is the public API. But the two pieces aren't directly related under-the-hood.)
 
 import abc
 import dataclasses
+import inspect
 from typing import (
     Annotated,
     ClassVar,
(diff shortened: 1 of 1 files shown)


johannahaffner:
I'll rebase https://github.com/patrick-kidger/equinox/pull/1098 and https://github.com/patrick-kidger/equinox/pull/1099 on this once its in :)

PR #1102, opened 2025-09-19
Fixing representations of vmapped enumerations

LennartGevers:
I'm adressing the Issue patrick-kidger/diffrax#690 where printing `Solution.result` fails on a vmap'ed diffreqsolve output.

The problem is that when evaluation item._value.item(), _value is an array of dimension > 0 and thus has no python scalar representation which makes .item() fail.
The bug is solved by vectorizing the part where the _index_to_message dict is accessed, using a lambda function.

The result of @f0uriest mwe is then

```
diffrax._solution.RESULTS<['' '']>
```

### DIFF
--- a/tests/test_enum.py
+++ b/tests/test_enum.py
@@ -93,7 +93,7 @@ def test_repr():
 
     result = return_enum(jnp.arange(2))
 
-    assert str(result).endswith(".test_repr.<locals>.A<['foo' 'foo']>")
+    assert str(result).endswith("test_enum.test_repr.<locals>.A<['foo' 'foo']>")
 
 
 def test_inheritance_and_len():


patrick-kidger:
LGTM! I think it'd be worth adding a test as well to ensure that we don't accidentally break this in the future?

LennartGevers:
> LGTM! I think it'd be worth adding a test as well to ensure that we don't accidentally break this in the future?

Yes, sure, I'll follow up with a testcase

LennartGevers:
Have I placed the test case in the correct spot? @patrick-kidger

patrick-kidger:
LGTM -- merged!

Thank you for adding in this fix :)

PR #1105, opened 2025-09-19
Enable sharding tests

patrick-kidger:
C.f. https://github.com/patrick-kidger/equinox/pull/1078#issuecomment-3312780896 @mjo22

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ typing.TESTING = True  # pyright: ignore
 
 jax.config.update("jax_numpy_dtype_promotion", "strict")
 jax.config.update("jax_numpy_rank_promotion", "raise")
+jax.config.update("jax_num_cpu_devices", 2)
 
 
 @pytest.fixture
--- a/tests/test_pmap.py
+++ b/tests/test_pmap.py
@@ -11,7 +11,7 @@ import pytest
 from .helpers import tree_allclose as _shaped_allclose
 
 
-(cpu,) = jax.devices("cpu")
+(cpu, _) = jax.devices("cpu")
 filter_pmap: Any = ft.partial(eqx.filter_pmap, devices=[cpu])  # pyright: ignore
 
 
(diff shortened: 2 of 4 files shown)


patrick-kidger:
@mjo22 we appear to be getting failures :(
WDYT?

A little debugging locally suggests that maybe these still aren't being sharded in the way we expect.

michael-0brien:
Looking at things locally, sharding on the inputs outside JIT in this line does pass:

```
def test_sharding_no_inside_jit():
    mlp = eqx.nn.MLP(2, 2, 2, 2, key=jr.PRNGKey(0))

    num_devices = 1
    mesh = jax.make_mesh((num_devices,), ("x",))
    sharding = jshard.NamedSharding(mesh, jshard.PartitionSpec("x"))
    sharded_mlp = eqx.filter_shard(mlp, sharding)
    assert _is_committed(eqx.filter(sharded_mlp, eqx.is_array))
    assert _is_sharded(eqx.filter(sharded_mlp, eqx.is_array), sharding)
```

This confirms that the bug fixed for JAX 0.7.2 in `with_sharding_constraint` outside JIT regions was indeed incorporated and fixed.

However, the assertion on the outputs in both tests fail. Instead of an array's sharding `x.sharding` being the defined `NamedSharding`, the sharding on the outputs are the `SingleDeviceSharding` object. These two shardings are really the same sharding in practice, so I don't fully understand if this is an error. Does that make sense?

michael-0brien:
Okay I was being silly. The issue is just in comparing shardings that belong to arrays vs the sharding we instantiated. In other words, this fails:

```python
def _is_sharded(pytree, sharding):
    return jax.tree.reduce(
        lambda x, y: x and y,
        jax.tree.map(lambda x: x.sharding == sharding, pytree),
    )
```

due to the comparison `x.sharding == sharding`. We can just change this function to

```python
def _is_sharded(pytree):
    return jax.tree.reduce(
        lambda x, y: x and y,
        jax.tree.map(lambda x: isinstance(x.sharding, jshard.NamedSharding), pytree),
    )
```

patrick-kidger:
So I think we would expect the the sharding to be preserved entirely, given that this is an elementwise operation.

Fortunately, it seems that setting the number of CPUs to 2 fixes the issue. This seems like a minor bug (?) in JAX, where the sharding swaps out for something equivalent in the single-device case.

(Specifically, it swaps from `NamedSharding(mesh=Mesh('x': 1, axis_types=(Auto,)), spec=PartitionSpec('x',), memory_kind=device)` to `NamedSharding(mesh=Mesh('x': 1, axis_types=(Auto,)), spec=PartitionSpec(), memory_kind=device)`, note change in `PartitionSpec`.)

michael-0brien:
Oh very interesting, good catch. I had thought making a comparison via `==` asked if python objects were the same in memory, but refreshing my memory this is the behavior of `is`. 

I’ll open an issue in JAX.

PR #1108, opened 2025-09-23
Shameless plug: add typinox to awesome-list

EtaoinWu:
So I made this library (Currently at its very early stages!) that allows you to write this:

```python
class AffineMap(TypedModule):
    k: Float[Array, "n m"]
    b: Float[Array, "n"]

    def __call__(self: Self, x: Float[Array, "m"]) -> Float[Array, "n"]:
        return jnp.dot(self.k, x) + self.b

f = AffineMap(k=jnp.ones((3, 2), dtype=float), b=jnp.zeros(3))
f(jnp.array([1., 2.]))
```

The initializer and member methods will automatically type-check at runtime.

### DIFF
--- a/docs/awesome-list.md
+++ b/docs/awesome-list.md
@@ -88,10 +88,13 @@ Normalizing flows and probabilistic machine learning.
 Converts PyTorch `state_dict`s into Equinox pytrees.
 
 **boltz-binder-design**: [GitHub](https://github.com/escalante-bio/boltz-binder-design)   
-Protein design by multi-objective hallucination (~ColabDesign in Equinox). 
+Protein design by multi-objective hallucination (~ColabDesign in Equinox).
 
 **ksim**: [GitHub](https://github.com/kscalelabs/ksim)  
 MJX robotics simulation library.
 
 **Mutax**: [GitHub](https://github.com/gerlero/mutax)  
 Differential evolution–based global optimization (with support for parallel evaluation).
+
+**Typinox**: [GitHub](https://github.com/EtaoinWu/typinox) [Docs](https://typinox.readthedocs.io/latest/)  
+Enhanced `beartype` run-time typechecking for `jaxtyping` annotated `equinox` Modules.


docs/awesome-list.md
```diff
@@ -88,10 +88,13 @@ Normalizing flows and probabilistic machine learning.
...
-Protein design by multi-objective hallucination (~ColabDesign in Equinox). 
+Protein design by multi-objective hallucination (~ColabDesign in Equinox).
 
 **ksim**: [GitHub](https://github.com/kscalelabs/ksim)  
 MJX robotics simulation library.
 
 **Mutax**: [GitHub](https://github.com/gerlero/mutax)  
 Differential evolution–based global optimization (with support for parallel evaluation).
+
+**Typinox**: [GitHub](https://github.com/EtaoinWu/typinox) [Docs](https://typinox.readthedocs.io/latest/)
```
patrick-kidger: Heads-up that this line needs to finish with a double space. This means markdown will treat it as a line break.

patrick-kidger:
Nice! One comment aside, LGTM. :)

IIUC the point is that `n` and `m` are now matched across both the init-time attributes and the call-time arguments?

Would this also work for
```python
class Foo(TypedModule):
    x: Float[...]

def bar(foo: Foo, baz: Float[...]): ...
```
? (FWIW in this case, I would argue that it is undesirable for this axis-matching to occur, as there may be a lot of code in between `Foo` and `bar` and this would be some spooky action at a distance if undesired / difficult to keep in sync if desired.)

EtaoinWu:
> FWIW in this case, I would argue that it is undesirable for this axis-matching to occur, as there may be a lot of code in between `Foo` and `bar` and this would be some spooky action at a distance if undesired / difficult to keep in sync if desired.

A quick test later:

```python
class Foo(TypedModule):
    x: Float[Array, "n m"]

    def equal(self, y: Float[Array, "n m"]) -> Bool[Scalar, ""]:
        return jnp.array_equal(self.x, y)

    def equal_val(self, y: Float[Array, "n m"]) -> Bool[Scalar, ""]:
        self._validate()   # for methods, this is opt-in
        return jnp.array_equal(self.x, y)

@jaxtyped(typechecker=beartype)
def foo_equal(foo: Foo, y: Float[Array, "n m"]) -> Bool[Scalar, ""]:
    return jnp.array_equal(foo.x, y)

def test_module_constraint_scope():
    foo = Foo(x=jnp.zeros((3, 4)))
    assert foo_equal(foo, jnp.zeros((3, 4)))
    assert foo.equal(jnp.zeros((3, 4)))
    assert foo.equal_val(jnp.zeros((3, 4)))

    assert not foo_equal(foo, jnp.ones((3, 4)))
    assert not foo.equal(jnp.ones((3, 4)))
    assert not foo.equal_val(jnp.ones((3, 4)))

    assert not foo_equal(foo, jnp.zeros((3, 5)))
    assert not foo.equal(jnp.zeros((3, 5)))
    with pytest.raises(TypeError, match=r"Float\[Array, 'n m'\]"):
        foo.equal_val(jnp.zeros((3, 5)))
```

docs/awesome-list.md
EtaoinWu: Added!

patrick-kidger:
Merged! :)

PR #1112, opened 2025-10-03
fix: include TypedNdArray in array-like types

nstarman:
The true-evaluating arguments to `is_array_like` should be a superset of `is_array`.

@patrick-kidger I think this should fix https://github.com/patrick-kidger/quax/pull/74


patrick-kidger:
LGTM!

PR #1116, opened 2025-10-11
Small Module cleanups

nstarman:
misc things I noticed when looking at Module recently.


patrick-kidger:
Ty! These all seem uncontroversial improvements to me.

PR #1119, opened 2025-10-12
perf: dynamic (un)flattening code generation

nstarman:
I haven't implemented the full logic, like the use of `_FLATTEN_SENTINEL`.

### DIFF
--- a/equinox/_module/_flatten.py
+++ b/equinox/_module/_flatten.py
@@ -1,0 +1,238 @@
+"""Utilities for generating optimized flatten/unflatten functions for Module."""
+
+import dataclasses
+import textwrap
+from enum import Enum
+from typing import Any, Final, Literal, TypeVar
+
+import jax.tree_util as jtu
+
+
+# Names for generated functions
+_FLAT_FUNC_NAME = "<generated_flatten_{0}>"
+_FLAT_KEYS_NAME = "<generated_flatten_with_keys_{0}>"
+_UNFLAT_NAME = "<generated_unflatten_{0}>"
+
+# Fields of Module wrappers to always include
+WRAPPER_FIELD_NAMES: Final = (
+    "__module__",
+    "__name__",
+    "__qualname__",
+    "__doc__",
+    "__annotations__",
+)
+
+# Indentation for generated code
+_INDENT: Final = " " * 4
+
+
+# Sentinel values for flattening/unflattening
+class _Sentinel(Enum):
+    """Sentinel values for flattening/unflattening."""
+
+    MISSING = "MISSING"
+
+
+MISSING = _Sentinel.MISSING
+
+# Code template for flattening the wrapper fields
+_GETTER = "get({k!r},MISSING)"  # get = obj.__dict__.get
+_FLAT_WRAPPERS = ", ".join([_GETTER.format(k=k) for k in WRAPPER_FIELD_NAMES])
+
+# Code template for flatten function
+_FLAT_CODE = f'''
+def {{func_name}}(obj: module_cls) -> {{return_annotation}}:
+    """Generated {{func_name}} function for {{qualname}}.
+
+    Dynamic fields: {{dynamic_fs}}
+    Static fields: {{static_fs}}
+    Wrapper fields: {WRAPPER_FIELD_NAMES}
+    """
+    get = obj.__dict__.get
+    return (
+        {{dynamic_vals}},
+        (
+            MISSING if len(obj.__dict__) == {{num_fields}} else ({_FLAT_WRAPPERS}),
+            {{static_vals}}
+        )
+    )
+'''
+
+# Code template for setting wrapper fields during unflattening
+_SET_WRAPPERS: Final = "\n".join(
+    f"self.__dict__[{name!r}] = waux[{i}]" for i, name in enumerate(WRAPPER_FIELD_NAMES)
+)
+
+# Code template for setting dynamic fields during unflattening
+_SET_DYNAMIC = """
+if data[{i}] is not MISSING:
+    object.__setattr__(self, {name!r}, data[{i}])
+"""[1:-1]  # (trim leading and trailing newlines)
+
+# Code template for setting static fields during unflattening
+_SET_STATIC = """
+if aux[{i}] is not MISSING:
+    object.__setattr__(self, {name!r}, aux[{i}])
+"""[1:-1]  # (trim leading and trailing newlines)
+
+# Code template for unflatten function
+_UNFLAT_FUNC = f'''
+def unflatten(
+    module_cls: type[T],
+    aux: {{aux_type}},
+    data: {{dynamic_type}},
+) -> T:
+    """Generated unflatten function for {{qualname}}.
+
+    Dynamic fields: {{dynamic_fs}}
+    Static fields: {{static_fs}}
+    Wrapper fields: {WRAPPER_FIELD_NAMES}
+    """
+    self = object.__new__(module_cls)
+    # Set fields directly by index
+{{setters_dynamic}}
+    if aux[0] is not MISSING:
+        waux = aux[0]
+{textwrap.indent(_SET_WRAPPERS, _INDENT * 2)}
+{{setters_static}}
+    return self
+'''
+
+_NS_BASE = {
+    "object": object,
+    "Any": Any,
+    "tuple": tuple,
+    "Literal": Literal,
+    "MISSING": MISSING,
+}
+
+
+def _make_tuple_type(count: int, elt: str, /) -> str:
+    """Generate a tuple type annotation string for a given count of elements."""
+    return "tuple[()]" if count == 0 else f"tuple[{', '.join([elt] * count)}]"
+
+
+def generate_flatten_functions(cls: type, fields: tuple[dataclasses.Field[Any], ...]):
+    """Generate optimized flatten/unflatten functions for a specific field config."""
+    # Separate dynamic and static fields
+    _dynamic_fs, _static_fs = [], []
+    for f in fields:
+        if f.metadata.get("static", False):
+            _static_fs.append(f.name)
+        else:
+            _dynamic_fs.append(f.name)
+    dynamic_fs, static_fs = tuple(_dynamic_fs), tuple(_static_fs)
+    n_dynamic_fs = len(dynamic_fs)
+    n_static_fs = len(static_fs)
+
+    # Extract the generated functions from respective namespaces to
+    # set the proper module reference.
+    module_name = getattr(cls, "__module__", cls.__name__)
+
+    # -------------------------------------------
+    # Generate flatten function
+
+    # Directly access dynamic fields by name
+    if n_dynamic_fs == 0:
+        dynamic_vals = "()"
+    else:
+        dynamic_exprs = [_GETTER.format(k=k) for k in dynamic_fs]
+        dynamic_vals = f"({', '.join(dynamic_exprs)},)"
+
+    # For static fields, we need to store their values in aux data
+    static_exprs = [_GETTER.format(k=k) for k in static_fs]
+    static_vals = f"{', '.join(static_exprs)}"
+
+    # Build return type annotation
+    dynamic_type = _make_tuple_type(n_dynamic_fs, "Any")
+    wrapper_type = "Literal[MISSING]|tuple[Any, ...]"
+    static_type = f"tuple[{wrapper_type}, {', '.join(['Any'] * n_static_fs)}]"
+    clsname = cls.__qualname__
+
+    # Generate flatten function code
+    flat_code = _FLAT_CODE.format(
+        func_name="flatten",
+        return_annotation=f"tuple[{dynamic_type}, {static_type}]",
+        qualname=clsname,
+        dynamic_fs=dynamic_fs,
+        static_fs=static_fs,
+        num_fields=n_dynamic_fs + n_static_fs,
+        dynamic_vals=dynamic_vals,
+        static_vals=static_vals,
+    )
+
+    # make flatten func
+    flat_ns = _NS_BASE | {"module_cls": cls}
+    exec(compile(flat_code, _FLAT_FUNC_NAME.format(clsname), "exec"), flat_ns)
+    flat_fn = flat_ns["flatten"]
+    flat_fn.__module__ = module_name
+    object.__setattr__(flat_fn, "__source__", flat_code)
+
+    # -------------------------------------------
+    # Generate flatten_with_keys function
+
+    # Generate flatten_with_keys values
+    if n_dynamic_fs == 0:
+        dynamic_key_vals = "()"
+    else:
+        key_exprs = [
+            f"(jtu.GetAttrKey({k!r}), {_GETTER.format(k=k)})" for k in dynamic_fs
+        ]
+        dynamic_key_vals = f"({', '.join(key_exprs)},)"
+
+    keys_dynamic_type = _make_tuple_type(n_dynamic_fs, "tuple[jtu.GetAttrKey, str]")
+
+    # Generate flatten_with_keys function code
+    flat_k_code = _FLAT_CODE.format(
+        func_name="flatten_with_keys",
+        return_annotation=f"tuple[{keys_dynamic_type}, {static_type}]",
+        qualname=clsname,
+        dynamic_fs=dynamic_fs,
+        static_fs=static_fs,
+        num_fields=n_dynamic_fs + n_static_fs,
+        dynamic_vals=dynamic_key_vals,
+        static_vals=static_vals,
+    )
+
+    # flatten with keys func
+    flat_k_ns = _NS_BASE | {"jtu": jtu, "module_cls": cls}
+    exec(compile(flat_k_code, _FLAT_KEYS_NAME.format(clsname), "exec"), flat_k_ns)
+    flat_k_fn = flat_k_ns["flatten_with_keys"]
+    flat_k_fn.__module__ = module_name
+    object.__setattr__(flat_k_fn, "__source__", flat_k_code)
+
+    # -------------------------------------------
+    # Generate unflatten function - directly set fields by index
+    # Extract types from flatten return type: tuple[dynamic_type, static_type]
+
+    # Set dynamic fields by index
+    unflat_dynamic = [
+        _SET_DYNAMIC.format(i=i, name=k) for i, k in enumerate(dynamic_fs)
+    ]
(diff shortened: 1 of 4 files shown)


nstarman:
@patrick-kidger This is ~80% of the way there. I think it's a tiny bit slower than the other PR but still gets equinox into the same ballpark as straight JAX.

The same test file, with the methods on the equinox module removed, should benchmark this PR. 

This PR is missing the `_flatten_sentinel` and wrapper params logic. I'm AFK for most of the rest of the week, so if you like this approach and want to polish this off, that'd be awesome :).

(A third alternative would be to mypyc Module and let `c` eliminate the overhead.)

equinox/_module/_module.py
```diff
@@ -40,90 +39,164 @@ def StrictConfig(
...
+    # Namespace for flatten functions (they need module_cls)
+    flatten_ns = {
+        "jtu": jtu,
+        "object": object,
+        "module_cls": cls,
+        "Any": Any,
+        "tuple": tuple,
+    }
+
+    # Namespace for unflatten function (takes module_cls as parameter)
```
nstarman:
```suggestion
    # Namespace for unflatten function
```

equinox/_module/_module.py
```diff
@@ -40,90 +39,164 @@ def StrictConfig(
...
+    Static fields: {static_fs}
+    """
+    return {key_tuple}, {static_aux}
+'''
+
+    # -------------------------------------------
+    # Generate unflatten function - directly set fields by index
+    # Extract types from flatten return type: tuple[dynamic_type, static_type]
+
+    unflatten_signature = (
```
nstarman: Can be rewritten as a single string

patrick-kidger:
I really like this approach. If there's a need for greater speed then codegen sounds like a great way to get things to be fast-by-default.

I definitely don't have time to pursue this myself but I'll happily merge this once it's ready.

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,225 @@
...
+    """Generate optimized flatten/unflatten functions for a specific field config."""
+    # Separate dynamic and static fields
+    _dynamic_fs, _static_fs = [], []
+    for f in fields:
+        if f.metadata.get("static", False):
+            _static_fs.append(f.name)
+        else:
+            _dynamic_fs.append(f.name)
+    dynamic_fs, static_fs = tuple(_dynamic_fs), tuple(_static_fs)
+    # aux_fs = WRAPPER_FIELD_NAMES + static_fs
```
nstarman: TODO: use

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,225 @@
...
+        dynamic_vals = "()"
+
+    # For static fields, we need to store their values in aux data
+    # static_exprs = [f"getattr(obj, {k!r}, MISSING)" for k in aux_fs]
+    # static_aux = f"({', '.join(static_exprs)},)"
+    if static_fs:
+        static_exprs = [f"getattr(obj, {k!r}, MISSING)" for k in static_fs]
+        static_aux = f"({', '.join(static_exprs)},)"
+    else:
+        static_aux = "()"
```
nstarman: TODO: swap the current static_aux for the commented one above

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,225 @@
...
+
+    unflatten_lines: list[str] = []
+    if dynamic_fs or static_fs:
+        unflatten_lines.append("# Set dynamic fields directly by index")
+    # Set dynamic fields directly by index
+    unflatten_lines.extend(
+        SET_DYNAMIC_BASE.format(i=i, name=k) for i, k in enumerate(dynamic_fs)
+    )
+    # # Set wrapper fields from aux_data
+    # unflatten_lines.append(SET_WRAPPER_LINES)
```
nstarman: TODO: use

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,225 @@
...
+    # Set static fields from aux_data
+    unflatten_lines.extend(
+        SET_AUX_BASE.format(i=i, name=k)
+        # for i, k in enumerate(static_fs, start=len(WRAPPER_FIELD_NAMES))
+        for i, k in enumerate(static_fs, start=0)
```
nstarman: TODO: use the incremented start.

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,225 @@
+"""Utilities for generating optimized flatten/unflatten functions for Module."""
+
+__all__ = (
+    "WRAPPER_FIELD_NAMES",
+    "generate_flatten_functions",
+)
```
patrick-kidger: Drive-by nit, I use leading underscores to denote private/public, and do not use `__all__`. Too annoying trying to keep it up-to-date!

patrick-kidger:
Awesome! I really like the look of this PR.

The performance hit on the extra magic attributes – I think the common case is that none of these are present on the instance, and that instead they are class attributes. Perhaps this could be exploited, so that e.g. flattening check their presence and sets a boolean flag, and then unflattening skips this codepath if it is not required?

nstarman:
The last commit needs a lot of cleanup, but if the `'__module__' in obj.__dict__` guard is fine for filtering the wrapper attributes, then this PR gets us much closer to JAX speeds.
I'm clocking ~9.3 µs for the Equinox module versus ~7.7 µs for the JAX pytree in the benchmark notebook. It used to be 12-15 µs. Without the wrapper stuff it's ~8 µs, so there's still room for improvement.

nstarman:
@patrick-kidger is `'__module__' in obj.__dict__` sufficient a check?

nstarman:
Ok. Final approximate timings on that performance notebook:

- Baseline is `@partial(jax.tree_util.register_pytree_with_keys_class)` ~ 7.7 µs
- `Module` before this PR ~ 12.5 µs
- `Module` after this PR ~ 10 µs
- `Module` after this PR, if I delete all wrapper field related code ~ 9 µs

So this PR is a ~50% improvement ((12.5 - 10) / (12.5 - 7.7)) but there's still another 2.3 µs to go.

equinox/_module/_module.py
```diff
@@ -12,119 +11,27 @@
...
 import numpy as np
-from jaxtyping import Array, Bool, PyTree
+from jaxtyping import Array, Bool
 
 from .._filters import is_array, is_array_like, is_inexact_array_like
 from .._pretty_print import tree_pformat
 from .._tree import tree_equal
 from ._better_abstract import better_dataclass, BetterABCMeta
 from ._field import field
+from ._flatten import _generate_flatten_functions, _WRAPPER_FIELD_NAMES
```
patrick-kidger: Nit, I use underscores to indicate private-to-file, so this is not an allowed import.

(Probably just remove the underscores from the names.)

equinox/_module/_prebuilt.py
```diff
@@ -6,7 +6,7 @@
 from jaxtyping import PyTreeDef
 
 from ._field import field
-from ._module import Module, wrapper_field_names
+from ._module import _WRAPPER_FIELD_NAMES, Module
```
patrick-kidger: Nit, the wrapper field names should now be imported from the new file.

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+_WRAPPER_FIELD_NAMES: Final = (
+    "__module__",
+    "__name__",
+    "__qualname__",
+    "__doc__",
+    "__annotations__",
+)
+
+# Indentation for generated code
+INDENT: Final = " " * 4
```
patrick-kidger: Nit, values like these should be named beginning with an underscore. (As these are not public outside this file.)

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+class Sentinel(Enum):
+    """Sentinel values for flattening/unflattening."""
+
+    MISSING = "MISSING"
+
+
+MISSING = Sentinel.MISSING
+
+# Code template for flattening the wrapper fields
+FLAT_WRAPPERS = ", ".join(
```
patrick-kidger: Maybe `GET_WRAPPERS` (or `_GET_WRAPPERS`) for parity with the `SET_` flavour below?

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def make_tuple_type(count: int, element_type: str = "Any") -> str:
+    """Generate a tuple type annotation string for a given count of elements."""
+    if count == 0:
+        return "tuple[()]"
+    elif count == 1:
+        return f"tuple[{element_type}]"
+    else:
+        return f"tuple[{', '.join([element_type] * count)}]"
```
patrick-kidger: I think this also covers the 1 case?

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+    for f in fields:
+        if f.metadata.get("static", False):
+            _static_fs.append(f.name)
+        else:
+            _dynamic_fs.append(f.name)
+    dynamic_fs, static_fs = tuple(_dynamic_fs), tuple(_static_fs)
+
+    # Extract the generated functions from respective namespaces to
+    # set the proper module reference.
+    module_name = getattr(cls, "__module__", "equinox._module._module")
```
patrick-kidger: Should the backup name be either some sentinel `<generated>`, or (if that doesn't work) then for forward-compatibility we can use `__name__` here?

This point does also make me think - this change will probably mean that these functions become unpickleable. I think that's fine - it's not like they're needed to pickle the module object itself - just flagging it.

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+    """Generate a tuple type annotation string for a given count of elements."""
+    if count == 0:
+        return "tuple[()]"
+    elif count == 1:
+        return f"tuple[{element_type}]"
+    else:
+        return f"tuple[{', '.join([element_type] * count)}]"
+
+
+def _generate_flatten_functions(cls: type, fields: tuple[dataclasses.Field[Any], ...]):
```
patrick-kidger: Do we need both arguments? The latter is just `dataclasses.fields(cls)` from the former.

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # Extract the generated functions from respective namespaces to
+    # set the proper module reference.
+    module_name = getattr(cls, "__module__", "equinox._module._module")
+
+    # -------------------------------------------
+    # Generate flatten function
+
+    # Directly access dynamic fields by name
+    if dynamic_fs:
```
patrick-kidger: Nit, I avoid using implicit-cast-to-bool.

Here I think `if len(dynamic_fs) > 0:` would work?

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+        dynamic_vals=dynamic_vals,
+        static_vals=static_vals,
+    )
+
+    # make flatten func
+    flat_ns = NS_BASE | {"module_cls": cls}
+    exec(compile(flat_code, FLAT_FUNC_NAME.format(clsname), "exec"), flat_ns)
+    flat_fn = flat_ns["flatten"]
+    flat_fn.__module__ = module_name
+    object.__setattr__(flat_fn, "__source__", flat_code)
```
patrick-kidger, line 169: Do I understand correctly that `__source__` is not standardized, and this is just a nice addition?

(I like it!)

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+NS_BASE = {
+    "object": object,
+    "Any": Any,
+    "tuple": tuple,
+    "Literal": Literal,
+    "MISSING": MISSING,
+}
+
+
+def make_tuple_type(count: int, element_type: str = "Any") -> str:
```
patrick-kidger: Nit, I try to avoid default arguments in internal-only functions, too easy to make mistakes forgetting to pass them / less readable at the callsite.

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+    exec(compile(flat_code, FLAT_FUNC_NAME.format(clsname), "exec"), flat_ns)
+    flat_fn = flat_ns["flatten"]
+    flat_fn.__module__ = module_name
+    object.__setattr__(flat_fn, "__source__", flat_code)
+
+    # -------------------------------------------
+    # Generate flatten_with_keys function
+
+    if dynamic_fs:
+        key_exprs = [f"(jtu.GetAttrKey({name!r}), obj.{name})" for name in dynamic_fs]
```
patrick-kidger: This uses `obj.name` where the non-key version uses getattr to support `MISSING`. Probably we want the same for both?

equinox/_module/_flatten.py
```diff
@@ -0,0 +1,234 @@
...
+    """Generated {{func_name}} function for {{qualname}}.
+    
+    Dynamic fields: {{dynamic_fs}}
+    Static fields: {{static_fs}}
+    Wrapper fields: {_WRAPPER_FIELD_NAMES}
+    """
+    return (
+        {{dynamic_vals}},
+        (
+            MISSING if '__module__' not in obj.__dict__ else ({FLAT_WRAPPERS}),
```
patrick-kidger: We should probably check if any of the wrapped attributes are present.

patrick-kidger (commented):
Okay, many nits! But I basically really like this, it looks pretty much like exactly what we want to me.

equinox/_module/_flatten.py
nstarman: `_generate_flatten_functions` is called in `_Module.__new__` where `fields` is already computed, so this is faster.

equinox/_module/_flatten.py
nstarman: ✅

equinox/_module/_flatten.py
nstarman: Yeah, I don't know of any way to introspect the source code post `exec(compile())`, so I added `__source__`.

equinox/_module/_flatten.py
nstarman: ✅

equinox/_module/_module.py
nstarman: ✅

equinox/_module/_prebuilt.py
nstarman: ✅

equinox/_module/_flatten.py
nstarman: ✅

equinox/_module/_flatten.py
nstarman: ✅

equinox/_module/_flatten.py
nstarman: ✅

equinox/_module/_flatten.py
nstarman: ✅

equinox/_module/_flatten.py
nstarman: I tried that. It's pretty slow. Can we not just check for one attribute?  https://github.com/patrick-kidger/equinox/pull/1119#issuecomment-3429609710

equinox/_module/_flatten.py
patrick-kidger: In `type(Module).__new__` I think actually, so only when defining a new class. The overhead of this should be unnoticeable I think?

equinox/_module/_flatten.py
patrick-kidger: So in principle someone may have performed any kind of `object.__setattr__(x, "__doc__", my_doc)`, and the goal of these special-cased wrappers was to preserve any/all of these.

Maybe it would suffice to check if `len(obj.__dict__) == len(fields(obj))`, and take the slow path if this fails?

nstarman: Done!

equinox/_module/_flatten.py
nstarman: Small yes, but this is an internal-use-only function. Why not use any pre-computed values?

equinox/_module/_flatten.py
nstarman: I updated it to `cls.__name__`.

nstarman:
@patrick-kidger I think the only possibly unresolved comments are https://github.com/patrick-kidger/equinox/pull/1119#discussion_r2529663221 and https://github.com/patrick-kidger/equinox/pull/1119#discussion_r2529663469, but IMO they're good to go.

nstarman:
I tested this on `unxt` and saw a nice µs speed boost in flattening/unflattening. 
I plan to look at `quax` next for more performance enhancements.

patrick-kidger:
Nice!
Given that this change occurs on the bottom of the tech stack, it would be good to also run the diffrax/optimistix/lineax tests against this too?

Just in case some of the weirder cases (overriding fields with properties etc) catch us out.

equinox/_module/_flatten.py
patrick-kidger: It's not a strong feeling, but my rationale for

> Why not use any pre-computed values?

is greater maintainability / fewer footguns for drive-by contributors.

equinox/_module/_flatten.py
patrick-kidger: What motivates this choice? I don't think I've seen a name used as a module before.

nstarman:
```
uv add --upgrade  git+https://github.com/nstarman/equinox/@module-perf-2 
uv run pytest test -m "not slow" > diffrax_eqx1119_test.log 2>&1
```
[diffrax_eqx1119_test.log](https://github.com/user-attachments/files/23635980/diffrax_eqx1119_test.log)

There is an error in `diffrax`, but when I re-run the tests outside of the custom equinox branch I get the same error, so I don't think it's related.


```
uv add --upgrade  git+https://github.com/nstarman/equinox/@module-perf-2 
uv run --extra tests pytest tests > ~optimistix_eqx1119_test.log 2>&1
```
[optimistix_eqx1119_test.log](https://github.com/user-attachments/files/23634389/optimistix_eqx1119_test.log)


```
uv add --optional tests pytest beartype 
uv run --extra tests pytest tests > lineax_eqx1119_test.log 2>&1
```
[lineax_eqx1119_test.log](https://github.com/user-attachments/files/23635794/lineax_eqx1119_test.log)

johannahaffner:
The diffrax errata were fixed in https://github.com/patrick-kidger/diffrax/pull/696, and should not occur on `dev`, so I can confirm that these are unrelated :)

patrick-kidger:
Awesome! Merged 🎉

PR #1124, opened 2025-10-20
perf: cache dataclass field names in _ModuleMeta

nstarman:
This PR, as named, increased performance during initialization by pre-computing the list of allowed field names and avoiding doing set unions before checking membership. `dataclasses.fields` does stuff beyond returning `__dataclass_fields__`, so it's performant to skip calling this function once per `__setattribute__` call.
@patrick-kidger I've happy to incorporate changes s.t. `__dataclass_field_names__` isn't stored on the class.

### DIFF
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -40,6 +40,7 @@ wrapper_field_names: Final = {
 
 _abstract_module_registry = weakref.WeakSet()
 _has_dataclass_init = weakref.WeakKeyDictionary()
+_module_info = weakref.WeakKeyDictionary()
 _flatten_sentinel = object()
 
 
@@ -394,6 +395,9 @@ class _ModuleMeta(BetterABCMeta):
         if has_dataclass_init:
             cls.__init__.__doc__ = init_doc  # pyright: ignore[reportPossiblyUnboundVariable]
 
+        # Cache the field names for later use.
+        _module_info[cls] = frozenset(f.name for f in fields)
+
         flattener = _ModuleFlattener(fields)  # pyright: ignore[reportArgumentType]
         jtu.register_pytree_with_keys(
             cls,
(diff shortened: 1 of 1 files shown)


nstarman:
See #1125 for fix to linting errors in this PR.

patrick-kidger:
Looks reasonable to me. Indeed I'd prefer not storing them on the class, in favour of simply closing over the names.

nstarman:
@patrick-kidger what about making an ``__equinox_info__`` object on the class that stores some relevant info, like 
```python
class EquinoxInfo(NamedTuple):
    has_dataclass_init: bool
    field_names: frozenset[str]
    ....


class ModuleMeta:
    def __new__(...):
        cls.__equinox_info__ = EquinoxInfo(...)
```

This would, like `__abstractvars__` and `__abstractclassvars__` be a safe addition to the class and would also enable other speedups, like a refactor of (code below) to reduce the need to crawl through an MRO or to store weak-keyed references that need resolving every time. https://github.com/patrick-kidger/equinox/blob/9e2920f7ecf20e5c12ef29e7665dbdbf348413d6/equinox/_module/_module.py#L322-L339

patrick-kidger:
If caching the information is required then we could definitely do this with a (weak) dictionary:
```python
_equinox_info = {}

_equinox_info[cls] = EquinoxInfo(...)
```
I'm reluctant to make it an attribute as this seems like an internal implementation detail, but would definitely become a thing that users depend upon. Introducing new invariants like that makes the job of maintenance much harder.

nstarman:
5.5x speedup of this check in `__setattr__`

CleanShot 2025-10-21 at 15 39 32

nstarman:
Test failure is unrelated, but obv prevents testing this PR...

patrick-kidger:
LGTM! (It's just the spurious JAX 0.8.0 failure we're seeing; the rest of the tests pass.)

Issue #1133, opened 2025-10-24
Option for `EQX_ON_ERROR="off"`


michael-0brien:
I am using `equinox.error_if` in library code, particularly in `__init__`s to check whether or not parameters take on allowed values. It is important that our classes can be instantiated across JIT boundaries, so we must use runtime checking.

Beyond myself, for use in library code it would be very useful to have `EQX_ON_ERROR="off"` to optionally remove all uses of `jax.lax.cond` in compiled programs for end-users (for GPU performance). Would Equinox accept a PR with this feature?

patrick-kidger:
So I considered this, but ended up settling on `EQX_ON_ERROR=nan` instead, which struck me as a safer alternative.

I take your point that this may result in reduced performance - JAX/XLA/etc maybe isn't smart enough to speculatively execute through a `lax.cond`.

I'd be happy to take a PR on this :)

michael-0brien:
This makes sense. This and the nan option both seem like they would have their use cases!

Issue #1120, opened 2025-10-16
Minor (?) test failure with JAX 0.8.0 - no warning emitted if JAX transformation set as attribute


johannahaffner:
I'm getting the following error with JAX 0.8.0: 

```python
(.jax-080) jhaffner@bs-mbpas-0019 equinox % pytest tests/test_module.py 
======================================================================================== test session starts ========================================================================================
platform darwin -- Python 3.13.3, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/jhaffner/Desktop/projects/equinox
configfile: pyproject.toml
plugins: jaxtyping-0.3.3
collected 64 items                                                                                                                                                                                  

tests/test_module.py ................................................F...............                                                                                                         [100%]

============================================================================================= FAILURES ==============================================================================================
______________________________________________________________________________________ test_jax_transform_warn ______________________________________________________________________________________

getkey = GetKey(seed=1816561341, call=11, key=Array([         0, 1816561341], dtype=uint32))

    def test_jax_transform_warn(getkey):
        class A(eqx.Module):
            linear: Callable
    
        class B(eqx.Module):
            linear: Callable
    
            def __init__(self, linear):
                self.linear = linear
    
        for cls in (A, B):
            for transform in (
                jax.jit,
                jax.grad,
                jax.vmap,
                jax.value_and_grad,
                jax.jacfwd,
                jax.jacrev,
                jax.hessian,
                jax.custom_jvp,
                jax.custom_vjp,
                jax.checkpoint,  # pyright: ignore
                jax.pmap,
            ):
>               with pytest.warns(
                    match="Possibly assigning a JAX-transformed callable as an attribute"
                ):
E               Failed: DID NOT WARN. No warnings of type (<class 'Warning'>,) were emitted.
E                Emitted warnings: [].

tests/test_module.py:883: Failed
--------------------------------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------------------------------
testing for: <function jit at 0x10826e8e0>
testing for: <function grad at 0x10826eac0>
testing for: <function vmap at 0x10826f6a0>
testing for: <function value_and_grad at 0x10826eb60>
testing for: <function jacfwd at 0x10826ee80>
testing for: <function jacrev at 0x10826f060>
testing for: <function hessian at 0x10826f1a0>
testing for: <class 'jax._src.custom_derivatives.custom_jvp'>
testing for: <class 'jax._src.custom_derivatives.custom_vjp'>
[...]

PR #1134, opened 2025-10-24
`EQX_ON_ERROR=off` option allows user to switch off `jax.lax.cond` calls

michael-0brien:
#1133 

@patrick-kidger, are there any other tests necessary do you think?

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -110,9 +110,9 @@ def test_nan():
 def test_off_tracetime():
     @jax.jit
     def f(x):
-        return eqx.error_if(x, True, "hi", on_error="off")
+        return eqx.error_if(x, True, "hi", on_error="off") + 1
 
-    _ = f(1.0)
+    assert jnp.isclose(f(1.0), 2.0)
 
 
 def test_off():
@@ -120,7 +120,7 @@ def test_off():
     def f(x, pred):
         return eqx.error_if(x, pred, "hi", on_error="off")
 
-    _ = f(1.0, True)
+    assert jnp.isclose(f(1.0, True), 1.0)
 
 
 def test_assert_dce():


tests/test_errors.py
```diff
@@ -107,6 +107,22 @@ def f(x, pred):
     assert jnp.isnan(y)
 
 
+def test_off_tracetime():
+    @jax.jit
+    def f(x):
+        return eqx.error_if(x, True, "hi", on_error="off")
+
+    _ = f(1.0)
```
patrick-kidger: It would be nice to assert that the output is the expected value, i.e. that this is just the identity function.

tests/test_errors.py
```diff
@@ -107,6 +107,22 @@ def f(x, pred):
     assert jnp.isnan(y)
 
 
+def test_off_tracetime():
+    @jax.jit
+    def f(x):
+        return eqx.error_if(x, True, "hi", on_error="off")
```
patrick-kidger: IIUC then I think XLA will just end up seeing this as the identity function. If so then note that XLA actually has special-casing for such functions – I don't think it matters here but I'd make this `error_if(...) + 1` just in case.

michael-0brien:
Done! Looks like a test in test_module.py is failing; I suspect it is due to a change in JAX 0.8.0.

johannahaffner:
> Looks like a test in test_module.py is failing; I suspect it is due to a change in JAX 0.8.0.

Yes, this was documented in https://github.com/patrick-kidger/equinox/issues/1120, raised in https://github.com/jax-ml/jax/issues/32701 and fixed in https://github.com/jax-ml/jax/pull/32720.

patrick-kidger:
Awesome! LGTM, merged 🚀

PR #1138, opened 2025-11-04
perf: opt for small number of fields

nstarman:
And use in the hash.

For n<~10 tuples are supposed to be faster than frozenset. Using a tuple then enables faster hashing. 

If you think n>10 is common / should have an optimized path, we could detect n in `__new__` and change `_module_info` to storing an object like

```
_module_info = weakref.WeakKeyDictionary()

class ModuleInfo(NamedTuple):
    names_tuple: tuple[str, ...]
    names_set: frozenset[str]
```

### DIFF
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -6,7 +6,16 @@ import types
 import warnings
 import weakref
 from collections.abc import Callable, Hashable
-from typing import Any, cast, Final, Literal, ParamSpec, TYPE_CHECKING, TypeVar
+from typing import (
+    Any,
+    cast,
+    Final,
+    Literal,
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Looks reasonable to me, though I think this may worsen the performance of `__setattr__`, which needs to perform an inclusion check?

nstarman:
I just did performance testing. The frozenset is faster, even for small n number of fields! I'll make the change I suggested in https://github.com/patrick-kidger/equinox/pull/1138#issue-3585880783.

nstarman:
Ok. Now there's no performance hit.

patrick-kidger:
LGTM :)

PR #1140, opened 2025-11-04
perf: localize init check

nstarman:
Avoids `dir` call to reduce initialization time by 55 percentage points!

Top is with this PR.
Bottom is before.

CleanShot 2025-11-04 at 12 16 26@2x


nstarman:
@patrick-kidger I think this PR is ready. It's a massive performance enhancement to initialization.

patrick-kidger:
LGTM!

Issue #1087, opened 2025-08-22
Non-passing equinox tests on GPU


mjo22:
Hi, I could use some helping getting tests to pass locally on GPU. I am getting failures in test_ad.py, test_nn.py, and test_noinline.py. test_noinline.py ends up hanging and I need to quit. This is on equinox version 0.11.0, but I have seen it across versions of equinox. I've also gotten the failure on totally different machines and JAX installs / methods of installs, so I suspect there is just something I don't understand. Any help would be greatly appreciated!

This is the result of my `jax.print_environment_info()`:

```
jax:    0.4.28
jaxlib: 0.4.28.dev20250611
numpy:  2.2.4
python: 3.11.11 (main, Jun 11 2025, 12:47:59) [GCC 13.3.0]
jax.devices (1 total, 1 local): [cuda(id=0)]
process_count: 1
```

And this is the test that fails in test_ad.py:

```python

    def test_pytree_jacfwd():
        class NeuralNetwork(eqx.Module):
            layers: list
            extra_bias: jax.Array

            def __init__(self, key):
                key1, key2, key3 = jax.random.split(key, 3)
                self.layers = [
                    eqx.nn.Linear(2, 8, key=key1),
                    eqx.nn.Linear(8, 8, key=key2),
                    eqx.nn.Linear(8, 2, key=key3),
                ]
                self.extra_bias = jax.numpy.ones(2)

            def __call__(self, x):
                for layer in self.layers[:-1]:
                    x = jax.nn.relu(layer(x))
                return self.layers[-1](x) + self.extra_bias

        def loss(model, x, y):
            pred_y = jax.vmap(model)(x)
            return jax.numpy.mean((y - pred_y) ** 2)

        x_key, y_key, model_key = jax.random.split(jax.random.PRNGKey(0), 3)
        x = jax.random.normal(x_key, (3, 2))
        y = jax.random.normal(y_key, (3, 2))
        model = NeuralNetwork(model_key)
>       assert tree_allclose(
            eqx.filter_grad(loss)(model, x, y), eqx.filter_jacfwd(loss)(model, x, y)
        )
       assert Array(False, dtype=bool)
```

johannahaffner:
Are you sure that the jaxlib version is correct? The Equinox and JAX versions are both roughly two years old, so your jaxlib version ending in 20250611 (11th of June 2025) seems off to me.

[...]

PR #1146, opened 2025-11-16
Make tests pass on nvidia gpu

Artur-Galstyan:
Currently as a draft

### DIFF
--- a/tests/test_sharding.py
+++ b/tests/test_sharding.py
@@ -7,17 +7,20 @@ import jax.tree_util as jtu
 import pytest
 
 
-# Info on the skips: on CPU-only, JAX can simulate mulitple devices
-# but as soon as you have a GPU(s), JAX will use those, so if you have
-# only a single GPU (i.e. a single device) it makes no sense to "test"
-# sharding
-
-
-@pytest.mark.skipif(
-    condition=len(jax.devices()) < 2,
-    reason="Test requires > 1 device to verify implicit sharding propagation",
+pytestmark = pytest.mark.skipif(
+    jax.default_backend() != "cpu",
+    reason=(
+        "Sharding tests require multiple devices. JAX can simulate multiple "
+        "devices on a single CPU but cannot easily do so on a single GPU/TPU. "
+        "Therefore, we skip these tests on non-CPU backends."
+    ),
 )
+
+
 def test_sharding_no_inside_jit():
+    assert len(jax.devices()) > 1, (
+        "Test requires > 1 device to verify implicit sharding propagation"
+    )
     mlp = eqx.nn.MLP(2, 2, 2, 2, key=jr.PRNGKey(0))
 
     num_devices = 2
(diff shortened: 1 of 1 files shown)


tests/test_pmap.py
```diff
@@ -281,4 +282,4 @@ def foo(x):
     y = jax.pmap(foo, out_axes=out_axes)(x)
     z = filter_pmap(foo, out_axes=out_axes)(x)
     assert y.shape == z.shape
-    assert (y == z).all()
+    assert (np.array(y) == np.array(z)).all()
```
Artur-Galstyan, line 285: For the future reviewer: this was needed because JAX can't compare arrays from different devices and I thought it'd be easiest to just move them to CPU using Numpy

tests/test_while_loop.py
```diff
@@ -465,7 +465,13 @@ def body(carry):
...
     # So we have two orders of magnitude safety margin each way, so the test shouldn't
     # be flaky.
-    assert speed < 0.1
+    if jax.default_backend() != "cpu":
+        # for non-cpu cases and because of the checkpointing, there will be a
+        # large spike as soon as the 50_000 threshold is crossed. Also, the overhead
+        # of moving data to the GPU is much higher, so we give it a much wider margins
+        assert speed < 5.0
+    else:
+        assert speed < 0.1
```
Artur-Galstyan: To be honest, it probably makes more sense to skip this on a non-CPU environment. @future_reviewer: wdyt?

tests/test_nn.py
Artur-Galstyan, line 1: Reasoning here was that JAX is not able to compare ints and floats on a GPU, so instead I explicitly cast it to floats, which makes both the cpu and the gpu happy

tests/test_while_loop.py
Artur-Galstyan: Decided to just skip when on non-cpu, otherwise too flaky (can't determine the speeds for all gpus)

tests/test_noinline.py
```diff
@@ -2,10 +2,15 @@
...
 import jax.random as jr
+import pytest
 
 from .helpers import tree_allclose
 
 
+if jax.default_backend() != "cpu":
+    pytest.skip("Skipping tests on non-CPU backend", allow_module_level=True)
+
+
```
Artur-Galstyan, line 13: Skipping this test because of this comment [here](https://github.com/patrick-kidger/equinox/issues/1087#issuecomment-3218009887)

Artur-Galstyan:
The tests now pass on a Mac as well as on a single-GPU machine (RTX 5090). However, I was also using JAX 0.8.1, which released on the 18th of November. I don't think the tests will pass on older versions :(

Artur-Galstyan:
Nice, tests are passing on GitHub too! \**pops champagne\**

tests/test_sharding.py
```diff
@@ -4,13 +4,26 @@
...
+import pytest
 
 
+# Info on the skips: on CPU-only, JAX can simulate mulitple devices
+# but as soon as you have a GPU(s), JAX will use those, so if you have
+# only a single GPU (i.e. a single device) it makes no sense to "test"
+# sharding
+
+
+@pytest.mark.skipif(
```
patrick-kidger: Maybe we could make this condition be CPU-vs-GPU, so that a single-CPU would explicitly fail this test rather than silently passing.

I don't want these tests to accidentally end up always being skipped because we misconfigure the CPU.

Artur-Galstyan: I've added a pytestmarker to skip if on non-cpu and explicitly assert that `len(jax.devices()) > 1` to make sure that the CPU is setup correctly

patrick-kidger:
Merged! 🎉

PR #1148, opened 2025-11-19
types: is_array(_like) provides TypeGuard

nstarman:
This PR:

1. enhances the return type annotations for `is_array` and `is_arraylike` so that they do better type narrowing.
2. updates pyright to get rid of a CI warning (ruff also got updated as part of `uv run pre-commit autoupdate`)
3. simplifies a few functions I saw while addressing pyright errors

### DIFF
--- a/equinox/_filters.py
+++ b/equinox/_filters.py
@@ -1,5 +1,15 @@
 from collections.abc import Callable
-from typing import Any, Final, overload, TypeVar
+from typing import (
+    Any,
+    Final,
+    get_args,
+    overload,
+    TYPE_CHECKING,
+    TypeAlias,
+    TypeGuard,
+    TypeVar,
+)
+from typing_extensions import Never, Protocol
 
 import jax
 import jax.core
@@ -17,31 +27,40 @@ AxisSpec = bool | Callable[[Any], bool]
 #
 
 
-# Workaround https://github.com/patrick-kidger/equinox/issues/1095 in a backward
-# compatible manner.
-_NDARRAY_TYPES: Final = (np.ndarray, np.generic)
-_ARRAY_TYPES = (*_NDARRAY_TYPES, jax.Array)  # JAX < 0.7.2
+# Define array types
+# Note that we don't accept `np.generic` as a suitable duck-type for arrays,
+# because `generic` includes `np.object_` and `np.flexible` which are not
+# array-like.
+NDArrayType: TypeAlias = np.ndarray | np.number | np.bool_
+_NDARRAY_TYPES: Final = get_args(NDArrayType)
+_ARRAY_TYPES = _NDARRAY_TYPES + (jax.Array,)
 
 try:  # JAX 0.7.2
     from jax._src.literals import (
         LiteralArray,  # pyright: ignore[reportAttributeAccessIssue]
     )
-
-    _ARRAY_TYPES += (LiteralArray,)
 except ImportError:
-    pass
+    LiteralArray = Never  # type: ignore[assignment, misc]
+else:
+    _ARRAY_TYPES += (LiteralArray,)
 
 try:  # JAX > 0.7.2
     from jax._src.literals import (
         TypedNdArray,  # pyright: ignore[reportAttributeAccessIssue]
     )
-
-    _ARRAY_TYPES += (TypedNdArray,)
 except ImportError:
-    pass
+    TypedNdArray = Never  # type: ignore[assignment, misc]
+else:
+    _ARRAY_TYPES += (TypedNdArray,)
+
+# Type alias for type checkers. Never disappears from unions: T | Never == T
+if TYPE_CHECKING:
+    ArrayTypes: TypeAlias = NDArrayType | jax.Array | LiteralArray | TypedNdArray  # type: ignore[valid-type]
+else:
+    ArrayTypes = object
 
 
-def is_array(element: Any) -> bool:
+def is_array(element: Any) -> TypeGuard[ArrayTypes]:
     """Returns `True` if `element` is a JAX array or NumPy array."""
     return isinstance(element, _ARRAY_TYPES)
 
@@ -51,7 +70,22 @@ def is_array(element: Any) -> bool:
 _ARRAY_LIKE_TYPES: Final = (*_ARRAY_TYPES, float, complex, bool, int)
 
 
-def is_array_like(element: Any) -> bool:
+class HasJaxArray(Protocol):
+    def __jax_array__(self) -> jax.Array: ...
+
+
+# For static analysis most jax functions don't know they accept `HasJaxArray`
+# objects. So for static analysis we don't include it in the definition of
+# `ArrayLikeTypes`.  However at runtime we do want to accept them because
+# runtime type-checkers like `beartype` will inspect the annotations and will
+# error if `HasJaxArray` *isn't* included.
+if TYPE_CHECKING:
+    ArrayLikeTypes: TypeAlias = jax.typing.ArrayLike
+else:
+    ArrayLikeTypes: TypeAlias = jax.typing.ArrayLike | HasJaxArray
+
+
+def is_array_like(element: Any) -> TypeGuard[ArrayLikeTypes]:
     """Returns `True` if `element` is a JAX array, a NumPy array, or a Python
     `float`/`complex`/`bool`/`int`.
     """
--- a/equinox/_filters.py
+++ b/equinox/_filters.py
@@ -2,14 +2,14 @@ from collections.abc import Callable
 from typing import (
     Any,
     Final,
-    get_args,
+    Literal as L,
     overload,
     TYPE_CHECKING,
     TypeAlias,
     TypeGuard,
     TypeVar,
 )
-from typing_extensions import Never, Protocol
+from typing_extensions import Never, Protocol, runtime_checkable
 
 import jax
 import jax.core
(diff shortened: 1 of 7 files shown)


tests/test_jit.py
```diff
@@ -28,16 +28,18 @@ def test_filter_jit(donate, getkey):
         eqx.nn.MLP(2, 2, 2, 2, key=getkey()),
     ]
     array_tree = [{"a": a, "b": b}, (c,)]
-    mlp_add = jtu.tree_map(lambda u: u + 1 if eqx.is_array(u) else u, general_tree[-1])
+    mlp_add = jtu.tree_map(
+        lambda u: jnp.asarray(u) + 1 if eqx.is_array(u) else u, general_tree[-1]
```
nstarman: `is_array` checks for `np.generic`, but this doen't support `u+1`. Do you want to restrict from `np.generic` to `np.number | np.bool | ...` ?

nstarman:
@patrick-kidger this improvement to the return type is uncovering a real issue: `np.generic` isn't always a full stand-in for `np.ndarray`. Also, and this isn't an issue we can solve, jax has limited static support for `__jax_array__`.

nstarman:
https://github.com/numpy/numpy/pull/30335 will eventually allow for some simplifications!

equinox/_filters.py
```diff
@@ -1,5 +1,15 @@
 from collections.abc import Callable
-from typing import Any, Final, overload, TypeVar
+from typing import (
+    Any,
+    Final,
+    Never,
```
patrick-kidger: At least right now I think we're still only asking for Python 3.10. Let's pull this from `typing_extensions` instead?

equinox/_filters.py
```diff
@@ -1,5 +1,15 @@
...
+    overload,
+    runtime_checkable,
+    TypeAlias,
+    TypeGuard,
```
patrick-kidger, line 9: Perhaps we should check for 3.13+ and use `TypeIs` in that case?

equinox/_filters.py
```diff
@@ -20,28 +30,33 @@
...
-
-try:  # JAX 0.7.2
-    from jax._src.literals import (
-        LiteralArray,  # pyright: ignore[reportAttributeAccessIssue]
-    )
-
-    _ARRAY_TYPES += (LiteralArray,)
-except ImportError:
-    pass
+NDArrayType: TypeAlias = np.ndarray | np.generic
```
patrick-kidger: Do we need both `_NDARRAY_TYPES` and `NDArrayType`? One is a `get_args` of the other.

equinox/_filters.py
```diff
@@ -20,28 +30,33 @@
...
+        )
+    except ImportError:
+        _TypedNdArray = Never  # Never disappears from unions
+    else:
+        _ARRAY_TYPES += (_TypedNdArray,)
+else:
+    _ARRAY_TYPES += (_TypedNdArray,)
 
+# Type alias for type checkers. Never disappears from unions: T | Never == T
+TypedNdArray: TypeAlias = _TypedNdArray  # pyright: ignore[reportInvalidTypeForm]
```
patrick-kidger: Do we need both `TypedNdArray` and `_TypedNdArray`? We could just put `_TypedNdArray` directly into the union.

equinox/_filters.py
```diff
@@ -51,7 +66,15 @@ def is_array(element: Any) -> bool:
 _ARRAY_LIKE_TYPES: Final = (*_ARRAY_TYPES, float, complex, bool, int)
 
 
-def is_array_like(element: Any) -> bool:
+@runtime_checkable
+class HasJaxArray(Protocol):
+    def __jax_array__(self) -> jax.Array: ...
+
+
+ArrayLikeTypes: TypeAlias = ArrayTypes | float | complex | bool | int | HasJaxArray
```
patrick-kidger: Would `jax.typing.ArrayLike` suffice or simplify this?

equinox/nn/_embedding.py
```diff
@@ -97,7 +97,7 @@ def __call__(
         A JAX array of shape `(embedding_size,)`, from the x-th index of the embedding
         table.
         """
-        if is_array_like(x) and jnp.shape(x) == ():
+        if is_array_like(x) and jnp.shape(x) == ():  # pyright: ignore[reportArgumentType]  # for __jax_array__
```
patrick-kidger: If we're going to need `pyright: ignore`s then so are end-users. This would seem to defeat the usefulness of the typeguard in the first place. We could consider just dropping the `__jax_array__` from the static type annotation?

tests/test_jit.py
patrick-kidger: That sounds reasonable to me. I don't think `jnp.array(<some np.generic that is not a number or bool>)` is going to work in general anyway.

equinox/_filters.py
nstarman: I don't think that's what we want in this case. `TypeIs` doesn't enable type narrowing within the union.

equinox/_filters.py
nstarman: ~I can check to make sure, but I think so, for static reasons.~
Edit: yeah, I think it works.

equinox/_filters.py
nstarman: I think it covers everything except HasJaxArray. I can make the appropriate substitution.

equinox/nn/_embedding.py
nstarman: What about beartype checking the type annotation for something that implements `__jax_array__` but is not otherwise Array-like?

patrick-kidger: Maybe we add the `__jax_array__` protocol at runtime only? IIUC this would be on top of `jax.typing.ArrayLike`.

equinox/_filters.py
nstarman: Removed.

equinox/_filters.py
nstarman: Done!

equinox/nn/_embedding.py
nstarman: Done!

equinox/_filters.py
```diff
@@ -17,31 +27,40 @@
...
-# compatible manner.
-_NDARRAY_TYPES: Final = (np.ndarray, np.generic)
-_ARRAY_TYPES = (*_NDARRAY_TYPES, jax.Array)  # JAX < 0.7.2
+# Define array types
+# Note that we don't accept `np.generic` as a suitable duck-type for arrays,
+# because `generic` includes `np.object_` and `np.flexible` which are not
+# array-like.
+NDArrayType: TypeAlias = np.ndarray | np.number | np.bool_
+_NDARRAY_TYPES: Final = get_args(NDArrayType)
+_ARRAY_TYPES = _NDARRAY_TYPES + (jax.Array,)
```
nstarman, line 56: I think the change to type narrow `NDArrayType` from `generic` to `number | bool` means it excludes `bfloat16`!
🤔 
We want to include that, while still excluding `np.object_` and `np.flexible`.

patrick-kidger: Ah, interesting! I think just adding `jax.dtypes.bfloat16` should work?

It would also probably be fine to include `np.generic` as an overestimation, to be robust to some of the more esoteric 4-bit (etc) dtypes as well?

nstarman: That looks to be necessary since jax doesn't provide a nice union of the ml_dtypes that it imports.
I wish they would! Or just subclass `np.number`

nstarman:
Yeesh. Did not expect this small bit of type narrowing to be so difficult. The state of array scalars (instances of np.generic) is a big mess.

equinox/_filters.py
```diff
@@ -17,31 +27,60 @@
...
-    _ARRAY_TYPES += (TypedNdArray,)
 except ImportError:
-    pass
+    TypedNdArray = Never  # type: ignore[assignment, misc]
+else:
+    _ARRAY_TYPES += (TypedNdArray,)
 
+# Type alias for type checkers. Never disappears from unions: T | Never == T
+if TYPE_CHECKING:
+    ArrayTypes: TypeAlias = NDArrayType | jax.Array | LiteralArray | TypedNdArray  # type: ignore[valid-type]
```
patrick-kidger, line 78: The `type: ignore` here is the kind of thing that'll also introduce maintenance overhead, from the oddities between what type checkers accept / how JAX handles `TypedNdArray` / etc all intersect, and how none of these things are stable in how they behave going into the future.

Let's just drop `TypedNdArray` from this union, since it's so rare?

equinox/_filters.py
```diff
@@ -17,31 +27,60 @@
...
 
 
-# Workaround https://github.com/patrick-kidger/equinox/issues/1095 in a backward
-# compatible manner.
+# Define array types
+# Note that `np.generic` covers more than just the array-like dtypes,
+# e.g. `np.object_` and `np.flexible`. But `ml_dtypes` also defines dtypes
+# that inherit from `np.generic` and can't easily be listed here individually.
+@runtime_checkable
+class ArrayLikeGeneric(Protocol):
```
patrick-kidger, line 35: I'd rather keep this file simple than overly precise, I think. JAX will inevitably change something, and I don't want the maintenance burden of tracking things too precisely!

Let's just use `np.generic` and accept that this is an upper bound?

nstarman:
This PR isn't working so well. I'm inclined to close it and try again at some future point.

PR #1167, opened 2026-01-17
feat(nn): Export dot_product_attention functions

### LINKED ISSUE #764: Export attention function similar to torch.nn.functional.scaled_dot_product_attention
Currently, the functions exist in the `_attention.py` file but are not *explicitly* exported. But a lot of people want to write their own custom MHA implementation and could use these functions.

(I'm aware that I can simply import them nonetheless, but because it's not in the docs and not everyone goes through the source code, that can be easily overseen)

WDYT? Other framework have a dedicated "functional" package in them. It'd be great to have something similar.

s27183:
## Summary

- Export \`dot_product_attention\` and \`dot_product_attention_weights\` as public API.
- Closes #764

## Changes
- Added exports in \`equinox/nn/__init__.py\`

## Motivation
As discussed in #764, these functions are useful for users implementing custom attention variants.

### DIFF
--- a/equinox/nn/__init__.py
+++ b/equinox/nn/__init__.py
@@ -1,5 +1,9 @@
 from ._activations import PReLU as PReLU
-from ._attention import MultiheadAttention as MultiheadAttention
+from ._attention import (
+    dot_product_attention as dot_product_attention,
+    dot_product_attention_weights as dot_product_attention_weights,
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Thanks! On balance I'm afraid I'd rather keep these as private implementation details. I'd encourage you to copy-paste them into your own custom attention implementations if you need them though :)

PR #1175, opened 2026-01-28
Add missing axis type to documentation parallelism example

RunnersNum40:
As mentioned and fixed for the test suite in https://github.com/patrick-kidger/equinox/pull/1146, [JAX 0.8.1](https://docs.jax.dev/en/latest/changelog.html#jax-0-8-1-november-18-2025) changed the default `axis_types` of `jax.make_mesh` to `jax.sharding.AxisType.Explicit`. This PR fixes the docs as well, which won't run with an up-to-date version of JAX.


patrick-kidger:
Thank you for the fix! :)

PR #1178, opened 2026-02-02
feat: compiler options

knyazer:
Good morning, Patrick!

Hope you are doing well. I recently needed to use some compiler options with `.compile()` - mainly because of some issues with XLA leading to autotune spiking the memory consumption. Right now we have an ugly piece of code in our codebase that is just a simplified ad-hoc `eqx.filter_jit`. 

So I was wondering if we could upstream the compile options passing into eqx.filter_jit? There are probably a lot of sharp corners I'm cutting here, so I'm wondering if you think there is anything lacking.

Cheers,
Roman

### DIFF
--- a/equinox/_compile_utils.py
+++ b/equinox/_compile_utils.py
@@ -112,9 +112,9 @@ class Lowered(Module):
     def as_text(self):
         return self.lowered.as_text()
 
-    def compile(self):
+    def compile(self, /, compiler_options: dict[str, str | bool] | None = None):
         return Compiled(
-            self.lowered.compile(),
+            self.lowered.compile(compiler_options=compiler_options),
             self.info,
             self.preprocess,  # pyright: ignore
             self.postprocess,  # pyright: ignore
--- a/tests/test_jit.py
+++ b/tests/test_jit.py
@@ -6,6 +6,7 @@ import jax.numpy as jnp
 import jax.random as jrandom
 import jax.tree_util as jtu
 import pytest
+from jax.errors import JaxRuntimeError
 
 from .helpers import tree_allclose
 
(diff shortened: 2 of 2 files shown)


equinox/_compile_utils.py
```diff
@@ -108,13 +108,14 @@ class Lowered(Module):
     info: PyTree
     preprocess: types.FunctionType
     postprocess: types.FunctionType
+    compiler_options: dict[str, bool | int | str] | None = None
 
     def as_text(self):
         return self.lowered.as_text()
 
     def compile(self):
```
patrick-kidger: If the goal is to match the JAX API, would it make more sense to pass the options here to `compile`?

patrick-kidger (commented):
Hello hello :) Absolutely, very happy to support this. Just one minor comment!

equinox/_compile_utils.py
knyazer: What would you say about supporting it both as `eqx.filter_jit` kwarg and `compile` kwarg - e.g. `compile` kwarg would just override the eqx.filter_jit one, if present? I feel like using it as `eqx.filter_jit` kwarg is a bunch more ergonomic, but I agree that it is different from standard jax api.

But I'm also happy to keep it only as the `.compile` kwarg, if you feel strongly about not wiring it through the `eqx.filter_jit`

patrick-kidger: I generally prefer not to do overrides - it makes for a less ambiguous API to have it in just one place.

I think let's put it in `compile` to match JAX :)

knyazer:
@patrick-kidger Done!

Thank you for such an impressive reply speed and the flexibility to integrate these changes :)

patrick-kidger:
Great stuff, LGTM! Merged 🎉 :)

PR #1181, opened 2026-02-11
test: skip jax_pmap_shmap_merge deprecation

patrick-kidger:
Extracted out of https://github.com/patrick-kidger/equinox/pull/1179, opening a new PR trying to figure out why the CI is crashing.

(EDIT: I can't reproduce this locally, but probably it's something funny and OS-specific coupled to how pmap is deprecated. As pmap is now semi-deprecated I'm just going to remove the test.)

### DIFF
--- a/tests/test_module.py
+++ b/tests/test_module.py
@@ -880,7 +880,8 @@ def test_jax_transform_warn(getkey):
             jax.custom_jvp,
             jax.custom_vjp,
             jax.checkpoint,  # pyright: ignore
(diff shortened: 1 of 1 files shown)


lockwo:
Not sure if you saw https://docs.jax.dev/en/latest/migrate_pmap.html, but `smap` seems like the new axis int based pmap tool (I actually haven't tried it, only saw it looking at shard map source code, but looks like it could be good)

PR #1182, opened 2026-02-11
refactor scanMLP; version bump

patrick-kidger:
Tagging @nstarman – staring at the scan+MLP implementation a bit, I think it could be simplified so that the basic flow at both init and call time becomes `if depth == 0: ... elif scan: ... else: ...`. WDYT?

I've also done a version bump so that we can release this.


nstarman:
I like it. Looks much simpler. It will make `https://phasecurvefit.readthedocs.io/en/latest/` simpler too! I can get rid of the custom ScanMLP that started this journey.

nstarman:
🎉

Issue #978, opened 2025-03-18
Filter shard map


lockwo:
With all the new sharding tools, it would be nice to have more equinox versions of these jax functions since they are becoming the accepted way to approach parallelism. While filter_shard can do a lot, certain problems require more tools. One specific example is shard map, which is necessary in cases like https://github.com/jax-ml/jax/issues/26586. I think a filter_shard_map might be nice. Here is a simple example of what that might look like

```python
import equinox as eqx
from jax.sharding import PartitionSpec as P
from jax.experimental.shard_map import shard_map
import jax
import jax.numpy as jnp
from collections.abc import Callable
from typing import Any

class A(eqx.Module):
    a: str

    def __call__(self, x):
        if self.a == "a":
            return x + 1
        return x + 2

def f(a, x):
    return a(x)

a = A("a")

mesh = jax.make_mesh((10,), ("batch",))
spec = P("batch")
sharding = jax.NamedSharding(mesh, spec)
new_a, new_x = eqx.filter_shard((a, jnp.ones(10)), sharding)

class filter_shard_map(eqx.Module):
    f: Callable
    mesh: Any
    in_specs: Any
    out_specs: Any
    check_rep: bool

    def __call__(self, *args):
        arr, static = eqx.partition(args, eqx.is_array)

        def _f(_args):
            a = eqx.combine(_args, static)
            return self.f(*a)
        
        return shard_map(_f, self.mesh, in_specs=self.in_specs, out_specs=self.out_specs, check_rep=self.check_rep)(arr)

# this errors
#sharded_fun = eqx.filter_jit(shard_map(f, mesh, in_specs=spec, out_specs=spec, check_rep=False)) 
# this works
sharded_fun = eqx.filter_jit(filter_shard_map(f, mesh, in_specs=spec, out_specs=spec, check_rep=False))
_ = sharded_fun(new_a, new_x).block_until_ready()
```

I know it is experimental in jax, but there could also be an equinox.experimental.

lockwo:
Sure I get that. I like the idea of eqx.experimental, but I get it's hard to do. The only use case that I know might be relevant is for parallelizing SDE (and maybe ODE) sims (e.g. shard map solves the issues of https://github.com/patrick-kidger/diffrax/issues/407, https://github.com/jax-ml/jax/issues/20968, https://github.com/jax-ml/jax/issues/26586, I still have to rigorously check before closing the issues, but it seems that way so far).

carlesoctav:
[...]

PR #1183, opened 2026-02-16
Add shard map

lockwo:
Realized there wasn't a filter shard map but shard map was no longer experimental (https://github.com/patrick-kidger/equinox/issues/978). Thought it might be useful (since we wanted this feature in practice, see https://github.com/sugolov/jepax/blob/main/jepax/train/train_ijepa.py#L480).

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -84,6 +84,9 @@ plugins:
             - jax.ShapeDtypeStruct
             - jax.extend.core.ClosedJaxpr
             - jax.Device
+            - jax.sharding.Mesh
+            - jax.sharding.AbstractMesh
+            - jax.sharding.PartitionSpec
             - jax.sharding.Sharding
     - mkdocstrings:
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Ah, this looks very well done – but probably isn't something I can maintain, I'm afraid. I don't actually use `shard_map` in any of my workloads, so I'm largely unfamiliar with its details. (Moreover the JAX parallelism stuff has seemed to change in some subtle way every 6 months...)

I think for this reason this is better left in user code, if need be just with ad-hoc partition-shard_map-combine.

lockwo:
Yes jax distributed was a bit of a dynamic process for a while, but it does seem to be settling down into the three views of parallelism (auto, which is in the current example in equinox, explicit which I think can be gotten well enough with the filter_put/sharding_constraint, and shard_map). 

Fair enough on maintainability.

PR #1213, opened 2026-04-15
Add GPJax to list

thomaspinder:
GPJax now builds upon Equinox. It would be nice to be featured in the corresponding lists.


patrick-kidger:
Absolutely, lets add it! It's a bit of an informal rule but the ones in the README are usually the ones I use myself. Let's stick with just the awesome-list?

thomaspinder:
Sure! I've updated.

sammccallum:
Btw, the +'s introduced in front Peex (and now GPJax) are not meant to be there. I think these were introduced during the previous merge!

thomaspinder:
> Btw, the +'s introduced in front Peex (and now GPJax) are not meant to be there. I think these were introduced during the previous merge!

Oh, I had just copied Peex's entry and assumed the plusses were a formatting tag. I've removed the plusses from the entries for GPJax and Peex.

sammccallum:
Beautiful, thanks @thomaspinder!

patrick-kidger:
Merged! Thanks both :)

(It was awesome to see GPJax switching to Equinox btw, very happy to see that :) )

Issue #1216, opened 2026-04-18
Is it deliberate that the `_ModuleMeta` class does not use the generic defined for `Module`?


mishmish66:
Hello! I have noticed that [zuban](https://github.com/zubanls/zuban) gives type issues on subclasses of eqx.Module in a surprising way i.e.

main.py:
```
from jax import random as jr
import equinox as eqx

def foo(lin: eqx.nn.Linear):
    return (lin.weight ** 2).sum()

def main():
    l1 = eqx.nn.Linear(3, 3, key=jr.key(42))
    print(foo(l1))
    
if __name__ == "__main__":
    main()
```

running `uv run zuban check main.py` gives the following result:
```
main.py:9: error: Argument 1 to "foo" has incompatible type "Module"; expected "Linear"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

I am about to create a PR with a hypothetical fix if this was not intentional

patrick-kidger:
Hey there! I'm not familiar with zuban but you can see ove rin #1217 the reason this isn't there already.

Generally speaking, getting metaclasses and typecheckers to play well with each other is a bit tricky, so in practice it's just pyright (as it is ubiquitously the fastest + most correct, though in time perhaps `ty` or zuban will change that) that we aim to support.

mishmish66:
Thanks for the clarification!
It seems that this stems from an unspecified part of the spec I didn't know about. Super interesting deep dive, and I agree with your conclusion.
https://github.com/patrick-kidger/equinox/pull/1217#issuecomment-4284529636

prasanthcakewalk:
Late here, but a type-checker agnostic workaround is to hide the `__call__` method declaration from static type-checkers. Something like:
```python
class _ModuleMeta(BetterABCMeta):
    ...
    ...
    def _call(cls, *args, **kwargs):  # noqa: N805
        ...
        ...

    if not TYPE_CHECKING:
        __call__ = _call

    del _call
```
This will probably have to be done for `BetterABCMeta.__call__` as well (and possibly for `{_ModuleMeta,BetterABCMeta}.__new__`).
The workaround isn't needed for `mypy`, `pyright`, `ty`, or `pyrefly` (for now, at least), so may not be worth it.

PR #1217, opened 2026-04-18
add module metaclass to typevar call

mishmish66:
This addresses #1216, it's just a one line change to use the generic for `_ModuleMeta` in the `__call__` the same way it is used on `Module`'s `__new__` function

### DIFF
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -385,7 +385,7 @@ class _ModuleMeta(BetterABCMeta):
 
         return cls
 
-    def __call__(cls, *args: object, **kwargs: object):  # noqa: N805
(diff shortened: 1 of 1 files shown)


patrick-kidger:
It looks like the failing pre-commit reveals the answer this isn't there already:
```
/home/runner/work/equinox/equinox/equinox/_module/_module.py:388:23 - error: Type of parameter "cls" must be a supertype of its class "_ModuleMeta" (reportGeneralTypeIssues)
```

mishmish66:
Wow, this one was a rabbit hole! It seems that I've stumbled into a place where zuban and pyright differ on their decision on this unspecified part of the type spec.

Definitely does not seem like an issue equinox has to solve but rather a question for zuban and the python type spec

I should've caught that pre-commit message! I didn't use the --all-files flag on the pre-commit and it ignored my changes since they were already committed 🤦

Thanks for the accurate reponse, I will be closing the thread.

PR #1220, opened 2026-04-20
perf: cache per-class field metadata in _ModuleInfo to speed up Module init

nstarman:
`_ModuleMeta.__call__` previously recomputed three things on every instantiation:
- dataclasses.fields(cls) — allocates a new tuple each call
- f.metadata.get("converter") / f.metadata.get("static") — per-field dict lookups
- for parent_cls in ``cls.__mro__``:`` __check_init__`` — full MRO scan each call

All three are now computed once at class-definition time in `_ModuleMeta.__new__` and stored in four new `_ModuleInfo` fields:
- converter_fields: tuple of (name, converter) pairs
- static_field_names: tuple of static=True field names
- non_init_field_names: tuple of init=False field names
- check_init_methods: tuple of unbound `__check_init__` functions from the MRO

``__call__`` now does a single `_module_info[cls]` lookup and iterates over pre-filtered tuples, which are empty for the common case (no converters, no static fields, no `__check_init__`).

Asking Copilot to speed test this PR against main in a few scenarios.
CleanShot 2026-04-21 at 10 16 36

[bench_module_init.txt](https://github.com/user-attachments/files/26936518/bench_module_init.txt)

### DIFF
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -371,9 +371,9 @@ class _ModuleMeta(BetterABCMeta):
         # Cache the field names and per-instantiation metadata for later use.
         names = tuple(f.name for f in fields)
         converter_fields = tuple(
-            (f.name, f.metadata["converter"])
+            (f.name, converter)
             for f in fields
-            if "converter" in f.metadata
+            if (converter := f.metadata.get("converter")) is not None
         )
         static_field_names = tuple(
             f.name for f in fields if f.metadata.get("static", False)
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -48,6 +48,10 @@ class _ModuleInfo(NamedTuple):
 
     names_tuple: tuple[str, ...]
     names_set: frozenset[str]
+    converter_fields: tuple[tuple[str, Any], ...]  # (name, converter) pairs
+    static_field_names: tuple[str, ...]  # names of static=True fields
+    non_init_field_names: tuple[str, ...]  # names of init=False fields
+    check_init_methods: tuple[Any, ...]  # unbound __check_init__ funcs from MRO
 
 
 MSG_METHOD_IN_INIT: Final = """Cannot assign methods in __init__.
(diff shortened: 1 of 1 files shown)


nstarman:
I did a speed profile in `quax` and much of the speed comparison to raw jax relates to costs associated with eqx.Module. 
This PR should address one of the issues. Other issues include:

- `Module.__getattribute__` allocates a new `BoundMethod` Module per method access. This is very expensive.
- `Module.__setattr__` calls `jtu.tree_map` + `dataclasses.fields` per field write.
- `_is_magic`  is slow and can be replaced by a pre-computed frozenset of the non-magic attribute names.
- [x] `dataclasses.fields()` is not cached / using a cache in a few places.
- `_prebuilt.__post_init__` is heavyweight for tiny internal Modules. For example, each `BoundMethod` construction runs `__post_init__`, which goes through the full dataclass + eqx Module initialization chain even though the object has only one or two fields and a lifetime of a single expression.

nstarman:
@patrick-kidger This PR leads to a ~30% speedup of Module initialization and should be ready for review :)

equinox/_module/_module.py
```diff
@@ -403,35 +425,36 @@ def __call__(cls, *args: object, **kwargs: object):  # noqa: N805
...
+            object.__setattr__(self, name, converter(getattr(self, name)))
+
+        # Warn about static fields containing JAX arrays (post-converter values).
+        for name in info.static_field_names:
+            if any(jtu.tree_map(is_array, jtu.tree_leaves(getattr(self, name)))):
+                warnings.warn(
+                    "A JAX array is being set as static! This can result "
+                    "in unexpected behavior and is usually a mistake to do.",
+                    stacklevel=2,
+                )
```
patrick-kidger: This looks like a slight change to what we had before; in particular if the converter produces an inexact array then the not-init warning will not trigger. Is this desirable?

patrick-kidger (commented):
One nitty comment only – it looks like there is a slight change in behaviour – but the basic idea of caching this information makes sense to me.

equinox/_module/_module.py
nstarman: Are converters applied to non-init fields?

nstarman: / should they be?
/ should we conservatively assume they are?

patrick-kidger: so converters are specified at the user level: `field(converter=..., init=...)` so I think a user would expect a converter to be applied.

nstarman: Fixed!

patrick-kidger:
LGTM! Merged 🎉

PR #1222, opened 2026-04-21
Add return type to BoundMethod

nstarman:
Small changes, but does allow for improved type flow through BoundMethod to the bound method's return type.


equinox/_filters.py
```diff
@@ -23,7 +23,7 @@
 _ARRAY_TYPES = (*_NDARRAY_TYPES, jax.Array)  # JAX < 0.7.2
 
 try:  # JAX 0.7.2
-    from jax._src.literals import (
+    from jax._src.literals import (  # pyright: ignore[reportMissingImports]
```
nstarman, line 26: This is not related to the PR, but pyright was failing. 🤷

patrick-kidger:
LGTM!

PR #1223, opened 2026-04-21
Add quax library reference to README

nstarman:
🦆


patrick-kidger:
I don't think this quite meets the 'always useful' bar, Quax is probably the most niche of any of these libraries :)

(I do intentionally try to keep the README list very short.)

PR #1226, opened 2026-04-27
Add progress meter to equinox

ASKabalan:
Migrating ProgressMeter from Diffrax to equinox internal as mentioned in

https://github.com/patrick-kidger/optimistix/pull/202


nstarman:
Just wondering, is there a reason to not use jax-tqdm?

patrick-kidger:
Awesome, merged!

On jax-tqdm: as Equinox is itself the foundation of a lot of other package, then I tend to prefer to minimise dependencies, which will in turn not be installed on everyone else's machines. (In particular right now, we own all of our dependencies other than JAX itself, which is a nice property that it would be a shame to give up.)

PR #1228, opened 2026-05-09
Use `__getattr__` for wrapper field forwarding in BoundMethod

nstarman:
Replaces eager attribute copying in `__post_init__` with lazy forwarding via `__getattr__`, avoiding setattr on an immutable Module and only looking up wrapper fields on demand.
Since BoundMethod is mostly used for calling, this should be a speedup. Limited tests show it is, but speed will vary by exact usage.

### DIFF
--- a/equinox/_module/_prebuilt.py
+++ b/equinox/_module/_prebuilt.py
@@ -32,15 +32,6 @@ class BoundMethod(Module, Generic[_Return]):
     __func__: _FuncDescriptor[_Return] = field(static=True)
     __self__: Module
 
-    def __post_init__(self):
-        for field_name in WRAPPER_FIELD_NAMES:
-            try:
-                value = getattr(self.__func__, field_name)
-            except AttributeError:
-                pass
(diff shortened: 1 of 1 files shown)


nstarman:
A quick test shows it is a 5x for initialization. It's a 4x slowdown for accessing the WRAPPER_FIELD_NAMES specific attributes, but if `__call__` is the primary method used (which it should be) then this is a worthwhile tradeoff.

patrick-kidger:
How much does this actually turn up in practice? Ideally, a lot of this kind of thing all just gets compiled away under JIT?

nstarman:
You're right it shouldn't matter post JIT compilation, but it can be significant during the tracing for compilation.
In one test in `quax` for https://github.com/nstarman/quax/issues/128 I found that in a function a BoundMethod was created over 100 times, so a 5x speedup is meaningful.

patrick-kidger:
Fair enough! Merged! :)

PR #1240, opened 2026-06-11
Cache `WeightNorm._norm` so that identical modules compare equal

### LINKED ISSUE #965: `WeightNorm` causes unexpected PyTrees inequality
The following assertion fails.
```
import equinox as eqx
import equinox.random as jrandom

layer = eqx.nn.Linear(2, 2, key=jrandom.key(0))
m1 = eqx.nn.WeightNorm(layer)
m2 = eqx.nn.WeightNorm(layer)
assert eqx.tree_equal(m1, m2)  # unequal
```
The reason of failure is that `WeightNorm._norm` is different (== operation returns False) for every `WeightNorm` instance.

This also causes the failure of using `tree_equal` to check deserialized models.

discobot:
Fixes #965.

`WeightNorm.__init__` built a fresh `ft.partial` for `_norm` on every
instantiation, which (as suspected in the issue) only ever compares by
identity. Of the two options suggested there, this takes the cache-on-`axis`
route: `_norm` is now produced by a module-level `_make_norm(axis)` factory
wrapped in `ft.cache`, so instances with the same `axis` share one callable
and `tree_equal` holds.

I went with caching rather than `eqx.Partial` because the leaf stays a plain
`functools.partial`: the pytree structure is unchanged (friendlier to existing
treedefs and serialised models), and `_norm` remains a dynamic field that can
be patched with `eqx.tree_at` as before.

One extra finding from investigating: the per-instance callable also made
`eqx.filter_jit` retrace once per `WeightNorm` instance; traces are now
shared.

Added `test_weight_norm_equality` covering `tree_equal` for `axis=0` /
`axis=None` and a single-trace check under `filter_jit`. `tests/test_nn.py`,
`tests/test_tree.py` and `tests/test_serialisation.py` all pass locally (CPU).

### DIFF
--- a/equinox/nn/_weight_norm.py
+++ b/equinox/nn/_weight_norm.py
@@ -21,6 +21,18 @@ def _norm_except_axis(v: Array, norm: Callable[[Array], Scalar], axis: int | Non
         return jax.vmap(norm, in_axes=axis, out_axes=axis)(v)
 
 
+@ft.cache
+def _make_norm(axis: int | None) -> Callable[[Array], Scalar]:
+    # Cached so that two `WeightNorm`s with the same `axis` share the same callable,
+    # and so compare equal under `eqx.tree_equal`.
+    # https://github.com/patrick-kidger/equinox/issues/965
+    return ft.partial(
+        _norm_except_axis,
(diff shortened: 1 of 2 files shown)


patrick-kidger:

PR #1242, opened 2026-06-13
Fix eager `error_if` hanging when inputs are sharded over multiple devices

### LINKED ISSUE #1232: `equinox.error_if` hangs on error if one of the arguments has a multi-device replicated sharding
In eager mode, `equinox.error_if(x, cond, msg)` hangs forever in an XLA all-reduce rendezvous when the truthy `cond` (or `x`) lives on a `NamedSharding` whose mesh has more than one device, even though the value is fully replicated.

Expected behavior: raise `EquinoxRuntimeError(msg)` exactly as it does for a single-device array.

### Minimal reproducer

```python
import os
os.environ['XLA_FLAGS'] = '--xla_force_host_platform_device_count=2'

import jax
import jax.numpy as jnp
import equinox

mesh = jax.sharding.Mesh(jax.devices('cpu'), ('x',))
replicated = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec())

x = jax.device_put(jnp.array(1.0), replicated)
cond = jax.device_put(jnp.array(True), replicated)

try:
    equinox.error_if(x, cond, 'err')
except equinox.EquinoxRuntimeError:
    pass
```

```
E0519 23:08:20.320073 52455925 rendezvous.cc:116] [id=0] This thread has been waiting for `all reduce RendezvousKey{run_id=RunId: -1375800151, global_devices=[0, 1], num_local_participants=2, collective_op_kind=cross_module, op_id=1}` for 20 seconds and may be stuck. Expected 2 threads to join the rendezvous, but not all of them arrived on time.
E0519 23:09:00.321819 52455925 rendezvous.cc:41] All thread stack traces (if supported on this platform):
E0519 23:09:00.321926 52455925 rendezvous.cc:43] 
F0519 23:09:00.321934 52455925 rendezvous.cc:161] [id=0] Termination timeout for `all reduce RendezvousKey{run_id=RunId: -1375800151, global_devices=[0, 1], num_local_participants=2, collective_op_kind=cross_module, op_id=1}` of 40 seconds exceeded. Exiting to ensure a consistent program state. Expected 2 threads to join the rendezvous, but only 1 of them arrived on time.
*** Check failure stack trace: ***
    @        0x11d42a900  absl::lts_20260107::log_internal::LogMessage::Flush()
    @        0x11fd725c0  xla::internal::AwaitAndLogIfStuck()
    @        0x11fd5c86c  xla::cpu::InProcessCommunicator::AllReduce()
    @        0x11e1157ac  absl::lts_20260107::internal_any_invocable::RemoteInvoker<>()
    @        0x11e99e1ac  xla::cpu::CollectiveThunk::ExecuteWithCommunicator()
    @        0x11e115080  xla::cpu::AllReduceThunk::Execute()
    @        0x11fd50188  xla::cpu::ThunkExecutor::ExecuteSequential()
    @        0x11e119b0c  xla::cpu::ConditionalThunk::Execute()
    @        0x11fd50188  xla::cpu::ThunkExecutor::ExecuteSequential()
    @        0x11d6b3f3c  xla::CpuPjRtRawLoadedExecutable::Execute()::$_0::operator()()
    @        0x11d6bb8a0  absl::lts_20260107::internal_any_invocable::RemoteInvoker<>()
    @        0x11d6f3304  std::__1::__function::__func<>::operator()()
    @        0x120996e80  Eigen::ThreadPoolTempl<>::WorkerLoop()
    @        0x11d300100  Eigen::ThreadPoolTempl<>::WorkerLoop()
    @        0x11d2ffef8  std::__1::invoke[abi:nn190102]<>()
    @        0x11d2f0f90  tsl::(anonymous namespace)::PThread::ThreadFn()
    @        0x183587c58  _pthread_start
    @        0x183582c1c  thread_start
```

Observations:

- Placing only `x` (and leaving `cond` unsharded) or only `cond` on the multi-device replicated sharding is sufficient to trigger the hang.
- With `cond=False` (no error to raise) the call returns normally.
- With a single-device mesh (`--xla_force_host_platform_device_count=1`) the call raises as expected.

### Related

- #939: raises the same overall topic of `error_if` interacting with multi-device shardings, but focuses on the all-reduce overhead in the no-error case rather than the eager-mode hang reported here.
- #342: in-progress work on the same area, currently parked waiting on JAX fixes.

### Note

Written by Opus 4.7, edited by a human.

### System info

```
jax:    0.10.0
jaxlib: 0.10.0
numpy:  2.4.6
python: 3.14.0 (main, Oct 14 2025, 21:10:22) [Clang 20.1.4 ]
device info: cpu-1, 1 local devices"
process_count: 1
platform: uname_result(system='Darwin', release='25.4.0', version='Darwin Kernel Version 25.4.0: Thu Mar 19 19:30:44 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6000', machine='arm64')
equinox: 0.13.6
```

discobot:
Fixes #1232.

The hang happens because eager `error_if` routes through `filter_jit`, so a
multi-device-sharded input compiles an SPMD program over every device in the
sharding. The all-reduce in the abort trace sits inside the `lax.cond` error
branch: once the exception from `jax.pure_callback` kills one device's
execution thread, its peers never reach the collective and XLA aborts the
process when the rendezvous times out.

This PR handles that case before the jit: if every input is concrete (no
tracers) and fully addressable, and at least one is sharded over more than one
device, then the predicate is evaluated on the host. On error, the inputs are
copied to the host and passed to the existing `branched_error_if_impl_jit`,
which then compiles a single-device program -- so the usual
`EquinoxRuntimeError` reporting is unchanged. For the non-raising `on_error`
modes, each output leaf's original sharding is restored with `jax.device_put`.
Traced or non-fully-addressable inputs fall through to the current behaviour.

One known limitation: `error_if` under a user's `jax.jit` over multi-device
inputs still aborts in the same way (verified locally). There the multi-device
program is what the user asked for, and the underlying problem -- a raising
callback inside an SPMD program -- doesn't look fixable from inside `error_if`.
This PR covers eager mode, as reported.

Tested with the two CPU devices already configured in `tests/conftest.py`: the
new test aborts the process (exit 134) without the fix and passes with it,
covering replicated and non-replicated shardings, each of `x`/`pred` sharded
alone, and sharding preservation on the false-predicate and
`on_error="nan"`/`"off"` paths.

### DIFF
--- a/equinox/_errors.py
+++ b/equinox/_errors.py
@@ -287,6 +287,39 @@ def _error_outer(
 _error_outer_jit = _jit.filter_jit(_error_outer)
 
 
+def _eager_single_device(leaf) -> jax.Device | None:  # pyright: ignore[reportInvalidTypeForm]
+    # The device to gather `leaf` onto for an eager error check, or `None` to leave it
+    # untouched. Only a fully-addressable array spread over more than one device needs
+    # moving; single-device inputs (the common case) are left untouched.
+    if isinstance(leaf, jax.core.Tracer):
+        return None
+    if isinstance(leaf, jax.Array) and leaf.is_fully_addressable:
+        devices: set[jax.Device] = leaf.sharding.device_set  # pyright: ignore[reportInvalidTypeForm]
+        if len(devices) > 1:
+            return min(devices, key=lambda d: d.id)
+    return None
+
+
+def _to_single_device(leaf):
+    device = _eager_single_device(leaf)
+    return leaf if device is None else jax.device_put(leaf, device)
+
+
+def _restore_sharding(out: PyTree, like: PyTree) -> PyTree:
+    def restore(out_leaf, like_leaf):
+        if (
+            isinstance(out_leaf, jax.Array)
+            and isinstance(like_leaf, jax.Array)
+            and not isinstance(out_leaf, jax.core.Tracer)
+            and not isinstance(like_leaf, jax.core.Tracer)
+            and out_leaf.sharding != like_leaf.sharding
+        ):
+            return jax.device_put(out_leaf, like_leaf.sharding)
+        return out_leaf
+
+    return jtu.tree_map(restore, out, like)
+
+
 def _error_impl(
     x: PyTree,
     pred: Bool[ArrayLike, ""],
@@ -300,7 +333,18 @@ def _error_impl(
     # a bool and an int.
     # This ensures we can use `error_if` before init_google.
     if any(is_array(leaf) for leaf in leaves):
-        return _error_outer_jit(x, pred, msg, on_error=on_error, category=category)
+        # In eager mode the error check below is jitted. If an input is sharded over
+        # multiple devices this compiles an SPMD program, and when the error fires the
+        # exception raised inside `pure_callback` tears down one device whilst its peers
+        # hang forever at the next collective, aborting the whole process.
+        # (https://github.com/patrick-kidger/equinox/issues/1232)
+        # Pin the eager check to a single device -- a no-op when the inputs already live
+        # on one -- then put each output leaf back onto its original sharding.
+        single_x, single_pred = jtu.tree_map(_to_single_device, (x, pred))
+        out = _error_outer_jit(
+            single_x, single_pred, msg, on_error=on_error, category=category
+        )
+        return _restore_sharding(out, x)
     else:
         return _error_outer(x, pred, msg, on_error=on_error, category=category)
 
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -195,6 +195,44 @@ def test_traceback_runtime_custom():
         assert "EQX_ON_ERROR" not in str(e)
 
 
+# https://github.com/patrick-kidger/equinox/issues/1232
+def test_multi_device_eager():
+    # Without the fix, the error cases below deadlock at an XLA collective and then
+    # abort the whole process. `conftest.py` runs the suite with two CPU devices.
+    mesh = jax.sharding.Mesh(jax.devices("cpu"), ("x",))
+    replicated = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec())
+    x = jax.device_put(jnp.array(1.0), replicated)
+    true_pred = jax.device_put(jnp.array(True), replicated)
+    false_pred = jax.device_put(jnp.array(False), replicated)
+
+    # (x sharded, pred sharded), (only x sharded), (only pred sharded).
+    for x_i, pred_i in [(x, true_pred), (x, True), (jnp.array(1.0), true_pred)]:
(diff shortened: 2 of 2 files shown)


patrick-kidger:
Hi there! Thank you for this, and sorry for taking a few weeks to get around to it :)

The reason for the delay is that we've been beavering away at landing a number of other improvements to `error_if` – see https://github.com/patrick-kidger/equinox/pull/762, just merged into `dev` – and I wanted to land that first :)

Would you be interested in rebasing this branch on top of `dev`, and seeing how things go?

I need to stare at the implementation a little bit – I'm not an expert on multi-device JAX – but I don't love the special-casing of this case. Would it be possible to just 'always' move to a single device... which will be a no-op in the inherently-single-device case?

patrick-kidger:
Whoops sorry, this fell of the bottom of my to-do list. Coming back around to it now: this is very clean to read, thank you!

I do have one concern, around the performance – presumably this will result in a lot of data being sent back-and-forth? It seems like this is going to be a huge performance pessimization in the common case that the `error_if` is not triggered. It seems to me like it may actually be preferable to hang on the failure case than to always be slower in the happy case.

Do you think you can see a way to resolve that? I can imagine that this should be doable somehow, by only actually broadcasting the 'do we error or not' flag amongst the different devices.

spraharsh:
I have a few opinions on how this should be done, 

For the SPMD case instead of evaluating the predicate on the host

Here's an outline of how this should be solved depending on how leaves of x are sharded. Let us take an array leaf of x, `xleaf` and an axis `i`, if `x` is distributed along `i` for example, we get local predicate shards that must be collectively reduced with an OR to get whether error_if should fire.


#### If predicate is duplicated along `i`

1. Evaluate the predicate
2. The error should be handled on each individual device as seen fit, (i.e it may interact with other distributed arrays in downstream code)

This way there is no need for communication along devices and each processor should handle the error as it sees fit, (and any stop to the program should come downstream if there is an interaction)

#### If the predicate is distributed along `i`.

2. [Allreduce](https://huggingface.co/spaces/nanotron/ultrascale-playbook?section=reduce_&_allreduce) the distributed local (unreduced) predicates along i with an (OR), giving duplicated global predicates along i 
3. handle the error as you see fit on each device with 


I think this way, we only have extra comms if necessary without unecessary duplications, comms with host, or compute and also keeps the SPMD flow


## Edit:

^^ Sorry I realized that the original code did not support shard map. imo this needs to be split into two different issues

1. Support/Logic for shard_map/Explicit shardings that imo should have an allreduce under the hood (as seen in the automatic sharding case). I submitted a PR in https://github.com/patrick-kidger/equinox/pull/1262 which illustrates this can be done with just `pmax` as a collective (allreduced) OR

2. the eager_if issue which I left open

PR #1243, opened 2026-06-13
Don't warn about NumPy scalars in static fields

### LINKED ISSUE #863: "JAX array is set a static" warning is raised unwantedly
As of Equinox 0.11.6 and https://github.com/patrick-kidger/equinox/pull/800, the following MWE raises a `UserWarning: A JAX array is being set as static! This can result in unexpected behavior and is usually a mistake to do.`

```python
import numpy as np
import equinox as eqx

class Foo(eqx.Module):
    x: tuple[int, int] = eqx.field(static=True)

    def add_one(self):
        x_as_np = np.asarray(self.x)
        return Foo(tuple(x_as_np+1))

x = (3, 2)
foo = Foo(x)
foo.add_one()
# UserWarning: A JAX array is being set as static! This can result in unexpected behavior and is usually a mistake to do.
```

This means that one cannot perform numpy operations (which is often simpler than writing them in plain python) on a static attribute. This is a use-case we have in dynamiqs, see for instance [the method `__mul__`](https://github.com/dynamiqs/dynamiqs/blob/ca5590653df31ca91b28a16969840002e5bb20ad/dynamiqs/qarrays/sparse_dia_qarray.py#L220-L242) of [this class](https://github.com/dynamiqs/dynamiqs/blob/ca5590653df31ca91b28a16969840002e5bb20ad/dynamiqs/qarrays/sparse_dia_qarray.py#L26-L28) which represents an array in diagonal (DIA) sparse format. Note that we intentionally use `numpy` instead of `jax.numpy` to have "static" logic.

discobot:
Fixes #863.

The spurious warning comes from the static-field check in
`_ModuleMeta.__call__`, which maps `is_array` over the static leaves —
and `is_array` matches `np.generic`, so hashable NumPy scalars trip it
alongside actual arrays.

This implements the fix suggested in the thread: exclude `np.generic` from
that check. Static `np.ndarray`/`jax.Array` leaves still warn as before, and
the warning message text is unchanged so existing `filterwarnings` matches
keep working. (Happy to switch to the other option discussed — warning on any
unhashable leaf — if you'd prefer.)

Tests: extended `test_no_jax_array_static` with a still-warns case for a
static tuple containing an `np.ndarray`, and a must-not-warn case for a static
tuple of NumPy scalars; the latter fails before this change.

### DIFF
--- a/equinox/_module/_module.py
+++ b/equinox/_module/_module.py
@@ -444,9 +444,13 @@ class _ModuleMeta(BetterABCMeta):
             if any(jtu.tree_map(is_inexact_array_like, jtu.tree_leaves(val))):
                 warnings.warn(_MSG_FIELD_INIT_FALSE, stacklevel=2)
 
-        # Warn about static fields containing JAX arrays (post-converter values).
+        # Warn about static fields containing JAX or NumPy arrays (post-converter
+        # values). NumPy scalars (`np.generic`) are excluded: unlike arrays they are
+        # hashable, and so are safe to use as static fields.
+        # https://github.com/patrick-kidger/equinox/issues/863
         for name in info.static_field_names:
-            if any(jtu.tree_map(is_array, jtu.tree_leaves(getattr(self, name)))):
+            leaves = jtu.tree_leaves(getattr(self, name))
+            if any(is_array(x) and not isinstance(x, np.generic) for x in leaves):
                 warnings.warn(
                     "A JAX array is being set as static! This can result "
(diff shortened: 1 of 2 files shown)


patrick-kidger (commented):
I'm not sure this change is desirable, I'm afraid. The static-or-not question is largely decided not by hashability, but by whether or not `jax.jit` will trace the argument. I believe it will trace a numpy scalar.

PR #1245, opened 2026-06-16
Replace `_safe_zip` with built-in `zip(..., strict=True)`

paulgekeler:
**Summary**
Replaces the internal `_safe_zip` with built-in `zip(..., strict=True)`

**Motivation**
Equinox already requires Python 3.10+, allowing the use of the `strict` argument.

**Testing**
Ran tests; All pass.


patrick-kidger:
LGTM! ty :)

PR #1247, opened 2026-06-23
Fix broken arXiv link in `RMSNorm` documentation

himoto:
The RMSNorm documentation currently links to a non-working arXiv URL. This PR updates the reference to a working URL for the cited paper.


patrick-kidger:
Ty! I'm not sure where the 'browse.' ever came from!

PR #1249, opened 2026-07-18
Add a paragraph about how to use jaxon for serialisation to the documentation.

42Ar:
This pull request adds a paragraph to the documentation at `docs/examples/serialisation/` that informs the reader that the external library Jaxon can be used for serialisation.

The paragraph was added because using Jaxon has advantages over the currently recommended approach. With Jaxon, Hyperparameters can be arbitrary PyTrees and the user does not need to provide a "skeleton" pytree for deserialisation. Furthermore, Jaxon uses the HDF5 format which is well established for storing large amounts of structured data, including hierarchical metadata and n-dimensional arrays. Jaxon is compatible with equinox, because all Equinox modules are dataclasses, which are supported by Jaxon.

For now, only this paragraph is added to the documentation. In the future, it might make sense to fully integrate the library into Equinox.


patrick-kidger:
Nice, thank you!

I'd be inclined to avoid including installation/usage examples here, in case things change in Jaxon. Rather, it might be neater to add an admonition (with the `!!!` syntax) that just says "please go see this repo".

42Ar:
Thanks for your comment. I removed the installation instruction and the example and shortened the paragraph a bit :)

docs/examples/serialisation.ipynb
patrick-kidger, line 1: have you tested this locally? IIRC there needs to be a blank line between the `!!!` and its content.

42Ar: It works locally, so I believe the blank line is optional. I've added it now for style consistency.

patrick-kidger:
Merged! 🎉

PR #1250, opened 2026-07-20
Add adv-jax-math to awesome list

### LINKED ISSUE #1215: Sparse pullbacks, chunked vmap, and sharded transformations
I have authored variations of `jax.vmap` and the like which support features that are necessary for reasonable performance in many applications. Rather than having these robust implementations buried in my library, I think it would benefit the community if they could be made easily available to the community through a package like `equinox`.

I think `equinox` is a better fit than `jax` itself for this because the functions work well with `equinox`'s custom versions of `jax`'s most used transformations for `vjp`, `jvp`, `filter_jit`, `closure_convert` etc.

If `equinox` is not the package to support these useful transformations, perhaps this issue can lead to the creation of an external package solely dedicated to having a consistent set of these transformations. The performance of these transformations essentially determines the performance of automatic differentiation, vectorization, and parallelization  in `JAX`.

## Proposed transformations

The transformations that I propose adding are (can pick your preferred names of course)
- `vmap_chunked`
  - Basically, we often need to evaluate some expensive function over a bunch of inputs and then reduce it (e.g. sum). Currently the two main options are a single loop which is slow on GPU, or a full `vmap` which, despite JIT's best abilities, often results in materializing the full array. This tanks the memory of course, but often also tanks the compute speed because allocating so much memory is more expensive operation than doing the computations in sequence. Using `vmap_chunked` solves this.
  - The [vmap_chunked code](https://github.com/PlasmaControl/DESC/blob/8d485390d7f2899e8976284891cd61e06f2afc27/desc/batching.py#L183)[^1] is essentially ``jax.sequential_vmap`` but it supports choosing an arbitrary chunk size so that vmap can be done in chunks of `chunk_size` to recover fast performance.
  - An additional feature is that it also supports a reduction operation across chunks, so that the max memory usage is at most ``chunk_size``. For example, when ``chunk_size=1``, it consumes no more memory than a `scan`.
  - With a [single line change [2]](https://github.com/jax-ml/jax/blob/4d5d324fb678264a3059acf4da2facada905737a/jax/_src/api.py#L837) of replacing `vmap` with `vmap_chunked` in JAX's code for `jacrev` and `jacfun` the performance of computing the Jacobian of a function is greatly improved.
- `sparse_pullback`
  - This is a really useful decorator that painlessly enables doing sparse autodiff while still using dense computations for large performance improvements in compute speed and memory reduction.
  - See [this summary [3]](https://github.com/jax-ml/jax/issues/36862#issuecomment-4270887244) for its utility. The implementation can be found at [sparse pullback code [4]](https://github.com/PlasmaControl/DESC/blob/c8dd59e4cfafab29fb5f92a9b2d3597168364e3b/desc/derivatives.py#L98).
- I recall that @PhilipVinc has also written a transformation similar to `vmap_chunked` except that it shards the chunks onto the different devices. That would also be useful, though perhaps supporting sharding like this could be higher maintenance.

[^1]: The warnings are issues with `jax.vmap`. The `vmap_chunked` code has no bugs.

unalmis:
https://github.com/patrick-kidger/equinox/issues/1215#issuecomment-4279225663
> please also feel free to send a PR to the Equinox awesome-list in our docs

Resolves #1215

### DIFF
--- a/docs/awesome-list.md
+++ b/docs/awesome-list.md
@@ -138,5 +138,5 @@ Jump Processes in JAX
 **isax**: [Github](https://github.com/lockwo/isax)  
 Ising (like) models in JAX
 
-**adv-jax-math**: [Github](https://github.com/unalmis/adv-jax-math)
+**adv-jax-math**: [Github](https://github.com/unalmis/adv-jax-math)  
 Batching and sharding utilities for vectorization and differentiation in JAX.
--- a/docs/awesome-list.md
+++ b/docs/awesome-list.md
@@ -137,3 +137,6 @@ Jump Processes in JAX
 
 **isax**: [Github](https://github.com/lockwo/isax)  
 Ising (like) models in JAX
+
+**adv-jax-math**: [Github](https://github.com/unalmis/adv-jax-math)
+Batching and sharding utilities for vectorization and differentiation in JAX.


docs/awesome-list.md
```diff
@@ -137,3 +137,6 @@ Jump Processes in JAX
 
 **isax**: [Github](https://github.com/lockwo/isax)  
 Ising (like) models in JAX
+
+**adv-jax-math**: [Github](https://github.com/unalmis/adv-jax-math)
```
patrick-kidger: need a double space at the end here, in order to be rendered as a newline :-)

unalmis:
ready to merge

patrick-kidger:
merged! 🎉

PR #1253, opened 2026-07-27
Treat JAX and NumPy arrays as interchangeable when deserialising

### LINKED ISSUE #861: `eval_shape` incompatible with deserializing directly to host due to `__assert_same`
Suppose you have a large pytree where you want to ensure that the full tree is never on the JAX device (TPU/GPU). You might also want to minimize the allocation of transient arrays by using `eval_shape`. Your ser/de code would then look something like this:

```py
from __future__ import annotations

from pathlib import Path
from typing import TypeVar, TypeVarTuple

import equinox as eqx
import equinox._serialisation as eqx_ser
import jax
from numpy import ndarray

Shape = TypeVarTuple("Shape")
DType = TypeVar("DType")


def save(path: Path, array: ndarray[*Shape, DType]):
    with path.open("wb") as f:
        # We have to convert to JAX arrays because numpy doesn't handle bfloat16
        # https://github.com/jax-ml/ml_dtypes/issues/41
        eqx.tree_serialise_leaves(f, array)


def load(path: Path, like_fn: Callable[[], ndarray[*Shape, DType]]) -> ndarray[*Shape, DType]:
    with path.open("rb") as f:
        return eqx.tree_deserialise_leaves(
            f,
            eqx.filter_eval_shape(like_fn),
            filter_spec=lambda f, x: jax.device_get(eqx.default_deserialise_filter_spec(f, x)),
        )
```

But that errors with:

```
File …/lib/python3.11/site-packages/equinox/_serialisation.py:172, in _assert_same.<locals>._assert_same_impl(path, new, old)
    170     typeold = array_impl_type
    171 if typenew is not typeold:
--> 172     raise RuntimeError(
    173         f"Deserialised leaf at path '{jtu.keystr(path)}' has changed type from "
    174         f"{type(old)} in `like` to {type(new)} on disk."
    175     )
    176 if isinstance(new, (np.ndarray, jax.Array)):
    177     if new.shape != old.shape:

RuntimeError: Deserialised leaf at path '' has changed type from <class 'jax._src.api.ShapeDtypeStruct'> in `like` to <class 'numpy.ndarray'> on disk.
```
(The error message is slightly misleading in this case because the actual comparison we're performing and failing is between `jaxlib.xla_extension.ArrayImpl` (i.e. `array_impl_type`) and `numpy.ndarray`.)

Note that users can circumvent the issue by monkey-patching out the check but that's pretty ugly:

```py
def patched_assert_same(array_impl_type):  # type: ignore
    """Equinox generates a fixed `array_impl_type` that corresponds to a JAX array.
    Then `_assert_same_impl` swaps in this type for any `jax.ShapeDtypeStruct` in the `like`.
    It then compares the `like` types and types at the very end of deserialization.
    But this means we're forbidden from deserializing to the host with `eval_shape` and
    would instead have to deserialize the whole tree on device and transfer to the host.
    """

    def _assert_same_impl(path, new, old):  # type: ignore
        pass

    return _assert_same_impl


eqx_ser._assert_same = patched_assert_same  # type: ignore
```

JOhnsonKC201:
Fixes #861.

### Problem

Deserialising into a `filter_eval_shape` skeleton fails if the `filter_spec`
puts the result on the host:

```python
like = eqx.filter_eval_shape(eqx.nn.Linear, 2, 2, key=key)
eqx.tree_deserialise_leaves(
    path, like,
    filter_spec=lambda f, x: jax.device_get(eqx.default_deserialise_filter_spec(f, x)),
)
# RuntimeError: Deserialised leaf at path '.weight' has changed type from
# <class 'jax.ShapeDtypeStruct'> in `like` to <class 'numpy.ndarray'> on disk.
```

Shape and dtype match. Only the array kind differs.

### Cause

`_assert_same_impl` substitutes the concrete array type for a
`ShapeDtypeStruct` in `like`, then compares with `typenew is not typeold`. That
identity check leaves no room for a `filter_spec` to hand back a NumPy array
instead of a JAX one.

### Fix

Compare array leaves by array-ness instead of exact type, which is what you
suggested in the issue: "Maybe we just consider all kinds of JAX and NumPy array
interchangeable?".

Every non-array leaf keeps the strict identity check, and the shape and dtype
checks immediately below are untouched, so real mismatches still raise. I
verified both: a shape mismatch still errors, and the behaviour for a
non-array `like` leaf is byte-for-byte the same as on `main`.

### Test plan

- New `test_eval_shape_deserialise_to_host` in `tests/test_serialisation.py`.
  It fails on `main` with the `RuntimeError` above and passes with this change.
- Full suite: 472 passed, 4 skipped (baseline on a clean checkout was 471
  passed, 4 skipped, so the only difference is the added test).
- `ruff check` and `ruff format --check` clean on both files.

### DIFF
--- a/equinox/_serialisation.py
+++ b/equinox/_serialisation.py
@@ -186,13 +186,22 @@ def _maybe_open(
         yield path_or_file
 
 
+def _is_array_type(x) -> bool:
+    return isinstance(x, type) and issubclass(x, (np.ndarray, np.generic, jax.Array))
+
+
 def _assert_same(array_impl_type):
     def _assert_same_impl(path, new, old):
         typenew = type(new)
         typeold = type(old)
         if typeold is jax.ShapeDtypeStruct:
             typeold = array_impl_type
-        if typenew is not typeold:
+        # All kinds of JAX and NumPy array are interchangeable here: a
+        # `filter_spec` may legitimately leave a deserialised leaf on the host
+        # rather than on a device. The shape and dtype checks below still apply.
+        if typenew is not typeold and not (
+            _is_array_type(typenew) and _is_array_type(typeold)
+        ):
             raise RuntimeError(
                 f"Deserialised leaf at path '{jtu.keystr(path)}' has changed type from "
                 f"{type(old)} in `like` to {type(new)} on disk."
--- a/tests/test_serialisation.py
+++ b/tests/test_serialisation.py
@@ -262,3 +262,20 @@ def test_eval_shape(getkey, tmp_path):
     model3 = eqx.tree_deserialise_leaves(tmp_path, model2)
 
     assert eqx.tree_equal(model, model3, typematch=True)
+
+
+def test_eval_shape_deserialise_to_host(getkey, tmp_path):
+    # A `filter_spec` may leave the deserialised leaves on the host, in which
+    # case they are NumPy rather than JAX arrays. That should not be treated as
+    # a change of type relative to the `ShapeDtypeStruct`s in `like`.
+    model = eqx.nn.MLP(2, 2, 2, 2, key=getkey())
+    eqx.tree_serialise_leaves(tmp_path, model)
+
+    def filter_spec(f, x):
+        return jax.device_get(eqx.default_deserialise_filter_spec(f, x))
+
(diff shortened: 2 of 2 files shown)


patrick-kidger:
LGTM! Merged :-)

PR #1261, opened 2026-09-19
Fix closure-converted input checks within shard map

spraharsh:
I'm addressing https://github.com/patrick-kidger/diffrax/issues/765.

# Example failure
(Addressed in the test)
Adding zero to a scalar constant can change its sharding metadata without changing its value, shape, or dtype. Equinox then rejected it with a “different dynamic arguments” error.
```python
   @ft.partial(
        jax.shard_map,
        mesh=mesh,
        in_specs=P("i"),
        out_specs=P("i"),
        check_vma=False,
    )
    def run(x):
        constant = jnp.array(1.0)
        converted = eqx.filter_closure_convert(lambda y: x * y, constant)
        # Adding zero attaches the manual mesh without changing the value.
        return converted(constant + 0)

    x = jax.device_put(jnp.array([2.0, 3.0]), NamedSharding(mesh, P("i")))
    jax.jit(run)(x)
    
 ```

# Changelog

1. Updated `_check_closure_convert_input` and added a helper function `_normalise_closure_convert_aval` to ignore reasonable sharding metadata differences while keeping checks for incompatible input types, argument structure, and static values.

2. Added a small test that both illustrates the issue and fails before the fix and passes afterwards


spraharsh:
^ I don't feel somewhat comfortable with the pyright ignore commit but me (and codex) couldn't think of anything better so if you're okay with it

PR #1262, opened 2026-09-19
Make error_if work for shard_map

spraharsh:
### Why
The current commit generalizes error_if to work for shard_map with the following behavior as updated in the doc.

```
    - `pred`: a boolean for whether to raise an error. If vmap'd or shard_map'd then an
        error will be raised if any batch/shard element has `True`.
```
In case there is a NaN being generated, then we should return an array in the shape of NaN with the same shape/sharding as `x`.  

This requires a distributed allreduce OR across all shards which has been implemented with pmax. Before these changes there was no cross device OR and can therefore only be buggy by having sharding dependent behavior in an SPMD situation. 

```python
x = [1.0, 2.0, 3.0, 4.0]
pred = x<1.5  # [True, False, False, False]
```
distributed on 4 devices `error_if` logic would return `[NaN 2.0 3.0 4.0]` (Not withstanding type errors)
distributed on 2 devices it would return  `[NaN NaN 3.0 4.0]` (Not withstanding type errors).

### Fix

1. The distributed allreduce or is implemented with pmax here

```python
    # hasattr(jax.sharding, "get_abstract_mesh") keeps compatibility
    # with older jax versions
    if on_error != "off" and hasattr(jax.sharding, "get_abstract_mesh"):
        axes = jax.sharding.get_abstract_mesh().manual_axes
        if axes:
            pred = lax.pmax(pred, axes)  # allreduce boolean or
```

2. The axis restoration commit deals with missing varying-axis annotations on error-branch outputs. This makes their types match the unchanged-input branch, allowing lax.cond to pass the checks performed by `shard_map(check_vma=True)`


spraharsh:
The second thing I wanted to mention is `pmax` introduces allreduce comms even if there is no reason to, in case `pred` is duplicated. That can be changed, but the core of the current version is simple to understand

spraharsh:
It might be worth explicitly not having comms when not necessary, because it looks like it can eat up a lot of time if say sims are batched by parameter on a large number of accelerators.

For example in the profile I got for https://github.com/patrick-kidger/diffrax/issues/718#issuecomment-5747420426, you can see that there's a ton of allreduces being added by vmap

spraharsh:
I tried to remove comms when the array is replicated. The primary issue with doing this cleanly is propagating sharding information that we would need to avoid pmax.


There is this issue, assume we distribute x across 2 devices.
```python
x = [0, 1, 2, 3]   # device 1 has [0, 1] , device 2 has [2, 3]
condition = x[0] < 1.5 
pred = lax.cond(condition, lambda: True, lambda: False) # lax.cond loses information that the predicate is varying but the array is actually # device 1 has [True, True], device 2 has [False False]

result = eqx.error_if(x, pred, "negative", on_error="nan") # equinox assumes predicate is not varying i.e replicated, therefore it will try to forego pmax and can only give [NaN NaN 3 4] without comms
```

This feels like a jax bug, because a distributed predicate that is varying /should/ have metadata saying that it is varying

spraharsh:
^^ I have a fix in mind assuming predicate has the correct metadata attached but I just want to confirm I'm not misunderstanding any details by raising an issue in jax

Edit: I think I should stare at the nitty gritties of lax.cond a little more. There are canonical answers here for the above question, but maybe there's a reason people hadn't properly defined the sharding output

